PR #3, opened 2018-04-16
Ssr

AWare:
Very quick SSR from https://medium.com/styled-components/the-simple-guide-to-server-side-rendering-react-with-styled-components-d31c6b2b8fbf 
(does not pass linter and isn't typescript) https://medium.com/styled-components/the-simple-guide-to-server-side-rendering-react-with-styled-components-d31c6b2b8fbf


AWare:
Merging this so we can work off master instead.

PR #4, opened 2018-05-04
re-usable async component wrapper (paired with Alex)

twrichards:
Introduced re-usable 'AsyncLoader' component which can wrap any other component and present spinner while 'fetching' the data, then call the render function of the component it is wrapping, on success. PLUS - some styling and refactoring (splitting out main into some smaller components).


app/client/components/footer.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Footer = () => (
+  <footer
+    css={{
+      backgroundColor: "#333",
+      color: "#bdbdbd"
+    }}
+  >
+    <div className="copyright-container">
```
AWare: I don't think these need classNames.

AWare:
👍

PR #6, opened 2018-05-08
Type fixes

AWare:
This just makes some fixes to get this closer to passing type checking and linting. 

A lot of the lint rules seem not-useful. But this is probably worth thinking about. 

I've asked on the emotion slack about the css={} error we're getting from typescript.


twrichards:
lgtm

PR #7, opened 2018-05-10
TeamCity

twrichards:
Added build-tc script for testing (with Jest TeamCity integration inc. coverage) and 'bundling' prod artifacts (not yet used but will be once riff-raff is setup for this project)


app/client/__tests__/__snapshots__/main.tsx.snap
```diff
@@ -2,24 +2,75 @@
 
 exports[`Main renders something 1`] = `
 <div
-  css="[object Object]"
+  class="css-1q69dai"
```
AWare: We could use this here https://emotion.sh/docs/testing

PR #8, opened 2018-05-11
Px to rem

AWare:
Adds https://github.com/AWare/stylis-pxtorem so that we can use pixels.

This also adds a lot of typescript definitions. These should go upstream when possible. Also: there's a lot of duplication here as typescript is fussy about local imports in .d.ts files.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -14,6 +14,7 @@
     "@types/cookie-parser": "^1.4.1",
     "@types/helmet": "^0.0.37",
     "@types/jest": "^22.2.2",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -52,7 +53,7 @@
   },
   "dependencies": {
     "@types/express": "^4.11.1",
-    "@types/node": "^9.6.5",
+    "@types/node": "^10.0.8",
```
twrichards: any issues using type definition from a more recent version of node?

twrichards:
👍

PR #9, opened 2018-05-14
Healthcheck [WIP]

AWare:
Adds a healthcheck* 

*This needs rebasing as it's got loads of cruft from the type check branch


AWare:
I think we should examine how identity redirects are handled. It seemed a little hackish putting `_healthcheck` ahead of it while / was behind it.

twrichards:
+1

twrichards:
thanks for this

PR #14, opened 2018-05-18
uncaught exceptions: nevermore

AWare:
I moved the config to the webpack define plugin. I'm not sure how best to separate client and server secrets so am thinking of moving that plugin out of common.


app/server/config.ts
```diff
@@ -1,4 +1,10 @@
+declare var DOMAIN: string;
```
twrichards: what's the purpose of these outer vars?

AWare: this tells teamcity that these are valid globals

PR #16, opened 2018-05-18
attempt css props again

AWare:
Move to css props


AWare:
This is broken again.

app/client/components/footer.tsx
```diff
@@ -5,40 +5,40 @@ import { css } from "../styles/emotion";
...
+        <a href={"#top"} css={{ color: palette.neutral["1"] }}>
           back to top
         </a>
       </div>
     </div>
     <div>
       <div
-        className={css({
+        css={{
           backgroundColor: "#333",
```
paulbrown1982: Please can we pick a consistent code style for these? e.g. all camel, or all quoted property keys.

AWare: We've gone for camelCase.

PR #22, opened 2018-05-31
Parameterise the office CIDR

paulbrown1982:
It's already public everywhere, however it's better using a variable too.

### DIFF
--- a/cloud-formation/account-frontend-cf.yaml
+++ b/cloud-formation/account-frontend-cf.yaml
@@ -30,6 +30,9 @@ Parameters:
     Description: AMI ID
     Type: String
     Default: ami-09c0f070
+  OfficeCIDR:
+    Description: Office IP range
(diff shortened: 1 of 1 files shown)


cloud-formation/account-frontend-cf.yaml
```diff
@@ -227,11 +230,11 @@ Resources:
         - IpProtocol: tcp
           FromPort: 22
           ToPort: 22
-          CidrIp: 77.91.248.0/21
+          CidrIp: !Ref OfficeCIDR
         - IpProtocol: tcp
```
paulbrown1982: Note, we should get rid of this direct box access once we're happy with things and before we go live-live.

paulbrown1982:
Tested the new CF on code and no effective changes needed because this change is just internal to the CF.

PR #23, opened 2018-06-01
client side routing

twrichards:
Introduced client-side routing using [Reach Router](https://reach.tech/router), includes basic wizard with some branching all directly linked to URL


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+{
+  "version": "0.2.0",
+  "sourceRoot": "${workspaceRoot}/app",
+  "sourceMaps": true,
+  "sourceMapPathOverrides": {
+    // if you override this, you MUST provide all defaults again
```
AWare: I should probably have linked to https://gist.github.com/jarshwah/389f93f2282a165563990ed60f2b6d6c ? for this

app/client/components/cancel/stages/areYouSure.tsx
```diff
@@ -0,0 +1,9 @@
+import { Link } from "@reach/router";
+import * as React from "react";
```
AWare: We should probably standardise on `import React from "react"`

twrichards: done

PR #25, opened 2018-06-04
check for membership at start of membership cancellation flow

twrichards:
call to "user-attributes/me" of members-data-api and renders reasons (and importantly also the subsequent steps) ONLY if the user is a member


app/client/components/asyncLoader.tsx
```diff
@@ -40,10 +41,14 @@ export default class AsyncLoader<T> extends React.Component<
         return this.state.data ? (
           this.props.render(this.state.data)
         ) : (
-          <h1>Super Boom</h1>
+          <h1>Impossible Error</h1>
```
AWare: Maybe we should reconsider how best to deal with this case.

PR #27, opened 2018-06-07
executes actual cancel POST call to members-data-api at the end of flow...

twrichards:
… and added some of the proper copy and cancellation reasons, plus some cleanup of undefined handling in 'asyncLoader'


twrichards:
ammended commit

app/client/components/user.tsx
```diff
@@ -27,12 +25,58 @@ export interface MeResponse {
...
 
+export const CancellationReasonContext: React.Context<
+  string
+> = React.createContext("");
+
+export const CancellationTypeContext: React.Context<
+  string
+> = React.createContext("error");
+
+export const CancellationUrlSuffixContext: React.Context<
```
AWare: I'm not sure this is worth having in a context, the flow for friend tier membership should be separate. Also this `HasSubscriptionGetterContext` and `cancelType` seem to be managing the same piece of information.

twrichards:
working nicely in `CODE`

PR #29, opened 2018-06-08
Check that the SC_GU_U cookie seems reasonable.

AWare:
This checks that the SC_GU_U cookie contains an expiry where we expect it and that it's not passed already. 

- [x] extend the session
- [x] agree how to check this in the cookie


app/server/identity/identityMiddleware.ts
```diff
@@ -27,18 +32,33 @@ export const withIdentity: express.RequestHandler = (
...
+  try {
+    const now = new Date().getTime();
+    const cookieString = atob(cookies[SC_GU_U].split(".")[0]);
+    const cookieData = JSON.parse(cookieString);
+    const expiry = cookieData[1];
+    const remaining = expiry - now;
+    if (remaining < 60) {
+      log.info("Insufficient time left in session.");
+      res.redirect(
+        `https://profile.${conf.DOMAIN}/signin?returnUrl=${returnUrl}`
```
twrichards: probably want to move this full redirectURL into constant above, rather than repeating it twice (line 51 & 58)

AWare:
I'm going to give this some rework.

AWare:
reworked, just need to add some usage of the client side code

AWare:
I'm not 100% sure that we can use es-cookie on the server.

PR #31, opened 2018-06-14
Better dev reload experience.

AWare:
This splits out the dev client and server webpacks and uses webpack-dev-server to proxy everything and keep the client js up to date. (And automagically refreshes on client change) This seems to fix the 502 on reload as nginx can always connect to something. 

It's a little more involved unfortunately. And we don't have HMR yet.


twrichards:
@AWare I think this has some issues having brought master into my branch - doesn't always reload code changes - have to run bundle manually quite a lot.

AWare:
hmm, try running `nginx/setup.sh` again

PR #35, opened 2018-06-22
Logging

AWare:
Switches from bunyan logging to Winston, because it seems to find it easier to log to file.

### DIFF
--- a/cloud-formation/account-frontend-cf.yaml
+++ b/cloud-formation/account-frontend-cf.yaml
@@ -126,9 +126,6 @@ Resources:
 
 
           - DomainEnvVariable: !FindInMap [ StageVariables, !Ref Stage, DomainEnvVariable ]
-            Stack: !Ref Stack
-            Stage: !Ref Stage
-            App: !Ref App
   AppRole:
     Type: AWS::IAM::Role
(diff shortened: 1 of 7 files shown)


AWare:
Something here is really broken.

AWare:
Something here is fixed

cloud-formation/account-frontend-cf.yaml
```diff
@@ -121,9 +121,14 @@ Resources:
             # RUN
             systemctl enable account-frontend
             systemctl start account-frontend
+            /opt/cloudwatch-logs/configure-logs application ${Stack} ${Stage} ${App} /var/log/account-frontend.log
+            /opt/cloudwatch-logs/configure-logs audit ${Stack} ${Stage} ${App} /var/log/account-frontend-audit.log
 
-          - DomainEnvVariable: !FindInMap [ StageVariables, !Ref Stage, DomainEnvVariable ]
 
+          - DomainEnvVariable: !FindInMap [ StageVariables, !Ref Stage, DomainEnvVariable ]
+            Stack: !Ref Stack
```
twrichards: we don't need these paramaters to be forwarded

PR #39, opened 2018-06-27
Make riffraff update amis

AWare:
Applies the cloudformation template updates before updating the ami parameter. 
This might also be soluble by removing the default ami parameter.

### DIFF
--- a/cloud-formation/account-frontend-cf.yaml
+++ b/cloud-formation/account-frontend-cf.yaml
@@ -29,7 +29,7 @@ Parameters:
   AMI:
     Description: AMI ID
     Type: String
-    Default: ami-5a747db0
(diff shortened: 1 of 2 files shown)


cloud-formation/account-frontend-cf.yaml
```diff
@@ -4,7 +4,7 @@ Parameters:
   VpcId:
     Type: String
     Description: VpcId of your existing Virtual Private Cloud (VPC)
-    Default: vpc-e6e00183
+    Default: ami-88a6af62
```
twrichards: im not sure you meant to update the default vpc to the ami id

AWare: oops
fixed

PR #42, opened 2018-07-03
Final MVP copy/design

twrichards:
Swathe of small tweaks to match the final copy/design from Elvis


app/client/components/accordion.tsx
```diff
@@ -0,0 +1,112 @@
...
+  selectedIndex: number;
+}
+
+const svgUpCaret = (
+  <svg viewBox="0 0 1024 574" width="100%" height="100%">
+    <path d="M1015 564q-10 10-23 10t-23-10L512 82 55 564q-10 10-23 10T9 564q-9-10-9-24t9-24L489 10q10-10 23-10t23 10l480 506q9 10 9 24t-9 24z" />
+  </svg>
+);
+
+const svgDownCaret = (
```
shaundillon: Slight pickiness, we could just rotate the other with CSS and avoid the redraw

app/client/components/callCentreNumbers.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Accordion } from "./accordion";
+
+export const CallCentreNumbers = () => (
+  <Accordion>
+    <div title="United Kingdom, Europe and rest of world">
+      <div>
+        <b>+ 44 (0) 330 333 6790</b>
```
shaundillon: We might want to match the true font-weight instead of just using a `<b>` tag here, as the fonts will have one specifically for bold.

app/client/components/cancel/paidMembershipFlow.tsx
```diff
@@ -54,20 +283,28 @@ const getReasonsRenderer = (routeableProps: RouteableProps) => (
...
-      )}
-    />
-  </CheckFlowIsValid>
+  <div>
+    <span
+      css={{
+        fontWeight: "bold",
+        fontSize: "20px",
+        position: "relative",
+        marginLeft: "-50vw",
```
shaundillon: Use `transform: translateX(-50%)` instead

app/client/components/cancel/stages/genericSaveAttempt.tsx
```diff
@@ -59,35 +60,71 @@ class FeedbackForm extends React.Component<
...
+    );
+  }
+
+  private getFeedbackThankYouRenderer(
+    reason: CancellationReason
+  ): () => ReactNode {
+    return () => (
+      <div
+        css={{
+          marginLeft: "15px",
```
shaundillon: Picky: this can be shorthand: `margin: "30px 0 0 15px"`

app/client/components/wizardRouterAdapter.tsx
```diff
@@ -18,8 +22,41 @@ interface RootComponentProps {
   path: string;
 }
 
+const estimateTotal = (currentStep: number, child: any) => {
+  // if(child && Array.isArray(child)) {
```
shaundillon: Maybe get rid of this?

app/client/components/cancel/stages/genericSaveAttempt.tsx
twrichards: i would rather it was explicit for readability

PR #49, opened 2018-07-16
Add media queries to oversized elements

shaundillon:


shaundillon:
@twrichards This is on CODE currently, just tested and it looks good. Not sure why github isn't picking up the builds but it passed

PR #68, opened 2018-07-25
Update dependencies and fix emotion types

shaundillon:


app/client/__tests__/main.tsx
```diff
@@ -1,4 +1,4 @@
-import "dom-testing-library/extend-expect";
```
shaundillon: https://github.com/kentcdodds/dom-testing-library/issues/69

app/client/components/buttons.tsx
```diff
@@ -28,23 +29,24 @@ export const LinkButton = ({
   // ) {
   //   console.error("insufficient constrast");
   // }
+  const linkStyles = css`
```
shaundillon: The `Interpolation` type only works on template literals

shaundillon:
Will fix this after my holiday

PR #70, opened 2018-07-27
Fix Cloudwatch logging

jacobwinch:
I noticed that this wasn't working, after some digging it turned out:

1. The log file was not being created on the EC2 instance due to filesystem permissions.
2. There were no IAM permissions for pushing logs to Cloudwatch.

This PR fixes those issues and removes a couple of redundant pieces of config/code.

### DIFF
--- a/app/server/log.ts
+++ b/app/server/log.ts
@@ -14,14 +14,3 @@ export const log = winston.createLogger({
     new winston.transports.Console({ format: winston.format.simple() })
   ]
 });
-
-export const audit = winston.createLogger({
-  level: "info",
-  format: winston.format.json(),
-  transports: [
-    new winston.transports.File({
-      filename: `${location}/manage-frontend-audit.log`
-    }),
-    new winston.transports.Console({ format: winston.format.simple() })
-  ]
-});
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -101,7 +101,9 @@ Resources:
             # add user
             groupadd manage-frontend
             useradd -r -s /usr/bin/nologin -g manage-frontend manage-frontend
+            touch /var/log/manage-frontend.log
             chown -R manage-frontend:manage-frontend /etc/gu
+            chown manage-frontend:manage-frontend /var/log/manage-frontend.log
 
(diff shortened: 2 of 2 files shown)


app/server/log.ts
```diff
@@ -14,14 +14,3 @@ export const log = winston.createLogger({
     new winston.transports.Console({ format: winston.format.simple() })
   ]
 });
-
-export const audit = winston.createLogger({
```
jacobwinch: This was never used so I removed it

cfn.yaml
```diff
@@ -125,8 +127,6 @@ Resources:
             systemctl enable manage-frontend
             systemctl start manage-frontend
             /opt/cloudwatch-logs/configure-logs application ${Stack} ${Stage} ${App} /var/log/manage-frontend.log
-            /opt/cloudwatch-logs/configure-logs audit ${Stack} ${Stage} ${App} /var/log/manage-frontend-audit.log
```
jacobwinch: as above

cfn.yaml
```diff
@@ -146,13 +146,16 @@ Resources:
             Action:
               - sts:AssumeRole
       Policies:
-        - PolicyName: UpdateSSHKeys
```
jacobwinch: We don't need these permissions, since we are using SSM instead.

cfn.yaml
```diff
@@ -251,3 +254,9 @@ Resources:
           FromPort: 443
           ToPort: 443
           CidrIp: 0.0.0.0/0
+
+  ManageFrontendLogGroup:
```
jacobwinch: Adding the log group in CFN so that we can enforce a retention policy here.

app/server/log.ts
paulbrown1982: I think there might have been a requirement for an audit logger.

jacobwinch: @paulbrown1982 - I've merged this as I'd like to have at least some logging working for this app. Let's reintroduce the audit logger in a separate ticket if there is a genuine requirement for this.

PR #72, opened 2018-07-30
Update the AMI as part of the CFN update

jacobwinch:
Prior to this we were attempting to update the launch config with an old AMI id (i.e. the one specified in the CFN template) during the first CFN update, and then updating the launch config again to use the correct (latest) AMI. 

As well as being inefficient, this method completely stopped working this morning as the AMI referenced in CFN no longer exists (presumably due to https://github.com/guardian/amigo/pull/196).

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -8,17 +8,12 @@ deployments:
     app: manage-frontend
     parameters:
       templatePath: cloudformation/cfn.yaml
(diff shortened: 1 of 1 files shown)


jacobwinch:
@paulbrown1982 - agreed. Simplifying setting up a new project is definitely on the Developer Experience team's list!

PR #116, opened 2018-09-12
Tweaks for Gus

twrichards:
- Made the action required banner within the membership tab yellow
- Moved overall headings in the two flows to align vertically with everything else


paulbrown1982:
👍

PR #118, opened 2018-09-17
remove blue from nav

AWare:
This just removes the blue from the nav, bringing us closer to  https://github.com/guardian/frontend/pull/20352


paulbrown1982:
Current:
picture 508
Target:
picture 509

PR #120, opened 2018-09-26
fixed broken membership FAQ links 

twrichards:
Also reduced complexity by always using theguardian.com rather than environment specific domain


jacobwinch:
Aside: I noticed that we are linking to contribute.theguardian.com, these should probably point at support.theguardian.com now, but let's change that in a different PR.

PR #126, opened 2018-10-02
responsive breadcrumb style progress bar and applied it to the payment update flow

twrichards:
_NOTE: also tested on IE_

**Step 1**
image
**Step 2**
image
**Step 3**
image
**Mobile/Responsive**
image
**Before**
image


app/client/components/payment/update/stripeLogo.tsx
```diff
@@ -3,19 +3,16 @@ import React from "react";
 export const StripeLogo = () => (
   <svg width="119px" height="26px">
     <path
-      fill-rule="evenodd"
```
twrichards: this line was invalid html (copied from Stripe svg asset) - just noticed in this change so removing

PR #128, opened 2018-10-03
Refactoring/Genericising to add contributions card payment flow

twrichards:
Some renaming/refactoring/restructuring/genericising to allow for contributions card payment update flow (also reduced some package tangle). Also added a basic contributions tab (full parity with profile will follow in a future PR).

PR #129, opened 2018-10-04
genericised cancellation flow

twrichards:
_Continues the refactoring work from https://github.com/guardian/manage-frontend/pull/128_

This change genericises the cancellation flow and few accompanying wider improvements/refactors


app/client/components/cancel/cancellationReason.ts
```diff
@@ -0,0 +1,13 @@
+export interface CancellationReason {
+  reasonId: string;
```
jacobwinch: I know this isn't new, but we could introduce a type for these?

app/client/components/cancel/cancellationSummary.tsx
```diff
@@ -50,5 +50,7 @@ export const CancellationSummary = (cancelType: string) => (
   isCancelled(subscription) ? (
     actuallyCancelled(cancelType, subscription)
   ) : (
-    <GenericErrorScreen loggingMessage="Not actually cancelled" />
+    <GenericErrorScreen
+      loggingMessage={"Not actually cancelled : " + cancelType}
```
jacobwinch: Again, not a new thing, but we could add a bit more context to this logging to help newcomers?

app/client/components/cancel/cancellationReason.ts
twrichards: done

PR #130, opened 2018-10-08
re-worked the NoMembership page into a generic NoProduct page 

twrichards:
re-worked the NoMembership page into a generic NoProduct page which provides more [product specific] granularity to the tracking and adds referral tracking for support.theguardian.com

_includes copy tweaks from Gus_


app/client/components/checkFlowIsValid.tsx
```diff
@@ -9,13 +10,20 @@ const renderChildrenIfValidated = (props: CheckFlowIsValidProps) => (
...
+        productType={props}
+      />
+    </PageContainer>
   );
 
 export type MeValidator = (me: MeResponse) => boolean;
 
 export interface CheckFlowIsValidProps extends ProductType {
   children: any;
+  supportRefererSuffix: string;
```
jacobwinch: Does this get used for tracking or something? We could use a type here?

app/client/components/noProduct.tsx
```diff
@@ -0,0 +1,31 @@
...
+  inTab: boolean;
+  supportRefererSuffix: string;
+}
+
+export const NoProduct = (props: NoProductProps) => (
+  <div>
+    <h2>You do not currently have a {props.productType.friendlyName}.</h2>
+    {props.inTab ? (
+      <p>
+        To manage your existing contribution or subscription, please select from
```
jacobwinch: Hmm actually, doesn't this text still imply that the user is on the membership tab?

twrichards: fair, but this is just a subtle prompt to navigate to other tabs, will review with Gus and amend in a future PR if required

app/client/components/checkFlowIsValid.tsx
twrichards: this is appended on the end of other strings which are used in tracking, so collating the possibilities into a type won't really add anything here

jacobwinch: Thanks for explaining this 👍

PR #135, opened 2018-10-15
Contributions cancellation flow

twrichards:
Added all the copy and revised flow from @ionamckendrick with various refactors to facilitate it.

image
image
image
image


app/client/components/cancel/cancellationReason.ts
```diff
@@ -9,16 +9,24 @@ export interface CancellationReason {
...
+  | "mma_direct_debit"
+  | "mma_one_off"
+  | "mma_wants_monthly_contribution"
+  | "mma_wants_annual_contribution"
   | "mma_health"
   | "mma_break_from_news"
   | "mma_values"
-  | "mma_other";
+  | "mma_other"
+  | "";
```
jacobwinch: Why do we include an empty string here?

twrichards: this is to allow the CancellationReasonContext (see cancellationContexts.tsx) to be parameterised by the CancellationReasonId type but still have a valid defaultValue. There are potentially other ways to achieve this but this seemed the neatest, if it is too nasty I can have a go at refactoring it.

jacobwinch: I'm not keen on it because it makes the type definition of CancellationReasonId a bit meaningless...

Can we make null the default value, and do something like this:
_React.Context<CancellationReasonId | null>_

PR #136, opened 2018-10-18
Contributions Tab (parity with profile)

twrichards:
Replicates and improves upon the contributions tab from profile.
image

## Key Feature : Amount Updater
NOW uses the bounds from https://github.com/guardian/support-frontend/blob/5aa237d/assets/helpers/contributions.js which differ based on currency and  frequency (monthly vs annual) and also has an improved 'sense check' where you are asked to double check if you increase your contribution by x10 (indicative of an accidental extra key press).

image

image

image

image

image

image

image

image

...or, if it fails...

image

### DIFF
--- a/app/client/components/payment/update/stripeCardInputForm.tsx
+++ b/app/client/components/payment/update/stripeCardInputForm.tsx
@@ -1,9 +1,8 @@
 import { NavigateFn } from "@reach/router";
 import React from "react";
 import { ReactStripeElements } from "react-stripe-elements";
-import palette from "../../../colours";
 import { maxWidth } from "../../../styles/breakpoints";
-import { sans } from "../../../styles/fonts";
+import { validationWarningCSS } from "../../../styles/fonts";
 import { Button } from "../../buttons";
 import { GenericErrorScreen } from "../../genericErrorScreen";
 import { Spinner } from "../../spinner";
@@ -21,7 +20,6 @@ export interface StripeCardInputFormProps
 
 export interface StripeCardInputFormState {
   isGeneratingToken: boolean;
-  isValid: boolean;
   error: {
     code?: string;
     message?: string;
@@ -36,7 +34,6 @@ export class StripeCardInputForm extends React.Component<
 > {
   public state: StripeCardInputFormState = {
     isGeneratingToken: false,
-    isValid: false,
     error: {},
     readyElements: []
   };
@@ -140,9 +137,7 @@ export class StripeCardInputForm extends React.Component<
       return (
         <p
           css={{
-            color: palette.red.medium,
-            fontFamily: sans,
-            fontSize: "0.8rem",
+            ...validationWarningCSS,
             marginTop: "5px"
           }}
         >
--- a/app/client/components/updatableAmount.tsx
+++ b/app/client/components/updatableAmount.tsx
@@ -1,6 +1,5 @@
 import React, { ChangeEvent } from "react";
 import {
-  Subscription,
   SubscriptionPlan,
   WithSubscription
 } from "../../shared/productResponse";
@@ -20,13 +19,16 @@ enum MinOrMax {
 // min & max values based on https://github.com/guardian/support-frontend/blob/5aa237d/assets/helpers/contributions.js
 const calculateMinOrMaxAmount = (
   minOrMax: MinOrMax,
-  plan: SubscriptionPlan
+  subscriptionPlan: SubscriptionPlan
 ) => {
-  if (plan.interval.toLowerCase() === "month") {
-    if (plan.currencyISO === "NZD" || plan.currencyISO === "AUD") {
+  if (subscriptionPlan.interval.toLowerCase() === "month") {
+    if (
+      subscriptionPlan.currencyISO === "NZD" ||
+      subscriptionPlan.currencyISO === "AUD"
+    ) {
       return minOrMax === MinOrMax.minimum ? 10 : 200;
     }
-    if (plan.currencyISO === "CAD") {
+    if (subscriptionPlan.currencyISO === "CAD") {
       return minOrMax === MinOrMax.minimum ? 5 : 166;
     }
     return minOrMax === MinOrMax.minimum ? 2 : 166; // monthly default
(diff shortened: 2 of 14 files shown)


twrichards:
Original amount updater implementation on profile : https://github.com/guardian/frontend/pull/18506

twrichards:
Tested in IE11 and on iPhone6S, via BrowserStack

app/client/components/payment/update/stripeCardInputForm.tsx
```diff
@@ -21,7 +20,6 @@ export interface StripeCardInputFormProps
 
 export interface StripeCardInputFormState {
   isGeneratingToken: boolean;
-  isValid: boolean;
```
twrichards: this was an un-used 'state' variable so removed

app/server/server.ts
```diff
@@ -156,6 +160,22 @@ Object.values(ProductTypes).forEach((productType: ProductType) => {
         })()
     )
   );
+
+  server.use(
```
twrichards: moved banner redirect into the dynamic routes (one per ProductType) rather than being a broad on any string

ionamckendrick:
A small thing, but could we switch to "a month" rather than "per month". Not a blocker, but if you were happy to change ... Otherwise it's looking good!

twrichards:
> A small thing, but could we switch to "a month" rather than "per month". Not a blocker, but if you were happy to change ... Otherwise it's looking good!

@ionamckendrick done (although not updated the screenshots - too many)

app/client/components/updatableAmount.tsx
```diff
@@ -0,0 +1,287 @@
...
+    isApplyingUpdate: false,
+    currentAmount: this.initialAmount,
+    newAmount: this.initialAmount,
+    validationMessage: ""
+  };
+
+  public render(): React.ReactNode {
+    return (
+      <div css={wrappingContainerCSS}>
+        {this.state.isApplyingUpdate ? (
```
jacobwinch: I find it pretty hard to follow code this type of code (where we use conditional logic and each expression is pretty big ReactNode). 

Perhaps extracting some nodes into consts and/or using some smaller functions would help to make this more readable?

app/server/server.ts
```diff
@@ -44,7 +44,11 @@ const globals: Globals = {
 
 server.use(
   (req: express.Request, res: express.Response, next: express.NextFunction) => {
-    res.header("Cache-Control", "private");
+    // this header is VERY IMPORTANT and prevents caching (on both CDN and in browsers)
```
jacobwinch: I would address this in its own PR - it doesn't seem related to the Contributions tab to me?

app/client/components/updatableAmount.tsx
```diff
@@ -0,0 +1,287 @@
...
+import AsyncLoader from "./asyncLoader";
+import { Button } from "./buttons";
+import { wrappingContainerCSS } from "./productPage";
+
+enum MinOrMax {
+  minimum,
+  maximum
+}
+
+// min & max values based on https://github.com/guardian/support-frontend/blob/5aa237d/assets/helpers/contributions.js
```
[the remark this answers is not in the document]
jacobwinch: Also, I quite like the support-frontend convention of separate this logic into helpers files - what do you think about doing something similar in this project?

app/client/components/updatableAmount.tsx
```diff
@@ -0,0 +1,287 @@
...
+export class UpdatableAmount extends React.Component<
+  UpdatableAmountProps,
+  UpdatableAmountState
+> {
+  public initialAmount = this.props.subscription.nextPaymentPrice / 100.0;
+  public state = {
+    inEditMode: false,
+    isApplyingUpdate: false,
+    currentAmount: this.initialAmount,
+    newAmount: this.initialAmount,
```
jacobwinch: It might be clearer if this was null to begin with?

app/client/components/updatableAmount.tsx
```diff
@@ -0,0 +1,287 @@
...
+                currentAmount: this.state.newAmount
+              });
+              return null;
+            }}
+            loadingMessage={"Updating..."}
+            errorRender={() => {
+              this.setState({
+                isApplyingUpdate: false,
+                validationMessage:
+                  "Updating failed this time. Please try again later..."
```
jacobwinch: We should add a card to analyse why this update might fail and whether this message is always going to be appropriate (definitely one for a follow up PR though)

app/server/server.ts
twrichards: moved it to https://github.com/guardian/manage-frontend/pull/140 (which also nicely covers one the remediation Trello card)

app/client/components/updatableAmount.tsx
```diff
@@ -0,0 +1,287 @@
+import React, { ChangeEvent } from "react";
```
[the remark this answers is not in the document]
twrichards: not really since this component is used to render the amount both in and out of edit mode, so it is an 'updatable amount'

app/client/components/updatableAmount.tsx
twrichards: not sure i agree, these state values need to start with the seed value from the API, mainly so the validation can show they need to change the value before they can confirm, see...
image

app/client/components/updatableAmount.tsx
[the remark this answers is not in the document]
twrichards: added tests.

not sure I agree regarding moving out into global config file as that undermines the encapsulation of this self contained component. If in future we need these thresholds elsewhere in the application then that will warrant centralising them.

app/client/styles/fonts.ts
```diff
@@ -328,3 +330,9 @@ export const fonts = `
 export const sans = `"Guardian Text Sans Web", "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif`;
 export const serif = `"GuardianTextEgyptian", Georgia, serif`;
 export const headline = `GH Guardian Headline, ${serif}`;
+
+export const validationWarningCSS = {
```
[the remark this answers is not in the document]
twrichards: since its used in a few places and its purely typeface related CSS its a good enough home for it for now

PR #139, opened 2018-10-19
added a new profileReferrer url param used in the signin/reauthenticate step

twrichards:
added a new `profileReferrer` url param added during the signin/reauthenticate step so we can better track pageviews as to how/when they were authenticated 

for example:

1. if the incoming request is for `https://manage.theguardian.com/foo/bar?baz=test&INTCMP=test`
2. and signin is required, the profile signin url would be `https://profile.theguardian.com/signin?INTCMP=test&returnUrl=https%3A%2F%2Fmanage.thegulocal.com%2Ffoo%2Fbar%3FprofileReferrer%3Dsignin%26baz%3Dtest%26INTCMP%3Dtest`
3. after successfully signing in the user ends up on `https://manage.theguardian.com/foo/bar?`**`profileReferrer=signin`**`&baz=test&INTCMP=test`

also refactored the construction of the redirects to be cleaner

NOTE: this change was suggested by @jfsoul as part of a review of tracking

### DIFF
--- a/app/server/identity/identityMiddleware.ts
+++ b/app/server/identity/identityMiddleware.ts
@@ -9,42 +9,41 @@ export const withIdentity: express.RequestHandler = (
   res: express.Response,
   next: express.NextFunction
 ) => {
-  const returnUrl = url.format({
-    protocol: "https",
-    host: req.get("host"),
-    pathname: req.originalUrl
-  });
   const cookies = req.header("cookie");
-
   const user = getUser(cookies);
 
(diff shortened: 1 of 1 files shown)


twrichards:
@jfsoul just had some other ideas around tracking the time since the previous signin alongside, as this could help us adjust that one hour if for example it turns out most users have signed in 1.5hrs before, we'd perhaps adding too much friction
@paulbrown1982 also suggested adding a before and after timestamp to track how long they spent in the signin process

PR #157, opened 2018-11-19
add 'payment flow' only for 'print' products

twrichards:


app/client/components/payment/update/paymentUpdated.tsx
```diff
@@ -48,11 +49,16 @@ const ConfirmedNewPaymentDetailsRenderer = (subscription: Subscription) => {
               </div>
             ) : (
               <>
-                <div>
-                  <b>Next Payment:</b> {subscription.plan.currency}
-                  {(subscription.nextPaymentPrice / 100.0).toFixed(2)} on{" "}
-                  {formatDate(subscription.nextPaymentDate)}
-                </div>
+                {subscription.nextPaymentPrice &&
```
twrichards: these can be null for print products, so only shows this on the payment confirmed page if it has a `nextPaymentPrice` and `nextPaymentDate`

twrichards:
> Why do we return to /manage on subscribe? @paulbrown1982 

Because we're not migrating the rest of the manage page on subscribe over to the new platform for a while (since there's actually nowhere for it to go whilst we're stuck in the tab model) so for **paper** card update flow 'return to your account' and the similar button on completion page will go back to the subscribe platform for the time being.

PR #162, opened 2018-12-11
fixed annoying outline around buttons when clicked

twrichards:
Buttons used to have the focus highlight wrapping them once clicked, which was ugly/distracting for a short period of time until the desired action happened. Now only shows for actual focus (for keyboard navigation accessibility).

for example...
image


twrichards:
FYI trying to tease out commits from direct debit stuff to reduce the size of that PR

PR #163, opened 2018-12-11
genericise payment flow

twrichards:
This is all the pre-work before adding direct debit update flow. This PR includes...

- Lots of renames away from _card, token, stripe_ towards generic names *
- directory structure changes to group card stuff together
- Introduction of `NewPaymentMethodDetail` type with one for card `newCardPaymentMethodDetail.tsx`

*unfortunately the diff for some of the file renames/moves is showing them as new/deleted 😢


twrichards:
FYI trying to tease out commits from direct debit stuff to reduce the size of that PR

app/client/components/payment/update/confirmPaymentUpdate.tsx
```diff
@@ -0,0 +1,155 @@
...
+          isNewPaymentMethodDetail(newPaymentMethodDetail) &&
+          hasProduct(mdaResponse) ? (
+            <WizardStep
+              routeableStepProps={labelPaymentStepProps(props)}
+              extraFooterComponents={<QuestionsFooter />}
+            >
+              <h3>Please confirm your change from...</h3>
+              <CurrentPaymentDetails {...mdaResponse.subscription} />
+              <h3>...to...</h3>
+              {newPaymentMethodDetail.render()}
```
paulbrown1982: Out of interest only - and maybe I need that tutorial in React - why is this an imported module with newPaymentMethodDetail.render() and not just a nested custom JSX tag? e.g. <NewPaymentMethodDetail /> ?

twrichards: So the imported module `NewPaymentMethodDetail` just defines the type/interface. I happen to have named the variable (from the consumer) the same thing, but note the  lower case starting letter (as per convention), it just so happens to a have method on that type called `render` but that could be changed... It isn't actually a 'Component' and hence couldn't be displayed by using custom JSX.

twrichards: `NewPaymentMethodDetail` could be rewritten as an extension of Component and hence could be instantiated/called by adding a JSX tag, but that could be misleading I think.

twrichards:
> I need to chat to you about how the .Consumer stuff works

`Consumer`s are one part of the [`Context` api in React](https://reactjs.org/docs/context.html), and allow you to access values from much further up the component tree (specified with a `Provider`) without having to pass down as `props` through every single intermediate component. Particularly useful in this scenario as a number of the intermediate components are super generic (`WizardStep` for example) and so passing down as a prop would either make those not very re-usable OR you would lose the type safety (contexts have type parameters when used in TypeScript), not to mention how messy it would be.

> and what the single bar does in `export type PaymentUpdateTypeName = "card" | "direct_debit";`

This is a String Literal type (about half way down https://www.typescriptlang.org/docs/handbook/advanced-types.html) and simply means and property of this type can only be one of those values (at compile time of course). You can think of it as an enum (which TypeScript also has, but they have some problems to do with their default underlying number representation - which I recently investigated with @Ap0c)

PR #166, opened 2018-12-27
prevented double-counting of initial page view in Ophan

twrichards:
When checking the 5XX alert email (not related, and resolved itself) I noticed the first page was being double counted after my hasty change last Friday (https://github.com/guardian/manage-frontend/pull/165)

The double counting is because an initial event is sent when the Ophan tracking script loads the page and then again with the usual 'page change'


twrichards:
Tested in CODE ✅

PR #167, opened 2019-01-05
Integrated new identity redirect service

twrichards:
The manage my account area requires users to be signed-in/re-authenticated within the last hour. Previously this was done within the manage platform by reading the contents of the cookies (which wasn't ideal anyway) but since identity introduced session invalidation, the manage platform was unaware of this.

For every request, manage platform now calls a new IDAPI service (see https://github.com/guardian/identity/pull/1477 and https://github.com/guardian/identity/pull/1478, made by @Calum-Campbell) which returns either an _ok_ if the user signed-in/re-authenticated within the last hour, or an _ok_ with a `redirect` `url` in the response body...

- with /reauthenticate if the cookies are present and not expired but last auth was more than hour.
- with /signin if the cookies are expired or aren't present

 **Importantly this moves all the authentication concerns into the identity platform, which is where they should belong.**

In order to speak to IDAPI a client access token is required which is retrieved from a private S3 bucket (which is a new feature in manage platform, hence the new `aws-sdk` dependency).

**_To the end user, everything looks identical._**

- [x] Tested in CODE


app/server/idapiConfig.ts
```diff
@@ -0,0 +1,61 @@
...
+  host: string;
+  accessToken: string;
+}
+
+// tslint:disable-next-line:no-object-mutation
+process.env.AWS_PROFILE = "membership"; // TODO do this with CredentialProviderChain
+
+const s3 = new AWS.S3({
+  region: "eu-west-1"
+  // credentialProvider: new AWS.CredentialProviderChain([
```
jacobwinch: Is this going to be revisited in a subsequent PR or should it be removed?

app/server/idapiConfig.ts
```diff
@@ -0,0 +1,61 @@
...
+      ) {
+        const idapiConfig = parsed as IdapiConfig;
+        log.info("IDAPI: using " + idapiConfig.host);
+        return idapiConfig;
+      }
+      handleIdapiRelatedError(
+        "IDAPI config missing 'host' and/or 'accessToken' properties"
+      );
+    } catch (err) {
+      handleIdapiRelatedError("could not parse IDAPI config", err);
```
jacobwinch: Good to have this clear logging here, but if this config is fundamentally broken we probably want to fail the healthcheck too?

jacobwinch:
Looks good overall, 👍👍👍 for this especially:

> Importantly this moves all the authentication concerns into the identity platform, which is where they should belong.

However, I think it's worth adapting the healthcheck accordingly before we merge this.

app/server/idapiConfig.ts
twrichards: great point! I'll add it now

app/server/idapiConfig.ts
twrichards: it will be - it works fine now by setting the env variable (and is the way _dot-components_ is doing it - just @adamnfish suggested using the `CredentialProviderChain ` explicitly - which I will get round to doing it soon

app/server/identity/identityMiddleware.ts
```diff
@@ -1,53 +1,131 @@
...
+              Cookie: getCookiesOrEmptyString(req)
+            }
+          }
+        )
+          .then(
+            redirectResponse =>
+              redirectResponse.json() as Promise<RedirectResponseBody>
+          )
+          .then(redirectResponseBody => {
+            log.info(JSON.stringify(redirectResponseBody));
```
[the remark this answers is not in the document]
twrichards: I'm actually wondering whether to remove the log line as it will just spam the logs (even with more context). Thoughts?

twrichards:
@Calum-Campbell please could give this a quick once over?

app/server/identity/identityMiddleware.ts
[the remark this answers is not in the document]
jacobwinch: Yeah, perhaps just removing it would be better - this could be pretty noisy if we're logging it for every request?

app/server/idapiConfig.ts
twrichards: done

app/server/identity/identityMiddleware.ts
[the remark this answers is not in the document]
twrichards: removed

Calum-Campbell:
@twrichards Looks good to me from our side

PR #177, opened 2019-01-15
Include encryptedEmail query parameter in identity redirects

Mullefa:
The work identity have been doing recently means that a user's (encrypted) email will soon be included in payment failure links in emails. 

For payment failure links that send the user to the manage website, include the encrypted email in redirects to identity frontend so that the data can be utilised to facilitate user sign-in.

I've tested this locally by entering into the browser the url:

```
https://manage.thegulocal.com/payment/contributions?encryptedEmail=<email-token>
```

With email tokens corresponding to a current account, a guest account and a new user respectively. The results are as expected:

* current account: redirected to the second step where email address is autofilled and they are prompted for their password
* guest account: redirected to the second step where they are prompted to verify their email address
* new user: redirected to the second step where email address is autofilled and they are prompted to create an account


Mullefa:
Calum and me have tested the payment failure emails end-to-end on CODE (with this PR deployed to CODE) and its working as expected.

PR #178, opened 2019-01-15
Digipack tab parity

tjsilver:
Replicated the digipack tab from profile.theguardian.com. This required adding the following additional rows to the existing productPage implementation in `manage-frontend`:

- subscriber ID
- trial remaining
- subscription product

Global navigation and tabs within `manage-frontend` now point to the new tab.

_Changes to `frontend` to redirect traffic will follow soon._

image

### DIFF
--- a/app/client/components/userNav.tsx
+++ b/app/client/components/userNav.tsx
@@ -120,7 +120,7 @@ export class UserNav extends React.Component {
   public userNavItems: UserNavItem[] = [
     {
       title: "Comments & replies",
-      link: `/profile/user` // note this hits a redirect/proxy endpoint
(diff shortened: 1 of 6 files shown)


app/client/components/userNav.tsx
```diff
@@ -145,7 +145,7 @@ export class UserNav extends React.Component {
     },
     {
       title: "Digital Pack",
-      link: `${profileHostName}/digitalpack/edit`,
+      link: `/digitalpack`,
```
paulbrown1982: Utterly minor point - we could change this back to double-quotes now that there's no string interpretation required.

tjsilver: Done, along with all the others in this file.

PR #181, opened 2019-01-18
FIX signin/reauth redirect when sending the user to their comments management section

twrichards:
FIX: the redirect for sending the user to their comments management was behaving like an API endpoint and returning 401 when sign-in/reauth was required (poor user experience) - it now actually redirects them to sign-in/reauth first if needs be


twrichards:
I've tested this in CODE

PR #185, opened 2019-01-28
combined 'Subscriptions' tab (Paper, Weekly & Digipack - not Membership or Contributions)

tjsilver:
### Combined paper, weekly and digipack into new Subscriptions tab
Created a grouped product type ('ContentSubscription' - see https://github.com/guardian/members-data-api/pull/366) which allows paper, weekly and digipack subscriptions to be displayed together in one tab.
Also relies on https://github.com/guardian/members-data-api/pull/367

Corresponding tab labels and global nav changes on the main website : https://github.com/guardian/frontend/pull/20991 (which should be deployed after this change)

#### Details
- Mapped grouped product types back to the specific product where appropriate
- Disabled payment update flow for the grouped subscriptions
- Ensured that payment flows, etc for ContentSubscriptions always return back to /subscriptions
- Added functionality to display 'contact us' link when more than one subscription of paper or weekly is held, otherwise added link to `https://subscribe.theguardian.com/manage`

image

#### Messaging and links:
|  |  | Screenshot |
| --- | --- | --- |
| Paper | One subscription |image image <br /> _Goes to `https://subscribe.theguardian.com/manage`_|
|  Paper |More than one subscription |image|
|  Weekly |One subscription |image image _Goes to `https://subscribe.theguardian.com/manage`_|
| Weekly |More than one subscription | image|

### DIFF
--- a/app/client/components/productPage.tsx
+++ b/app/client/components/productPage.tsx
@@ -24,6 +24,7 @@ import { maxWidth, minWidth } from "../styles/breakpoints";
 import { headline } from "../styles/fonts";
 import { Button, LinkButton } from "./buttons";
 import { getCancellationSummary } from "./cancel/cancellationSummary";
+import { InlineContactUs } from "./inlineContactUs";
 import { MembershipLinks } from "./membershipLinks";
 import { NoProduct } from "./noProduct";
 import { PageContainer, PageHeaderContainer } from "./page";
@@ -182,6 +183,9 @@ const getProductDetailRenderer = (
   const productType = originalProductType.mapSoCalledToSpecific
     ? originalProductType.mapSoCalledToSpecific(productDetail)
     : originalProductType;
+  const alternateManagementCtaLabel =
+    productType.alternateManagementCtaLabel &&
+    productType.alternateManagementCtaLabel(productDetail);
   return (
     <div
       key={productDetail.subscription.subscriberId}
@@ -326,6 +330,20 @@ const getProductDetailRenderer = (
             ) : (
               undefined
             )}
+            {productType.alternateManagementUrl &&
+              alternateManagementCtaLabel &&
+              (productDetailListLength > 1 ? (
+                <div css={{ fontWeight: "bold" }}>
+                  {alternateManagementCtaLabel}, please <InlineContactUs />
+                </div>
+              ) : (
+                <a href={productType.alternateManagementUrl}>
+                  <Button
+                    text={alternateManagementCtaLabel + " click here"}
+                    right
+                  />
+                </a>
+              ))}
           </PageContainer>
         </>
       )}
--- a/app/client/components/productPage.tsx
+++ b/app/client/components/productPage.tsx
@@ -16,7 +16,7 @@ import {
 } from "../../shared/productResponse";
 import {
   createProductDetailFetcher,
-  ProductTypeWithProductPage
+  ProductTypeWithProductPageProperties
 } from "../../shared/productTypes";
 import palette from "../colours";
 import { maxWidth, minWidth } from "../styles/breakpoints";
@@ -141,7 +141,7 @@ const getPaymentMethodRow = (
 
 const getPaymentPart = (
   productDetail: ProductDetail,
-  productType: ProductTypeWithProductPage
+  productType: ProductTypeWithProductPageProperties
 ) => {
   if (productDetail.isPaidTier) {
     return (
@@ -173,9 +173,9 @@ const getPaymentPart = (
   }
 };
 
-const getProductRenderer = (productType: ProductTypeWithProductPage) => (
-  apiResponse: MembersDataApiResponse[]
-) => {
+const getProductRenderer = (
+  productType: ProductTypeWithProductPageProperties
+) => (apiResponse: MembersDataApiResponse[]) => {
   const productDetailList = apiResponse.filter(hasProduct).sort(sortByJoinDate);
   return (
     <>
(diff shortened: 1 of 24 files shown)


app/shared/productTypes.tsx
```diff
@@ -10,28 +10,31 @@ import { membershipCancellationReasons } from "../client/components/cancel/membe
...
 import { NavItem, navLinks } from "../client/components/nav";
 import { MeResponse } from "./meResponse";
-import { formatDate, Subscription } from "./productResponse";
+import { formatDate, ProductDetail, Subscription } from "./productResponse";
 
 export type ProductFriendlyName =
   | "membership"
   | "recurring contribution" // TODO use payment frequency instead of 'recurring' e.g. monthly annual etc
   | "newspaper subscription"
   | "digital pack"
```
jacobwinch: This list doesn't feel consistent. If we are using "newspaper subscription", I think it would be more consistent to have "digital pack subscription" and "Guardian Weekly subscription"?

app/client/components/productPage.tsx
```diff
@@ -173,9 +175,185 @@ const getPaymentPart = (
...
+    : originalProductType;
+  const alternateManagementCtaLabel =
+    productType.alternateManagementCtaLabel &&
+    productType.alternateManagementCtaLabel(productDetail);
+  return (
+    <div
+      key={productDetail.subscription.subscriberId}
+      css={{
+        background:
+          productDetailListLength > 1 && (listIndex + 1) % 2 !== 0
```
jacobwinch: It might be worth pulling this into a const to express the intent?

app/client/components/productPage.tsx
```diff
@@ -173,9 +175,185 @@ const getPaymentPart = (
...
+                  primary
+                  right
+                />
+              </PageContainer>
+            </div>
+          ) : (
+            undefined
+          )}
+          <PageContainer>
+            {productPageProperties.showSubscriptionId ? (
```
jacobwinch: I'd be interested to know why we still need to expose the subscription ID here?

twrichards: Carried over from both /manage and the Digital Pack tab on profile. I'm all for removing this in the long run, but it should be a separate change.
As it stands, at the very least, users need to be shown this so they can use the manual online cancellation form (only membership and contributions have the 3 step cancellation flow in manage)

app/shared/productTypes.tsx
```diff
@@ -219,28 +236,55 @@ export const ProductTypes: { [productKey: string]: ProductType } = {
...
+      title: "Subscriptions",
+      navLink: navLinks.subscriptions,
       noProductInTabCopy:
         "To manage your existing membership or contribution, please select from the tabs above.",
       tierRowLabel: "Subscription product",
-      showSubscriberId: true,
-      showTrialRemainingIfApplicable: true
+      showSubscriptionId: true
+    },
+    mapSoCalledToSpecific: (productDetail: ProductDetail) => {
```
jacobwinch: We are matching on strings, and we are only expecting one of three strings, so it feels like we could use a type to help us out here?

```
type subscription = "DigitalPack" | "Newspaper" | "Digital Pack"
```

As an aside, we typically use tier in the context of membership only, so I'm slightly surprised to find these values under productDetail.tier.

app/client/components/user.tsx
```diff
@@ -73,26 +85,31 @@ const User = () => (
...
-          key={productType.urlPart}
-          path={"/payment/" + productType.urlPart}
-          productType={productType}
-          currentStep={1}
-        >
-          <ConfirmPaymentUpdate
-            path="confirm"
+      {Object.values(ProductTypes)
+        .filter(
+          // i.e. don't create payment update flow for grouped product types
```
jacobwinch: This comment is helpful but perhaps having a shouldShowUpdatePaymentFlow boolean on the productType would make this even clearer?

app/client/components/productPage.tsx
jacobwinch: >  I'm all for removing this in the long run, but it should be a separate change.

Definitely - thanks for explaining

jacobwinch:
I'm not sure if it's better to fix this here (or in members-data-api) but I don't think we should be labelling the product as Guardian Weekly **- Domestic** or Guardian Weekly **- ROW** if we can avoid it.

I believe the parts in bold are coming through like this because there are two distinct products in the Zuora catalog (due to differing shipping costs). This is all ugly complexity that we should aim to hide from the user if possible!

paulbrown1982:
Is this change backwards compatible, so if the profile. site is not updated immediately after this is deployed, the /digitalpack route still works fine?

app/shared/productTypes.tsx
twrichards: also the types aren't available at **runtime**, so string literal types are only useful when passing known strings around

app/shared/productTypes.tsx
twrichards: done

app/client/components/user.tsx
twrichards: actually I didn't create an additional boolean for this because I want it to be based on whether the  product type is a grouping (i.e. it has grouped => specific function) - thereby minimising config.
However I did create a function in `productTypes.tsx` which removes the need for the comment 😄 
https://github.com/guardian/manage-frontend/blob/36305bb815ced4a071adb4a7486078e42edacb4d/app/shared/productTypes.tsx#L116-L117

PR #186, opened 2019-01-28
expose initial INTCMP query arg as GA dimension and in Ophan component event action

twrichards:
To help D&I correlate payment update success/failure events (and theoretically other things) back to email campaigns (`CCX`, `PF1`, `PF2` etc. - denoted in the **`INTCMP`** query parameter) the initial  INTCMP value is now exposed as both...

- a GA dimension (already a defined dimension in GA - with index 12)
- in the `campaignCode` of the Ophan component event (which required refactoring in order to send these events with the correct structure)

### TODO
- [x] Confirm with D&I that this is sufficient and makes things easy for them

### DIFF
--- a/app/shared/ophanTypes.ts
+++ b/app/shared/ophanTypes.ts
@@ -1,0 +1,73 @@
+export type OphanProduct =
+  | "CONTRIBUTION"
+  | "RECURRING_CONTRIBUTION"
+  | "MEMBERSHIP_SUPPORTER"
+  | "MEMBERSHIP_PATRON"
+  | "MEMBERSHIP_PARTNER"
+  | "DIGITAL_SUBSCRIPTION"
+  | "PAPER_SUBSCRIPTION_EVERYDAY"
+  | "PAPER_SUBSCRIPTION_SIXDAY"
+  | "PAPER_SUBSCRIPTION_WEEKEND"
+  | "PAPER_SUBSCRIPTION_SUNDAY"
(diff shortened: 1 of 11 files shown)


app/shared/ophanTypes.ts
```diff
@@ -0,0 +1,73 @@
...
+  | "EXPAND"
+  | "LIKE"
+  | "DISLIKE"
+  | "SUBSCRIBE"
+  | "ANSWER"
+  | "VOTE"
+  | "CLICK";
+
+export type OphanComponentType =
+  | "READERS_QUESTIONS_ATOM"
```
jacobwinch: Do we need all of these component types in the repo? They don't all seem that relevant to manage-frontend.

app/shared/productTypes.tsx
```diff
@@ -252,6 +268,7 @@ export const ProductTypes: { [productKey: string]: ProductType } = {
     allProductsProductTypeFilterString: "Weekly",
     urlPart: "guardianweekly",
     validator: (me: MeResponse) => me.contentAccess.guardianWeeklySubscriber,
+    getOphanProductType: () => "PRINT_SUBSCRIPTION", // TODO create a GUARDIAN_WEEKLY Product in Ophan data model
```
jacobwinch: I don't have enough context to properly understand the implications of this - but presumably D&I are aware that GW will be coming through under this generic type?

PR #189, opened 2019-02-07
BEAST: Emotion10 upgrade (pretty broad changes + some supporting version bumps)

twrichards:
So what was initially just intended as quick version bump of emotion, has turned into a massive migration to `@emotion` version 10.

The majority of the changes are just moving things round, wrapping them in the `css` function to avoid type errors (although in future PR I would like to refactor back to passing objects directly to the `css` prop as much as possible - as this allows for much nicer composition of the objects e.g. spread operations).

Along the way also bumped `typescript` (& related) devDependencies and improved the jest, tsc, tslint & webpack configurations - most of these additional changes were necessary to make the migration easier.

- [x] Manually regression tested in CODE (including with BrowserStack)


app/client/components/cancel/cancellationFlow.tsx
```diff
@@ -58,15 +59,15 @@ class ReasonPicker extends React.Component<
...
-            css={{
+            css={css({
               ...cssInheritFont,
               width: "100%",
               height: "32px",
               border: "1px black solid",
-              color: this.state.reasonPath ? undefined : palette.neutral["4"],
-              appearance: "menulist"
+              color: this.state.reasonPath ? undefined : palette.neutral["4"]
+              // appearance: "menulist"
```
jacobwinch: Are these commented out lines still required?

twrichards: this one is because the typecheck fails, and not sure its adding anything, but don't want to loose it just yet as I will be revisiting some of the stuff in this PR in the near future. In fact, kinda need to re-write this component, currently uses a `<select>` which is terrible on iOS (so want to rebuild as just a repeatable clickable `div`).

app/client/components/callCentreNumbers.tsx
```diff
@@ -1,11 +1,11 @@
+import { css } from "@emotion/core";
 import React from "react";
```
SiAdcock: I notice that you're still importing `React` into `*.tsx` files. Is this still necessary? I believe JSX should [transpile down to calls to `@emotion/core`'s `jsx`](https://emotion.sh/docs/@emotion/babel-preset-css-prop#features), which  is auto-imported by `@emotion/babel-preset-css-prop`. Is this correct?

twrichards: Interesting, if I remove that though I get some compilation issues, I wonder also whether it's worth leaving in for expressiveness to make clear that React is being used.

SiAdcock: Thanks for confirming that it isn't just me receiving these compilation issues. I think it's due to a bug in `@babel/plugin-transform-react-jsx` (fixed in babel/babel#9095), although this fix has yet to roll out to `@babel/preset-typescript` (proposed fix in babel/babel#9497).

My understanding is `@emotion/babel-preset-css-prop` implicitly adds the [`jsx`](https://emotion.sh/docs/@emotion/babel-preset-css-prop#features) function and `/** @jsx jsx */` pragma to each module, so `React` isn't used directly in the module. I think explicitly importing React is misleading, since it's not actually being used. However, we still need to import it to keep TypeScript happy, which feels a bit wonky.

In dotcom-rendering, I think we're going to keep importing React until the above issue in the Babel TypeScript preset is fixed, then hopefully we can remove it.

twrichards: thanks for the extra explanation.

In particular...
> In dotcom-rendering, I think we're going to keep importing React until the above issue in the Babel TypeScript preset is fixed, then hopefully we can remove it.

...I think we'll do the same with `manage-frontend`

twrichards:
After discussion with the team - we decided to merge this without a single explicit approver (considering it's such a big change that's hard for anyone else to give 👍 comfortably)

app/client/components/callCentreNumbers.tsx
SiAdcock: Hi @twrichards 

babel/babel#9497 has been merged, so you should be able to upgrade `@babel/preset-typescript` and remove the React imports, if you want to.

After some careful consideration, we've decided not to migrate to using the CSS prop, so we're sticking with the vanilla `emotion` API (`css` to generate classnames, `cx` for composition). I'll write a more detailed explanation elsewhere, but the short version is that the new API is optimised for universal React or running on the client, which we is an approach we are no longer taking. We have already encountered friction styling AMP elements, which uses a combination of custom SSR components and non-React client side presentational logic.

app/client/components/callCentreNumbers.tsx
```diff
@@ -20,8 +20,8 @@ export interface CallCentreNumbersProps {
 }
```
SiAdcock: Forgive silly question, I've noticed this in a few places, but is it necessary to pass these styles to the `css` function? I think object styles can be passed straight to the `css` prop, but I might be wrong (we're using string styles)

app/jest.config.js
```diff
@@ -4,5 +4,6 @@ module.exports = {
   },
```
SiAdcock: Given you're using `jest-ts` without specifying a babel config, I think your tests don't go through Babel, so none of the Babel plugins/presets will be picked up in your tests. This might not be a problem for you, but if you're relying on `@emotion/babel-preset-css-prop` to work in your tests, you might hit an issue (e.g. `css` prop is not a valid attribute). This was a problem for us, so I thought I'd share

PR #193, opened 2019-02-15
Include autoSignInToken query parameter in identity redirects

Mullefa:
Follows on from #177.

The work identity have been doing recently means that am auto sign-in token will soon be included in payment failure links in emails.

For payment failure links that send the user to the manage website, include the auto sign-in token in redirects to identity frontend so that the data can be utilised to facilitate user sign-in.

I've tested this locally by entering into the browser the url:

```
https://manage.thegulocal.com/payment/contributions?autoSignInToken=<auto-sign-in-token>
```

### DIFF
--- a/app/server/identity/identityMiddleware.ts
+++ b/app/server/identity/identityMiddleware.ts
@@ -52,11 +52,14 @@ export const augmentRedirectURL = (
     query: {
       ...parsedSimpleURL.query,
       ...(req.query.INTCMP && { INTCMP: req.query.INTCMP }), // if undefined then this param is omitted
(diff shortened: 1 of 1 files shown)


Mullefa:
I like the explicitness, but it does mean a new PR is required per query parameter update. Deferring to supporter experience. Happy to update PR if required.

twrichards:
> I like the explicitness, but it does mean a new PR is required per query parameter update. Deferring to supporter experience. Happy to update PR if required.

Yeah lets leave it for now, and revisit on the next time we need to add something here. Please feel free to merge...

PR #194, opened 2019-02-18
Prevent payment method update in unsafe scenarios and allow users with no payment method to add card

jacobwinch:
When a payment method is not correctly linked to a Zuora account _and_ the prerequisites for auto-cancellation are not met, it's possible for users to build up 'debts' whilst continuing to receive their subscription benefits. Currently many users in this bad state are unable to add a new payment method to fix the situation (which means that we cannot recover them easily). For other users, it _is_ possible to add a new payment method, but if they choose to do so this will inadvertently cause the payment for several invoices to be collected at once.

This PR allows us to use the updated response added by https://github.com/guardian/members-data-api/pull/370 to prevent unsafe payment method updates.

It also allows users without a correctly linked payment method to add a new payment card in safe scenarios (which enables us to start collecting their payments again 💰) 

## Screenshots

**Adding a payment method when there isn't one already linked / set up correctly**

image

image

image

image

**Restricting payment method update**

Payment method update is deemed unsafe:

image

### DIFF
--- a/app/client/components/payment/update/confirmPaymentUpdate.tsx
+++ b/app/client/components/payment/update/confirmPaymentUpdate.tsx
@@ -39,6 +39,12 @@ class ExecutePaymentUpdate extends React.Component<
     hasHitComplete: false
   };
 
+  private paymentMethodChangeType: string =
+    this.props.productDetail.subscription.paymentMethod ===
+    PaymentMethod.resetRequired
+      ? "reset"
+      : "update";
+
   public render(): React.ReactNode {
     return this.state.hasHitComplete ? (
       <PaymentUpdateAsyncLoader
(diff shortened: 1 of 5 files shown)


twrichards:
Looks spot on @jacobwinch - might be worth having some screenshots for what this scenario looks like (both on the product page and on each step of the payment update)?

twrichards:
Also, shall we offer creating of Direct Debit rather than just card?

jacobwinch:
> Also, shall we offer creating of Direct Debit rather than just card?

I think this would be worth adding if we decide to retain this feature longer term. However I am hoping that once we've helped the users who are currently in this state we'll put better preventative measures in place which will make maintaining this unnecessary 🤞

app/client/components/payment/update/confirmPaymentUpdate.tsx
```diff
@@ -83,7 +89,9 @@ class ExecutePaymentUpdate extends React.Component<
     ) {
       trackEvent({
         eventCategory: "payment",
-        eventAction: this.props.newPaymentMethodDetail.name + "_update_success",
+        eventAction: `${this.props.newPaymentMethodDetail.name}_${
```
jacobwinch: Not quite sure why the linter wants this to take up 3 lines though...

PR #195, opened 2019-02-19
Change cancellation reason picker

tjsilver:
... to use new radioButton component in place of `<select>` for better UX on iOS.

| | Before `<select>` | After |
| --- | --- | --- |
| **iOS** |  😱 😱 😱 😱 😱 😱😱 😱  image  | image |
| **Desktop** | image | image |

### DIFF
--- a/app/client/components/radioButton.tsx
+++ b/app/client/components/radioButton.tsx
@@ -11,41 +11,74 @@ export interface RadioButtonProps {
   groupName: string;
 }
 
-export const RadioButton = (props: RadioButtonProps) => (
-  <label
-    css={css({
-      display: "block",
-      cursor: "pointer",
-      marginBottom: "7px"
-    })}
-  >
-    <input
-      type="radio"
-      value={props.value}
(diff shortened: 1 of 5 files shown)


app/client/components/radioButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <input
+      type="radio"
+      value={props.value}
+      name={props.groupName}
+      onChange={props.onChange}
+      checked={props.checked}
+      css={{ display: "none" }}
+    />
+    <div css={{ position: "relative" }}>
+      <div
```
paulbrown1982: Is this component rendering an alternative radio button (o) to the <input> above??

tjsilver: yes

paulbrown1982: Let's just double check it works OK with some accessibility tools / Lynx etc.

tjsilver: Have now made the radio buttons accessible.

PR #196, opened 2019-02-20
improve display of physical subs

twrichards:
## _**Must be released after https://github.com/guardian/members-data-api/pull/371**_

Various improvements with the main underlying goal of ensuring is price rise is displayed correctly (once the API says it can be displayed, see https://github.com/guardian/members-data-api/pull/371).

### Now displays Join Date OR Start Date depending on product type 
_(rather than using join date as potentially misleading fallback or indeed showing both)_
image
image

### Price rises now display (if appropriate)
_(i.e. if `shouldBeVisible` is true - see https://github.com/guardian/members-data-api/pull/371)_
image

Including in the "Six for Six" scenario...
image

### Type of physical subscription now shown for Paper and Guardian Weekly

image

image

image

### DIFF
--- a/app/client/components/productPage.tsx
+++ b/app/client/components/productPage.tsx
@@ -155,12 +155,13 @@ const getPaymentPart = (
   if (productDetail.isPaidTier) {
     return (
       <>
-        {productDetail.subscription.nextPaymentDate && (
-          <ProductDetailRow
-            label={"Next payment date"}
-            data={formatDate(productDetail.subscription.nextPaymentDate)}
-          />
-        )}
+        {productDetail.subscription.nextPaymentDate &&
+          !productDetail.alertText && (
+            <ProductDetailRow
+              label={"Next payment date"}
+              data={formatDate(productDetail.subscription.nextPaymentDate)}
+            />
+          )}
         <ProductDetailRow
           label={
             productDetail.subscription.plan.interval.charAt(0).toUpperCase() +
--- a/app/client/components/productPage.tsx
+++ b/app/client/components/productPage.tsx
@@ -246,7 +246,16 @@ const getProductDetailRenderer = (
                   {listIndex + 1}
                 </h2>
               ) : (
-                <h2>{productType.alternateTierValue || productDetail.tier}</h2>
+                <h2>
+                  {productType.alternateTierValue || productDetail.tier}
+                  {productType.displayPlanName && (
+                    <i>
+                      {productType.displayPlanName(
+                        getMainPlan(productDetail.subscription)
+                      )}
+                    </i>
+                  )}
+                </h2>
               )}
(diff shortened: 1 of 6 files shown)


app/client/components/productPage.tsx
```diff
@@ -280,36 +318,51 @@ const getProductDetailRenderer = (
...
             ) : (
               undefined
             )}
-            <ProductDetailRow
-              label={"Start date"}
-              data={formatDate(
-                productDetail.subscription.start || productDetail.joinDate
-              )}
-            />
+            {productDetail.joinDate !== productDetail.subscription.start && (
```
jacobwinch: Personally I find being shown both a join date _and_ a start date slightly confusing - is join date the date that I signed up via the website and start date the date that my first paper will arrive? Was this a new request from UX?

app/client/components/productPage.tsx
```diff
@@ -152,26 +156,51 @@ const getPaymentPart = (
...
-              productType={productType}
-            />
+            <>
+              <UpdatableAmount
+                mainPlan={mainPlan}
+                subscriptionId={productDetail.subscription.subscriptionId}
+                productType={productType}
+              />
+              {futurePlan &&
+                futurePlan.productRatePlanId !== mainPlan.productRatePlanId &&
```
jacobwinch: Do we need the productRatePlanId comparison here? I think comparing the amounts would be sufficient?

app/client/components/productPage.tsx
paulbrown1982: Also from a language perspective, it’s a membership that you ‘join’ and a subscription that you ‘sign up for:’ and which ‘starts on:’.

app/client/components/productPage.tsx
paulbrown1982: +1, I agree with Jacob’s earlier comment that we shouldn’t expose the productRatePlanId from Zuora.

twrichards: removed in https://github.com/guardian/manage-frontend/pull/196/commits/3dd81ad407ac94a95d6158ed5e428bb39e2365fc

app/client/components/productPage.tsx
twrichards: @jacobwinch 

> is join date the date that I signed up via the website and start date the date that my first paper will arrive

yes to that but no not a request from UX, it was that previously join date was used as a fallback which occasionally misleading, I figured providing both would be explicit.

Based on @paulbrown1982's point about language, I could easily have it display join date for membership and start date for other subscription... I'll do that.

twrichards: done in https://github.com/guardian/manage-frontend/pull/196/commits/f1a04a98b3241bcfaa871a43a761c6c01a3380a0

PR #197, opened 2019-02-22
Add functionality to AB test user journeys on manage.

Mullefa:
When a user lands on manage from an external link, store the information of the ab test they were in (if applicable) so that it can be included on all component events sent by manage. Additionally, include the ab test data as query parameters if the user if kicked to manage (so that sign-in success can be measured across test variants).

With this PR, there are now 5 query parameters that are getting included (if present) if the user gets redirected to profile. So additionally this PR handles the inclusion of those query parameters in a more general manner.

### DIFF
--- a/app/client/components/analytics.tsx
+++ b/app/client/components/analytics.tsx
@@ -60,7 +60,8 @@ export const trackEvent = ({
           ]
         },
         action: "VIEW",
-        value: eventValue !== undefined ? `${eventValue}` : undefined
+        value: eventValue !== undefined ? `${eventValue}` : undefined,
+        abTest: window.guardian.abTest
       }
     });
   }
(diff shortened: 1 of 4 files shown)


app/server/identity/identityMiddleware.ts
```diff
@@ -45,22 +45,31 @@ export const augmentRedirectURL = (
...
+  // To avoid potential clashes with query parameters that have a special meaning on profile (e.g. error),
+  // only forward specific query parameters.
+  const queryParametersNamesToForward = [
+    "INTCMP",
+    // Some links in payment failure emails include the user's (encrypted) email as a query parameter
+    // and/or an auto sign-in token. If present, include these in the identity redirect url,
+    // since they can be utilised to facilitate sign-in by identity frontend.
+    "encryptedEmail",
+    "autoSignInToken",
+    // By passing these profile to, can measure the sign in rates across test segments.
```
jacobwinch: Minor: should this be (something like):

```suggestion
    // By passing these to profile, we can measure the sign in rates across test segments.
```

app/client/components/analytics.tsx
```diff
@@ -78,11 +79,23 @@ export class AnalyticsTracker extends React.PureComponent<{}> {
...
 
       if (window.guardian) {
         // tslint:disable-next-line:no-object-mutation
         window.guardian.INTCMP = this.INTCMP;
+
+        const abName = queryParams.abName;
+        const abVariant = queryParams.abVariant;
+
+        if (abName && abVariant) {
+          window.guardian.abTest = {
```
jacobwinch: I'm missing a bit of context here... but is it safe to assume that a user will only ever be part of one AB test when they hit /manage?

Mullefa: I think so. My understanding from talking with @twrichards: AB tests in manage are done through Optimize. So this AB test relates to ab tests that were initiated outside of the manage platform. 

Examples:
- payment failure emails
- payment failure banners

Maybe some context as a comment would be useful?

jacobwinch: I think this is fine as is, I just wanted to make sure I'd understood correctly 👍

Mullefa:
Have tested on CODE that the desired query parameters are getting forwarded to profile, and that the ab test data is getting included in component events sent by manage.

PR #198, opened 2019-02-22
Update membership FAQ

tjsilver:
- Updated questions
- Tidied up CSS
- Made link text more accessible
- Consolidated 3 separate accordions into one

### DIFF
--- a/app/client/components/membershipFAQs.tsx
+++ b/app/client/components/membershipFAQs.tsx
@@ -10,6 +10,12 @@ const headerCss = {
   borderBottom: `1px solid ${palette.neutral["5"]}`
 };
 
+const linkCss = {
+  textDecoration: "underline",
+  color: palette.blue.dark,
+  ":visited": { color: palette.blue.dark }
+};
+
 export const MembershipFAQs = (props: RouteableProps) => (
   <>
     <PageHeaderContainer selectedNavItem={navLinks.membership}>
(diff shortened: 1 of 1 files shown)


app/client/components/membershipFAQs.tsx
```diff
@@ -22,184 +28,85 @@ export const MembershipFAQs = (props: RouteableProps) => (
...
             <p>
-              Of course a paywall is something we have looked at and we would
-              never rule it out but putting one up now would get in the way of
-              growing and deepening the relationships that are critical for our
-              future.
+              At the moment, only Digital Pack offers an ad-free experience on
+              the website.{" "}
+              <a
+                css={linkCss}
+                href="https://support.theguardian.com/uk/subscribe/digital"
```
jacobwinch: I think we should link to /subscribe/digital, rather than the UK version of this page.

Using the former allows us to redirect to the most appropriate version of the Digital Pack landing page based on a user's geolocation (see https://github.com/guardian/support-frontend/blob/master/conf/routes#L90-L91).

app/client/components/membershipFAQs.tsx
```diff
@@ -255,174 +147,33 @@ export const MembershipFAQs = (props: RouteableProps) => (
...
-          <div>
-            We are currently reviewing how membership is priced, what it
-            includes/involves, and how it fits with the various print and
-            digital content packages that we offer readers, including
-            subscriptions.
-          </div>
-        </div>
-        <div title="I am under 18 - can I still be a Member?">
-          <div>No. You have to be 18 or over to join.</div>
+          <div>Yes, we now accept payments via Paypal.</div>
```
jacobwinch: Wow, these really haven't been updated in a while! 😅 

https://github.com/guardian/membership-frontend/pull/1519

app/client/components/membershipFAQs.tsx
tjsilver: Ah yes! Done.

PR #209, opened 2019-03-27
now builds and sends 'acquisitionData' query param when linking to 'support'

twrichards:
This replaces the INTCMP as it can provide more granular information and ensures correct attribution (for example in the 'Acquisitions Slicer').

Also refactored everything that links to 'support' to use this class/function.

See the URL when hovering over the Contribute button for example...
image

### DIFF
--- a/app/client/__tests__/__snapshots__/main.tsx.snap
+++ b/app/client/__tests__/__snapshots__/main.tsx.snap
@@ -999,7 +999,7 @@ exports[`Main renders something 1`] = `
                 >
                   Support The Guardian
                 </div>
-                <a
+                <div
                   css={
                     Object {
                       "display": "inline-block",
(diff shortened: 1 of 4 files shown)


twrichards:
FYI @joelochlann @jranks123

paulbrown1982:
Do you need to replace, or can you include both? I believe INTCMP is pivoted on in some heavily-used Google Analytics reports. Maybe you should speak to the BRR Acquisition team?

twrichards:
> @paulbrown1982
> Do you need to replace, or can you include both? I believe INTCMP is pivoted on in some heavily-used Google Analytics reports. Maybe you should speak to the BRR Acquisition team?

@joelochlann suggested I remove it

joelochlann:
ah sorry, I might have been missing a piece of the puzzle - thought it was redundant because it duplicates `componentId`

PR #217, opened 2019-04-26
cleaned up last few instances where 'X_GU_ID_FORWARDED_SCOPE' header wasn't being passed

twrichards:
The 'X_GU_ID_FORWARDED_SCOPE' header is integral to the 'auto sign-in tokens' (aka magic links) for the payment update flow (following CC Expiry, Direct Debit Mandate Failure and Payment Failure emails), see...

- https://github.com/guardian/identity/pull/1527
- https://github.com/guardian/members-data-api/pull/384
- https://github.com/guardian/identity-processes/pull/23

While end-to-end testing there were a few outstanding cases uncovered, where this header wasn't being passed when it should.

### DIFF
--- a/app/server/identity/identityMiddleware.ts
+++ b/app/server/identity/identityMiddleware.ts
@@ -165,9 +165,9 @@ export const withIdentity: (statusCode?: number) => express.RequestHandler = (
             headers: {
               "X-GU-ID-Client-Access-Token":
                 "Bearer " + idapiConfig.accessToken,
-              [X_GU_ID_FORWARDED_SCOPE]: getScopeFromRequestPathOrEmptyString(
-                req.path
-              ),
+              [X_GU_ID_FORWARDED_SCOPE]:
+                req.header(X_GU_ID_FORWARDED_SCOPE) ||
+                getScopeFromRequestPathOrEmptyString(req.path),
               Cookie: getCookiesOrEmptyString(req)
             }
           }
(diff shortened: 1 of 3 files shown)


app/server/identity/identityMiddleware.ts
```diff
@@ -165,9 +165,9 @@ export const withIdentity: (statusCode?: number) => express.RequestHandler = (
             headers: {
               "X-GU-ID-Client-Access-Token":
                 "Bearer " + idapiConfig.accessToken,
-              [X_GU_ID_FORWARDED_SCOPE]: getScopeFromRequestPathOrEmptyString(
-                req.path
-              ),
+              [X_GU_ID_FORWARDED_SCOPE]:
+                req.header(X_GU_ID_FORWARDED_SCOPE) ||
```
Mullefa: At the moment, am I correct in thinking that the value if header if defined will always equal the scope derived from the request path? In which case, what's the motivation of doing this? Optimisation? Future-proofing?

twrichards: so the actual endpoint that this `identityMiddleware.ts` will be intercepting may not itself include `/payment/` e.g. `/api/me/mma` which is general use api endpoint.
So the line in question first tries to use the SCOPE coming from the client-side (i.e. only if this general endpoint is being called from within the payment flow), but falls-back to one inferred from the request URL.

Mullefa: I see. And would this header ever not be set?

twrichards: i don't think there are any scenarios currently, but this does ensure the payment specific **api** endpoints for example, always send the right scope, regardless of what the client says.

PR #220, opened 2019-05-30
holiday-stop-api integration

twrichards:
#### Integrate the `holiday-stop-api` (see https://github.com/guardian/support-service-lambdas/pull/302).

The API sits behind an API Gateway, which requires...

- an `x-api-key` header, the value for which is fetched from a new config file along with the `host` (this differs obviously for each stage). Loading another config file from S3 warranted some refactoring/abstraction of this mechanism, so it's now much cleaner (and along the way found a way to use the `AWS.CredentialProviderChain` rather than overriding `AWS_PROFILE` environment variable, which is lovely - see https://github.com/guardian/dotcom-rendering/issues/575 for more info).
- and an `x-identity-id` header, the value for which was already coming back from the IDAPI call it makes before every request (due to `withIdentity()`) but this needed exposing to subsequent request handler, which is now achieved via `res.locals.identity.userId` - which required some refactoring (during which I was able to eliminate an API call in the `profileRedirectHandler` which is used for the 'Comments & replies' link in the global navigation).

... the integration also has the notion of `normalMode` and `testMode` and the correct instance of the `holiday-stop-lambda` is called based on the value in the...
- `X-Gu-Membership-Test-User` this is somewhat coincidentally the same header as comes back from members-data-api elsewhere in manage, but should be sent for all interactions with the holiday-stop-api

The URL structure in manage differs slightly from the API itself (as it does for other APIs manage interacts with like `members-data-api`)

| Method | manage | holiday-stop-api | Description |
| --- | --- | --- | --- |
| GET | `/api/holidays /{PRODUCT_URL_PART} /potential?startDate={yyyy-MM-dd}&endDate={yyyy-MM-dd}` | `/{STAGE}/potential?startDate={yyyy-MM-dd}&endDate={yyyy-MM-dd}`  (with `x-product-name-prefix` header set)  | returns an array of dates for each issue impacted between the dates for the given product |
| GET | `/api/holidays` | `/{STAGE}/hsr` | returns all holiday stops (past & present) for the user |
| GET | `/api/holidays /{PRODUCT_URL_PART}` | `/{STAGE}/hsr` (with `x-product-name-prefix` header set) | returns all holiday stops (past & present) but including a calculated 'first available date' based on the type of product |
| GET | `/api/holidays /{PRODUCT_URL_PART} /{SUBSCRIPTION_NAME}` | `/{STAGE}/hsr /{SUBSCRIPTION_NAME}` (with `x-product-name-prefix` header set) | returns all holiday stops (past & present) for the user filtered on the specified subscription also including a calculated 'first available date' based on the type of product |
| POST | `/api/holidays /{PRODUCT_URL_PART}` | `/{STAGE}/hsr` | creates a new all holiday stop, example body `{ "start": "2023-06-10", "end": "2024-06-14", "subscriptionName": "A-S00071783" }`|
| DELETE | `/api/holidays /{PRODUCT_URL_PART} /{SUBSCRIPTION_NAME} /{SF_ID}` | `/{STAGE}/hsr /{SUBSCRIPTION_NAME} /{SF_ID}` | deletes the holiday stop request from SalesForce (with Id matching `{SF_ID}`) |

NOTE: `{PRODUCT_URL_PART}` can only be `guardianweekly` at this time based on the definitions in `productTypes.ts` (where this is mapped to `Guardian Weekly` for the `x-product-name-prefix` header)

### DIFF
--- a/app/server/awsIntegration.ts
+++ b/app/server/awsIntegration.ts
@@ -1,0 +1,61 @@
+import * as AWS from "aws-sdk";
+import { GetObjectRequest } from "aws-sdk/clients/s3";
+import Raven from "raven";
+import { conf } from "./config";
+import { log } from "./log";
+
+const REGION = "eu-west-1";
+
+const PROFILE = "membership";
+
+const CREDENTIAL_PROVIDER = new AWS.CredentialProviderChain([
+  () => new AWS.SharedIniFileCredentials({ profile: PROFILE }),
+  ...AWS.CredentialProviderChain.defaultProviders
+]);
+
+export const S3 = new AWS.S3({
+  region: REGION,
+  credentialProvider: CREDENTIAL_PROVIDER
+});
+
+export const handleAwsRelatedError = (message: string, detail?: any) => {
+  log.error(message, detail);
+  Raven.captureMessage(message);
+};
+
+export const s3ConfigPromise = <ConfigInterface>(
+  configPathPart: string,
+  ...fields: string[]
+) =>
+  (async () => {
+    const configPath: GetObjectRequest = {
+      Bucket: "gu-reader-revenue-private",
+      Key: `manage-frontend/${conf.STAGE}/${configPathPart}-${conf.STAGE}.json`
+    };
+
+    const s3PromiseResult = await S3.getObject(configPath).promise();
+
+    if (s3PromiseResult.Body) {
+      try {
+        const parsed = JSON.parse(s3PromiseResult.Body.toString());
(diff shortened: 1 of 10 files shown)


twrichards:
TESTED IN PROD
image
NOTE that at the time of writing BOTH `normalMode` and `testMode` for PROD are pointing to the CODE instance of `holiday-stop-api` - see https://github.com/guardian/support-service-lambdas/pull/302 for more details)

PR #221, opened 2019-06-10
Bump browserslist-useragent

tomrf1:
the current version is reported as high severity by snyk

PR #222, opened 2019-06-10
Update yarn.lock

tomrf1:
follow up from https://github.com/guardian/manage-frontend/pull/221


tomrf1:
latest commit looks much better.
I was using a very old yarn (as a result of not running `nvm use` under /app)

PR #223, opened 2019-06-13
consent banner

twrichards:
## Added the standard Guardian consent banner
image

- Displays only if **`GU_TK`** cookie is NOT present
- The `I'm OK with that` button writes the cookie on the parent domain, i.e. not `manage` sub-domain
- Fairly similar to https://github.com/guardian/support-frontend/pull/1677
- Uses another top-level `<Router>` instance in `user.tsx` to cleanly prevent the banner on the payment flow.

_Also required additions to `<Button>` component (to display the ✔️) plus some tweaks/corrections to the shared colour palette (although this file should be refactored in a future PR to exactly match https://design.theguardian.com/#colour-palette)._

### DIFF
--- a/app/client/colours.ts
+++ b/app/client/colours.ts
@@ -34,7 +34,7 @@ const palette = {
   yellow: {
     light: "#fff8e5",
     medium: "#ffe500",
-    dark: "#edd600"
+    dark: "#ffbb50"
   },
   neutral: {
     "1": "#121212",
--- a/app/client/components/analytics.tsx
+++ b/app/client/components/analytics.tsx
@@ -24,14 +24,11 @@ export interface Event {
   eventValue?: number;
 }
 
-export const trackEvent = ({
-  eventCategory,
-  eventAction,
-  product,
-  eventLabel,
-  eventValue
-}: Event) => {
-  if (window.ga) {
+export const trackEvent = (
+  { eventCategory, eventAction, product, eventLabel, eventValue }: Event,
+  alsoTrackInGA = true
+) => {
+  if (alsoTrackInGA && window.ga) {
     window.ga(
       "send",
       "event",
@@ -66,6 +63,8 @@ export const trackEvent = ({
     });
   }
 };
+
+export const trackEventInOphanOnly = (event: Event) => trackEvent(event, false);
 
 export const applyAnyOptimiseExperiments = () => {
   if (typeof window !== "undefined" && window.ga && window.dataLayer) {
--- a/app/client/components/buttons.tsx
+++ b/app/client/components/buttons.tsx
@@ -19,6 +19,8 @@ export interface CommonButtonProps {
   primary?: true;
   hollow?: true;
   hide?: boolean;
+  hoverColour?: string;
+  leftTick?: true;
 }
 
 export interface LinkButtonProps extends CommonButtonProps {
@@ -30,15 +32,20 @@ export interface ButtonProps extends CommonButtonProps {
   onClick?: () => void;
 }
 
-const applyArrowStyleIfApplicable = (
+const applyIconStyleIfApplicable = (
   hover: boolean,
   left?: true,
-  right?: true
+  right?: true,
+  leftTick?: true
 ) => {
   if (left) {
     return hover ? styles.leftHover : styles.left;
   } else if (right) {
     return hover ? styles.rightHover : styles.right;
+  } else if (leftTick) {
+    return {
+      padding: "4px 21px 3px 16px"
+    };
   }
   return {
     padding: "1px 18px 0 18px",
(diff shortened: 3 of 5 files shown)


twrichards:
FYI @rupertbates

twrichards:
One outstanding question for me, is whether we should add tracking events for impressions of this banner, and click throughs (to either the main CTA, or the other links)?

twrichards:
Tested in CODE ✅

mchv:
We can't use GA until the user has consented and please notice that this banner is expected to be updated with the categories from the `Phase 2 - GDPR` work which is expected to be delivered before end of this quarter.

twrichards:
> We can't use GA until the user has consented and please notice that this banner is expected to be updated with the categories from the `Phase 2 - GDPR` work which is expected to be delivered before end of this quarter.

Thanks @mchv - duly noted... to double check we are safe to track impressions & click events on the banner with Ophan though, correct?

Yes coincidentally this repo will actually be the home of the `Phase 2` UI etc. (currently being implemented in this repo by @guardian/commercial-dev )

mchv:
> Thanks @mchv - duly noted... to double check we are safe to track impressions & click events on the banner with Ophan though, correct?

This is a good question, but I need to clarify few terms below due to the ambiguity of `Ophan` to answer correctly:
 
 - `Ophan tracker`  is a client-side piece of code, which sends various generic information including attention-time and clicks to `Ophan storage` , `Datalake` and any other datastore that may use those informations in accordance to legal and security requirements.  
   
 - `Ophan storage` is where a subset of information from `Ophan tracker` is stored to be used by `Ophan dashboard` with a specific retention policy linked to its usage.  
 
 - [`Ophan dashboard`](https://dashboard.ophan.co.uk) is the dashboard enabling users (in particular audience team) to view aggregated data from `Ophan storage` with specific privileges and restrictions in accordance to GDPR legislation and privacy guidelines.  

 - `Datalake` is where a subset of information from `Ophan tracker` is stored to be used by `Data Analysts` with a specific retention policy and specific applications (such as our Campaign Management Tool) receive pieces of data based on user consent.

**About tracking GDPR consent**

In [GDPR Article 6](https://gdpr-info.eu/art-6-gdpr/) mention the following alternatives to consent, `(c)`:
```
processing is necessary for compliance with a legal obligation to which the controller is subject
``` 
and `(f)`
```
processing is necessary for the purposes of the legitimate interests pursued by the controller or by a third party, except where such interests are overridden by the interests or fundamental rights and freedoms of the data subject which require protection of personal data, in particular where the data subject is a child.
```

As GDPR requires us for compliance to `store` that a user has consented, I think that either `Article 6 (c)` or `(f)` allows us  to track impressions & click events on the GDPR banner using `Ophan tracker` and store that data in the `Datalake`.

**About first party vs third party**

Reading above, one may think that `(c)` could be applied to justify use of a third-party solution. I am not sure this is the case. However, european union data protection and privacy advisory group clearly advise for first party rather than third party.      

In [`Directive 2002/58/EC`](https://ec.europa.eu/justice/article-29/documentation/opinion-recommendation/files/2012/wp194_en.pdf) regarding relation to `First party analytics` in `Section 4.3`
it says:
```
4.3 First party analytics

Analytics are statistical audience measuring tools for websites, which often rely on cookies.
These tools are notably used by website owners to estimate the number of unique visitors, to detect the most preeminent search engine keywords that lead to a webpage or to track down website navigation issues.
Analytics tools available today use a number of different data collection and analysis models each of which present different data protection risks. A first- party analytic system based on “first party” cookies clearly presents different risks compared to a third-party analytics system based on “third party” cookies.
There are also tools which use “first party” cookies with the analysis performed by another party.
This other party will be considered as a joint controller or as a processor depending on whether it uses the data for its own purposes or if it is prohibited to do so through technical or contractual arrangements.

While they are often considered as a “strictly necessary” tool for website operators, they are not strictly necessary to provide a functionality explicitly requested by the user (or subscriber).
In fact, the user can access all the functionalities provided by the website when such cookies are disabled. As a consequence, <b>these cookies do not fall under the exemption</b> defined in CRITERION A or B.

However, the Working Party considers that first party analytics cookies are not likely to create a privacy risk when they are strictly limited to first party aggregated statistical purposes and when they are used by websites that already provide clear information about these cookies in their privacy policy as well as adequate privacy safeguards.
Such safeguards are expected to include a user friendly mechanism to opt-out from any data collection and comprehensive anonymization mechanisms that are applied to other collected identifiable information such as IP addresses.
```
And being more explicit 
```
First party analytics should be clearly distinguished from third party analytics, which use a common third party cookie to collect navigation information related to users across distinct websites, and which pose a substantially greater risk to privacy.
```

PR #225, opened 2019-06-13
Server refactor

paddyo:
## Description
A proposed refactor that breaks down the main server.ts file into smaller components, which in future could be broken down further. There is no intention to merge this until various significant pieces have been done, and I'll have to bear the merge conflicts for my sins.

## Changes
* `identity` directory now `middleware`, contains middleware components.
* Split out `apiHandler` and its various wrappers into a dedicated middleware file, `apiMiddleware`.
* Various routes have been split into `Routers`, kept in `routes`.
* Add `index.d.ts` file for general type declarations.


mario-galic:
Single routes file can help onboarding as it provides in a sense a defined entry point to the codebase without having to maintain separate endpoint documentation processes. Besides, the blasphemy of having this single inlined file will quickly be offset by modularisation explosion under that layer, which should anyways satisfy urges for perceived successful componentisation.

PR #227, opened 2019-06-24
initial holiday stops functionality for Guardian Weekly

twrichards:
#### Important

- 'responsiveness' will follow in subsequent PR shortly (this version just works for desktop)
- apologies to the reviewers that it's so massive (largely came about from dipping in and out of this piece of work)
- estimated credits and amend/delete functionality will follow in subsequent PRs as the corresponding API features are added
- the 'per year' allowance is provided by the `holiday-stop-api` but the enforcement of that limit and the current level of usage per year (based on anniversary of subscription rather than calendar or financial year) is all done client-side currently - in future this will move to the `holiday-stop-api`
- with this PR as it stands the only way to see the button to enter the holiday stops flow (on Guardian Weekly subs) is to have `showHolidayStops=true` query param...
image
...will show...
image
... beneath a Guardian Weekly sub. 
NOTE: there is a placeholder ready to uncomment which will canary release this feature to roughly 10% of users based on the identity ID.

<hr/>

### Link from Subscriptions tab
image

### Holiday Overview Screen
with no existing holiday stop requests...
image
with some existing holiday stop requests...
image

### Holiday Date Chooser Screen
before any selection...
image
with a valid selection 'this year'...
image
with an invalid selection...
image
with a selection that spans the anniversary date of the subscription...
image

### Holiday Review Screen
image

### Holiday Confirmation Screen
image
Clicking 'Schedule another suspension' will take them back to the Date Chooser Screen but will invoke the new 'ReFetch' functionality of the `AsyncLoader` to load the full list of existing (which should then include the one they've just added.

### Modals
The designs for Holiday Stops included a new feature which was 'modals', so this PR includes that implementation and two uses...
#### What is this date?
image
... will show... 
image
#### Questions? Check here
_this link shows in multiple places throughout the holiday stop flow_
image
... will show...
image

### Accessibility
@tjsilver put a lot of work into ensuring the holiday stops flow (in particular the data range chooser) was accessible from the start. In the interests of getting something deployed, a while back we decided to focus on functionality and revisit 100% accessibility in a future PR. That said the foundations are there with lots of placeholders which just require a bit more work.

### Other Notes
The date range chooser here is a wrapper around [`onefinestay/react-daterange-picker`](https://github.com/onefinestay/react-daterange-picker), with quite a few...

- overridden styles
- overriden/augmented React lifecycle methods
- work arounds for things like 'half day' states (since this library made by a hotel company it makes sense they would have a range end and another range start on the same day and display that as diagonal 'half day') because there were no props to change this behaviour/appearance

... in hindsight it would possibly have been less work to hand craft a date picker.


app/client/components/wizardRouterAdapter.tsx
```diff
@@ -64,22 +75,23 @@ export const ReturnToYourProductButton = (
...
     <PageContainer>
-      {props.routeableStepProps.stepLabels ? (
-        <ProgressBreadcrumb
-          current={props.routeableStepProps.currentStep}
-          labels={props.routeableStepProps.stepLabels}
-        />
-      ) : (
-        <PageContainerSection>
-          <ProgressCounter
+      {!!props.routeableStepProps.currentStep &&
```
arausuy: `!!`? Is this not not or am I just unaware of something?

twrichards: Correct, it is a NOT NOT, and is used to force a truthy (i.e. null/undefined or actual object) into a boolean.

In this use case it forms the opening part of some [React Conditional Rendering](https://reactjs.org/docs/conditional-rendering.html)

app/client/components/hackedDateRangePicker.tsx
```diff
@@ -0,0 +1,207 @@
...
+        .filter(dayCell => dayCell.textContent === dateToAsterisk.format("D"))
+        .forEach(dayCell => dayCell.classList.add(classNameForAddingAsterisk))
+    );
+
+const afterRenderActions = (props: WrappedDateRangePickerProps) => {
+  emboldenTodaysDate();
+  asteriskDate(props.dateToAsterisk);
+};
+
+class HackedDateRangePicker extends DateRangePicker {
```
arausuy: is it really hacked? Maybe it is just customised?

app/client/components/holiday/holidayDateChooser.tsx
```diff
@@ -0,0 +1,396 @@
...
+                        />
+                        <Link css={cancelLinkCss} to=".." replace={true}>
+                          Cancel
+                        </Link>
+                        <div>
+                          <Button
+                            text="Review details"
+                            right
+                            disabled={
+                              !!this.state.validationErrorMessage ||
```
arausuy: not not?

twrichards: yes because disabled expects boolean, so if `this.state.validationErrorMessage` is `undefined` then type checker will complain. as with other example this is just cajoling truthy/falsy into boolean

app/client/components/holiday/holidayQuestionsModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+  >
+    <h3>Things to remember</h3>
+    <ul>
+      <li>
+        You can suspend up to {props.annualIssueLimit} issues in one year.
+      </li>
+      <li>
+        A new suspension cannot begin from today as there is a notice period.
+      </li>
+      <li>Notice period is for our printing and delivery schedule.</li>
```
tjsilver: This reads a little weirdly... "The notice period..." maybe?

arausuy:
Looks great to me, thanks for the run through at desks

PR #228, opened 2019-06-25
Consents (Phase 2) endpoint - basic implementation

GHaberis:
Raising this PR so we don't try and merge everything in one huge pull request:

**Summary of changes in this PR:**

- Adds `get` handler for `/consent` route
- Adds basic implementation of component tree using dummy data...

```JS
<App />
  <ConsentManagementPortal />
    <PrivacySettings />
      <PurposeItem />[]
        <VendorItem />[]
```

- Adds clientside JS bundle entry point: `app/client/consent.tsx`
- Adds serverside HTML template fort consent page: `app/server/consent/html.ts`
- Includes `reset-css` dependency (for resetting browser defaults)
- Includes`babel-preval` dependency for reading the `reset.css` dependency from the filesystem

**What this doesn't include, but will follow in subsequent PRs:**

- Styling, the minimal styles applied so far are going to be replaced
- Font loading and font styles are currently unavailable
- Reading and writing from/to the consent cookie
- Accessibility (ARIA) attributes
- `Raven`/`Sentry` error logging
- `Ophan` integration and audit logging


app/yarn.lock
```diff
@@ -5,7 +5,6 @@
 "@babel/cli@^7.0.0":
   version "7.4.3"
   resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.4.3.tgz#353048551306ff42e5855b788b6ccd9477289774"
-  integrity sha512-cbC5H9iTDV9H7sMxK5rUm18UbdVPNTPqgdzmQAkOUP3YLysgDWLZaysVAfylK49rgTlzL01a6tXyq9rCb3yLhQ==
```
twrichards: the massive negative diff on this file is quite worrying - I'm pretty confident that you used an old version of yarn for this (same thing happened with @tomrf1) I recommend reverting just this file and ensuring you've run `nvm use` in the `app` directory before then running `yarn` to install your newly added dependencies - I would say this is pretty important before merging

app/package.json
```diff
@@ -130,6 +131,7 @@
     "react": "^16.4.1",
     "react-dom": "^16.4.2",
     "react-stripe-elements": "^2.0.1",
+    "reset-css": "^3.0.0",
```
twrichards: are we sure this will only affect the consents stuff and not change any styles on the rest of manage (where they might be relying on browser defaults)?

app/yarn.lock
GHaberis: OK will do, I'm pretty sure this is related to an older version of yarn and I'm happy to update as suggested!

app/package.json
GHaberis: Yep I can't see how it'll affect the rest of the codebase, it's only imported by a module that's unique to `consents`.

PR #232, opened 2019-07-08
dynamic copyright date in footer

twrichards:
@shtukas noticed copyright year was wrong in the footer on manage 😱 - so I've made it dynamic


shtukas:
Awesome. Thanks!

PR #235, opened 2019-07-08
CMP UI - add styles for modal iframe

GHaberis:
This PR adds the initial style rules required for the CMP modal.

Items of note:
- I've written the styles using the string styles technique as [documented](https://emotion.sh/docs/css-prop#string-styles) by `emotion`. I know elsewhere on `manage-frontend` Object styles are preferred, but because this code sits in isolation I felt it was OK to go with an approach which worked for me. I've still be able to use the `breakpoints` and `colours` utility modules used elsewhere on `manage-frontend` as these can work with either the Object/String notation.
- I've introduced a new dependency `desvg-loader`, this is a webpack loader that allows us to import SVG files into importable components.
- I've included an inline script in `html.ts` which means this page will reuse the fonts stored in localStorage by frontend. This uses the `font-loader` endpoint of frontend as documented [here](https://github.com/guardian/frontend/blob/master/docs/06-features-and-components/03-font-loader-route.md).

CMP

It's still using dummy data, doesn't contain any validation and the buttons have no logic wired into them yet.


twrichards:
Not sure if this just a quirk of the GIF but radio button might be clipping the right edge of the inner circle of the radio buttons...
image

GHaberis:
@twrichards this has been on hold for a few weeks whilst new designs are put together, however the vast majority of the work in this PR is still required so I'd like to go ahead and merge once it's approved, thanks!

GHaberis:
@twrichards thanks! We've looked into the clipping, there's a fix that involves altering the radio button's size. We'd need to go back to design to confirm that's OK, but there's a chance once the new designs arrive the radio buttons won't be there, so we decided to leave the change until we know.

PR #238, opened 2019-07-30
eliminated reliance on GU_SIGNIN_EMAIL cookie 

twrichards:
(as this being deprecated/decommed and was a nasty long-standing TODO) 

instead using the `email` provided by the IDAPI `/redirect` endpoint response (`email` field added in https://github.com/guardian/identity/pull/1579)


vlbee:
Late to the party but just wanted to say thanks for rolling this out 👍

PR #242, opened 2019-08-07
Fix Caught Exception Handling on Email Preferences

paddyo:
## Description
Adds missing error handling routines for caught exceptions

### DIFF
--- a/app/client/components/identity/useConsentOptions.ts
+++ b/app/client/components/identity/useConsentOptions.ts
@@ -1,4 +1,6 @@
+import Raven from "raven-js";
 import { useReducer } from "react";
+import { trackEvent } from "../analytics";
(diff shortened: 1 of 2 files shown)


app/client/components/identity/useConsentOptions.ts
```diff
@@ -78,7 +86,7 @@ const reducer = (state: State, action: Action) => {
 };
 
 export const Actions = {
-  error: () => ({ type: ActionType.ERROR }),
+  error: (payload: Error) => ({ type: ActionType.ERROR, payload }),
```
twrichards: if you wanted to use this approach for non-exception error messages the type here could be `string | Error` and `Raven.captureException()` would handle it. Not needed yet though.

PR #243, opened 2019-08-08
Fix missing consents bug on Email and Marketing tab

paddyo:
## Description
It is possible for a user to have no consent information, we suspect these are users using a legacy setup. IDAPI will return no consent information. The moment they submit a consent option, their consent information will be created for them and they will be able to carry on as usual.


paddyo:
Yes I went into code and deleted consent information off a user, the unprotected code crashed as expected, the patch let the page load and clicking a consent caused the consents information to start being returned from that point onwards

PR #248, opened 2019-08-14
Fix minor bug causing APIFetch's last resort error decode to fail

paddyo:
## Description
As a last resort, APIFetch will attempt to decode a response as text if the previous attempt at decoding it as JSON fails. However the fetch response object may only call a decode function once, so a clone is needed.


paddyo:
I'm thinking there's actually no point in doing any JSON conversion as it will be converted back into a string in the error.

twrichards:
@paddyo you gonna merge this? causes the odd Sentry alert

paddyo:
This will be made obsolete by the next migration PR

PR #249, opened 2019-08-15
remove "browserslist-useragent" since we're not really making proper use of it

twrichards:
A **long time back** we added `browserslist-useragent` dependency in-order to be able to interpret our '[browserlist](https://browserl.ist/)' config (in the `package.json` and used by 'babel') and use it at runtime (server-side) to compare against the `User-Agent`, so we could present an 'unsupported browser' banner accordingly - see - see https://github.com/guardian/manage-frontend/pull/101.

Shortly after we observed some mistaken displays of this banner on browsers which were actually supported, so we added some logging (https://github.com/guardian/manage-frontend/pull/102) but ultimately disabled the banner by forcing the `supportedBrowser` boolean to true (https://github.com/guardian/manage-frontend/pull/104) leaving the logging in place to see if we could figure out a pattern to the mistakes... **this was almost a year ago**.

**I don't anticipate being able to diagnose the problems with the `matchesUA` function any time soon and since `browserslist-useragent` and the underlying transitive dependency `useragent` has a Snyk vulnerability it makes sense to remove this stuff for the foreseeable.** _Not to mention it spamming the logs and Sentry._

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -121,7 +121,6 @@
     "@reach/router": "^1.2.1",
     "aws-sdk": "^2.383.0",
     "base-64": "^0.1.0",
-    "browserslist-useragent": "^3.0.0",
     "color": "^3.0.0",
     "express": "^4.16.3",
     "helmet": "^3.12.1",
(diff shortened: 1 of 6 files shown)


arausuy:
If this "feature" is providing zero value and defaulted to off it is essentially dead code.

PR #250, opened 2019-08-15
IAB compliant CMP UI

ripecosta:
This PR adds an IAB compliant CMP UI. This is a work-in-progress. Final designs will be applied when ready.

- Removes the PECR purposes for now
- Makes the CMP UI IAB compliant
- Fetches IAB vendor list from Frontend's endpoint.
- Parses the IAB vendor list and adds the description for each vendor as well as build the UI state
- Lists IAB vendors and and features
- Updates the previous 2 buttons to 4 new buttons
- Adds smooth scrolling for the 'Options' button

Note: Does not add logic to read/save the cookie.

## Screenshot
Screenshot 2019-08-15 at 16 40 59


app/client/components/consent/PrivacySettings.tsx
```diff
@@ -101,58 +66,87 @@ const buttonStyles = css`
...
   constructor(props: {}) {
     super(props);
 
-    this.state = { purposes };
+    this.state = { iabPurposes: {} };
   }
 
   public componentDidMount(): void {
+    fetch(
+      "https://assets.guim.co.uk/data/vendor/4f4a6324c7fe376c17ceb2288a84a076/cmp_vendorlist.json"
```
jeteve: Wrong link. The constant link is:

https://www.theguardian.com/commercial/cmp/vendorlist.json

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -101,58 +66,87 @@ const buttonStyles = css`
...
+    )
+      .then(response => {
+        if (response.ok) {
+          return response.json();
+        } else {
+          throw new Error("Vendor List not ok");
+        }
+      })
+      .then(json => {
+        this.buildState(this.parseVendorList(json));
```
jeteve: At this point, this is not json anymore, just a free formed javascript object.

app/client/components/consent/OnOffRadio.tsx
```diff
@@ -129,4 +128,16 @@ export class OnOffRadio extends Component<Props, {}> {
       </div>
     );
   }
+
+  public shouldComponentUpdate(nextProps: Props): boolean {
+    if (this.props.selectedValue === nextProps.selectedValue) {
+      return false;
+    }
+    return true;
```
ioanna0: could simplify to 
`return this.props.selectedValue !== nextProps.selectedValue`

twrichards: you could make it even cleaner and turn it into an arrow function (which can be an expression and thus avoid the `return` keyword)
```suggestion
    public shouldComponentUpdate = (nextProps: Props) => this.props.selectedValue !== nextProps.selectedValue;
```

ripecosta: 👍

app/client/components/consent/PrivacySettings.tsx
ripecosta: 👍

app/client/components/consent/PrivacySettings.tsx
ripecosta: 👍

ripecosta: I've was trying to use the right link but hit issues with CORS. Let's talk about this tomorrow

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -212,37 +226,252 @@ export class PrivacySettings extends Component<{}, State> {
...
+    const animationStep: FrameRequestCallback = timestamp => {
+      if (!start) {
+        start = timestamp;
+      }
+      // Elapsed milliseconds since start of scrolling.
+      const time: number = timestamp - start;
+      // Get percent of completion in range [0, 1].
+      const percent: number = Math.min(time / duration, 1);
+
+      window.scrollTo(0, startingY + diff * percent);
```
adamnfish: This animation looks well-written from a performance point of view, but it doesn't have any easing. Linear animations look jarring and unnatural, typically an easing function is used so the animation is non-linear. Here's some research from Google's material design (which uses cubic bezier):
https://material.io/design/motion/speed.html#controlling-speed

It's normally better to use CSS transitions. It saves all this code, you get stuff like easing for free, and browsers are really good at optimising them. Since this is doing scroll position I don't think that'll be possible though!

PR #252, opened 2019-08-20
Write IAB cookie from the CMP UI

ripecosta:
This change adds the ability to save the IAB cookie from the CMP UI.


app/client/components/consent/PrivacySettings.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { CmpSeparator } from "./CmpSeparator";
+import { writeVendorConsentCookie } from "./Cookie";
+
+const CMP_ID = 112;
+const CMP_VERSION = 1;
+const CONSENT_SCREEN = 0;
+const CONSENT_LANGUAGE = "en";
 
 const iabVendorListURL =
   "https://assets.guim.co.uk/data/vendor/4f4a6324c7fe376c17ceb2288a84a076/cmp_vendorlist.json";
```
faresite: I know it's not part of this PR but how likely is this to change?

PR #253, opened 2019-08-21
Read IAB cookie from CMP UI

ripecosta:
CMP UI now reads the IAB cookie and updates its state accordingly on load.

This allows up to show the user's choices when resurfacing the UI.

### DIFF
--- a/app/client/components/consent/PrivacySettings.tsx
+++ b/app/client/components/consent/PrivacySettings.tsx
@@ -5,7 +5,7 @@ import palette from "../../colours";
 import { CmpCollapsible } from "./CmpCollapsible";
 import { CmpItem } from "./CmpItem";
 import { CmpSeparator } from "./CmpSeparator";
-import { writeIabCookie } from "./Cookie";
+import { readIabCookie, writeIabCookie } from "./Cookie";
 
 const CMP_ID = 112;
 const CMP_VERSION = 1;
@@ -151,7 +151,6 @@ export class PrivacySettings extends Component<{}, State> {
         // tslint:disable-next-line: no-console
         console.log("ERROR:", error);
       });
-    // TODO: get cookies here
   }
 
   public render(): React.ReactNode {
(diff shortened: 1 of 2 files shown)


GHaberis:
@ripecosta could we rename `Cookie.tsx` to `cookie.ts`? It doesn't contain `jsx` so doesn't need the `.tsx` extension and the uppercase filename is by convention only used when we export a `Class`.

ripecosta:
@GHaberis Of course. Will raise a PR to do just that.

> @ripecosta could we rename `Cookie.tsx` to `cookie.ts`? It doesn't contain `jsx` so doesn't need the `.tsx` extension and the uppercase filename is by convention only used when we export a `Class`.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -285,16 +284,29 @@ export class PrivacySettings extends Component<{}, State> {
   }
 
   private buildState(iabVendorList: ParsedIabVendorList): void {
+    if (iabVendorList && iabVendorList.purposes) {
+      // TODO: Trigger error
+      // tslint:disable-next-line: no-console
+      console.log("ERROR: iabVendorList not present");
```
GHaberis: The console log doesn't reflect what's happening here, it'll log "iabVendorList not present" when `iabVendorList` and `iabVendorList.purposes` are present? Is this `if` block necessary?

twrichards: also we should probably be using Sentry for this sort of error logging

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -285,16 +284,29 @@ export class PrivacySettings extends Component<{}, State> {
...
     this.iabVendorList = iabVendorList;
 
-    if (iabVendorList && iabVendorList.purposes) {
-      const iabPurposes = iabVendorList.purposes.reduce((acc, purpose) => {
+    const iabStr = readIabCookie();
+
+    let iabPurposes = {};
+    if (iabStr) {
+      const iabData = new ConsentString(iabStr);
+      iabPurposes = iabVendorList.purposes.reduce((acc, purpose) => {
```
GHaberis: Question: `iabVendorList` contains the latest list of vendors, but the consent string only contains the vendors at the time it was saved. I assume `iabData.isPurposeAllowed` will return false for any new vendor ID's that aren't present in the iabString? Do we want to cross check any new vendor ID's with the vendor IDs we have saved in the IAB string and if they're not present set them to `true` - as we always force the vendor to `true`?

app/client/components/consent/PrivacySettings.tsx
ripecosta: This was meant to cehck if `iabVendorList` and `iabVendorList.purposes` exist to please the type checker but actually we don't need it because of how we typed the function parameters. Will remove on a separate PR.

GHaberis: @twrichards we have a card to integrate Sentry, these logs provide placeholders for now whilst this is in development but they'll all be removed soon.

PR #254, opened 2019-08-21
enable loading CMP UI in iframe

GHaberis:
This PR makes a couple changes:

- Update the `/consent` route in `server.ts` to remove the `X-Frame-Options` header that prevents routes on `manage.theguardian.com` being loaded as iframes. This only affects the `/consent` route.
- Update `PrivacySettings.tsx` to include a function `close()` that sends a `postMessage` to the parent to instruct it to close the iframe once settings have  been saved or the close button has been clicked.
- Moves `doScrolling` outside the `PrivacySettings` class as it didn't need to be on it.


app/server/server.ts
```diff
@@ -44,7 +44,17 @@ server.use(routes.core);
 server.use("/profile/", routes.profile);
 server.use("/api/", routes.api);
 server.use(routes.productsProvider("/api/"));
-server.use("/consent/", routes.consent);
+
+server.use(
+  "/consent/",
+  (req, res, next) => {
+    // we want this route to be loaded in an iframe
+    res.removeHeader("X-Frame-Options");
```
twrichards: I would highly recommend using [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) rather than `X-Frame-Options` and I would suggest limiting to Guardian domains.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -537,3 +501,40 @@ export class PrivacySettings extends Component<{}, State> {
...
+    window.scrollTo(0, startingY + diff * percent);
+
+    // Proceed with animation as long as we wanted it to.
+    if (time < duration) {
+      window.requestAnimationFrame(animationStep);
+    }
+  };
+
+  // Bootstrap our animation - it will get called right before next frame shall be rendered.
+  window.requestAnimationFrame(animationStep);
```
twrichards: how well does this play with React?

app/server/server.ts
GHaberis: Yep, happy to look at `Content-Security-Policy` - I did begin looking at restricting iframe access to our domains. I believe the `X-Frame-Options` header is added by the `express` security library `helmetjs` that's being used here, I'm not sure this header plays alongside `Content-Security-Policy` so we may still need to remove it.

app/client/components/consent/PrivacySettings.tsx
GHaberis: We've tested it and it seems to behave as expected

app/server/server.ts
GHaberis: @twrichards I've updated to set a `Content-Security-Policy` header.

app/server/server.ts
```diff
@@ -48,8 +48,18 @@ server.use(routes.productsProvider("/api/"));
 server.use(
   "/consent/",
   (req, res, next) => {
-    // we want this route to be loaded in an iframe
+    // This route can be loaded in an iframe from the domains listed below only
+    const allowedDomains = [
```
twrichards: You can probably build a single URL from the conf.DOMAIN constant, that way were not mentioning gulocal.com in the headers of all responses

GHaberis:
OK, thanks for feedback, I believe I've addressed points raised, I've also tested on `CODE` and everything is behaving as expected.

PR #260, opened 2019-08-23
Send message from CMP UI to notify parent consent has been saved

GHaberis:
This PR introduces a new post message emitted from the CMP UI to notify it's `window.parent` consent has been saved. The module [lib/cmp.js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/lib/cmp.js) that currently lives in frontend will listen for this message and upon receiving it will execute any callbacks that have been scheduled to execute when consent changes.

The corresponding PR that handles these events is here: https://github.com/guardian/frontend/pull/21766

### DIFF
--- a/app/server/consent/html.ts
+++ b/app/server/consent/html.ts
@@ -82,8 +82,6 @@ const html: (
                 e.data.fonts.forEach(useFont);
               }
             });
-
-            document.body.appendChild(iframe);
           };
 
           document.addEventListener('DOMContentLoaded', loadFonts);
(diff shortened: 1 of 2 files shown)


app/server/consent/html.ts
```diff
@@ -82,8 +82,6 @@ const html: (
                 e.data.fonts.forEach(useFont);
               }
             });
-
-            document.body.appendChild(iframe);
```
GHaberis: This line of font loading logic wasn't required (it was appending an iframe to the document body that already existed in the body), I noticed this error because I saw 2 network requests for the `font-loading` endpoint in the Nework panel.

PR #261, opened 2019-08-27
Add PECR purposes

ripecosta:
This PR adds the 3 PECR purposes (Essential, Functional and Performance) to the CMP.

This version doesn't use the final version of designs (yet to be created) for the UI so it's not visually polished. It only aims at adding the underlying logic so the designs are easily added when they become available.

## Screenshot
Screenshot 2019-08-27 at 16 23 02

### DIFF
--- a/app/client/components/consent/PrivacySettings.tsx
+++ b/app/client/components/consent/PrivacySettings.tsx
@@ -88,27 +88,58 @@ const buttonStyles = css`
   margin: 0 6px;
 `;
 
+const integStyles = css`
+  margin-right: 5px;
+  border: none;
+  border-radius: 15px;
+  padding: 5px 10px;
+  background-color: ${palette.neutral[6]};
+`;
+
 const cleanGuPurposeList: GuPurposeList = {
   purposes: [
     {
       id: 1,
       name: "Essential",
       description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
-      integrations: []
+      integrations: [
+        {
+          name: "Ophan",
+          policyUrl: "https://www.theguardian.com/info/privacy"
+        },
+        {
+          name: "Confiant",
+          policyUrl: "https://www.confiant.com/privacy"
+        }
+      ]
     },
     {
       id: 2,
       name: "Functional",
       description:
         "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque quis malesuada ante.",
-      integrations: []
+      integrations: [
+        {
+          name: "Pinterest",
+          policyUrl: "https://policy.pinterest.com/"
+        }
+      ]
     },
     {
       id: 3,
       name: "Performance",
       description:
         "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque quis malesuada ante. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos.",
-      integrations: []
+      integrations: [
+        {
+          name: "Sentry",
+          policyUrl: "https://sentry.io/privacy/"
+        },
+        {
+          name: "Google Analytics",
+          policyUrl: "https://policies.google.com/privacy?hl=en-US"
+        }
+      ]
     }
   ]
 };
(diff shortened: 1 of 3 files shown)


app/client/components/consent/PrivacySettings.tsx
```diff
@@ -82,17 +87,75 @@ const buttonStyles = css`
   margin: 0 6px;
 `;
 
+const integStyles = css`
+  margin-right: 5px;
+  border: none;
+  border-radius: 15px;
```
GHaberis: I know these styles are very likely to be pulled out, but they are half making an anchor look like a button - I say half as the anchor still has a visited link colour and an underline even though it appears to be a button - plus the border-radius is different to the actual buttons - it's not a blocker for me as these will 100% change but I want to raise my concern at introducing multiple differing button-like styles.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -82,17 +87,75 @@ const buttonStyles = css`
...
+  padding: 5px 10px;
+  background-color: ${palette.neutral[6]};
+`;
+
+const cleanGuPurposeList: GuPurposeList = {
```
GHaberis: Any reason we're naming this `cleanGuPurposeList` and not just `guPurposeList`? It got me looking for a (non-existant) `dirtyGuPurposeList` and the code to clean it.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -212,10 +280,26 @@ export class PrivacySettings extends Component<{}, State> {
     );
   }
 
-  private buildState(iabVendorList: ParsedIabVendorList): Promise<void> {
+  private buildState(
+    guPurposeList: ParsedGuPurposeList,
```
GHaberis: OK I asked about  the naming of `cleanGuPurposeList` above, I'm guessing that was done so we could name this parameter `guPurposeList`.

app/client/components/consent/PrivacySettings.tsx
GHaberis: I see why you've done this as `guPurposeList` is used elsewhere, once `cleanGuPurposeList` has been parsed.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -212,10 +280,26 @@ export class PrivacySettings extends Component<{}, State> {
...
+    iabVendorList: ParsedIabVendorList
+  ): Promise<void> {
+    // tslint:disable-next-line: no-object-mutation
+    this.guPurposeList = guPurposeList;
     // tslint:disable-next-line: no-object-mutation
     this.iabVendorList = iabVendorList;
 
+    const guState = readGuCookie();
+
+    let guPurposes = {};
```
GHaberis: You could lose the `let` and `guState` here by doing:

```js
const guPurposes = readGuCookie() || guPurposeList.purposes.reduce((acc, purpose) => {
    return {
        ...acc,
        [purpose.id]: !purpose.id ? true : null
    };
}, {});
```

ripecosta: Good point!

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -232,25 +314,55 @@ export class PrivacySettings extends Component<{}, State> {
...
+  private renderGuPurposeItems(): React.ReactNode {
+    if (!this.guPurposeList || !this.guPurposeList.purposes) {
+      return "";
+    }
+
+    return this.guPurposeList.purposes.map(
+      (purpose: ParsedGuPurpose): React.ReactNode => {
+        const { id, name, description, integDescription } = purpose;
+
+        // Remove radio button for Essential
```
GHaberis: I like your use of `optProps` but I'm not totally sure about the `id === 0` check, I know the comment above explains what's going on but comments are easy to remove. I'd  much rather a more descriptive variable name was used in place of  `id === 0` to  demonstrate what's going on here. Either `const  hideRadio = id === 0` or  stick a boolean `hideRadio` on the `purpose`.

ripecosta: 👍

PR #266, opened 2019-09-06
canary launch GW holiday stops to  approx. 10% of users (based on their identity ID)

twrichards:
This PR displays the entry button into the holiday stops flow for Guardian Weekly, on the subscriptions tab, for approx 10% of users 🚀.

This approx 10% is determined by the users identity ID, specifically if their identity ID ends in `2`. Using this approach for canary release is to ensure that the same customers who initially see this functionality always have access to it, including across devices - which isn't guaranteed when using other approaches.

#### Related PRs
Main PR - https://github.com/guardian/manage-frontend/pull/227
Responsiveness PR - https://github.com/guardian/manage-frontend/pull/267

### DIFF
--- a/app/client/components/productPage.tsx
+++ b/app/client/components/productPage.tsx
@@ -391,10 +391,11 @@ const getProductDetailRenderer = (
                   />
                 </a>
               ))}
-            {shouldHaveHolidayStopsFlow(
-              productType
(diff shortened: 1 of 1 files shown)


arausuy:
I'm good with this!

PR #274, opened 2019-09-16
holiday-stops: use V2 /potential endpoint of holiday-stop-api

twrichards:
Refactor to call V2 of the `/potential` endpoint of `holiday-stop-api` which now takes subscription name as a path parameter. This enables the transitionary code & naming in that lambda to be eliminated (see https://github.com/guardian/support-service-lambdas/pull/413).

PLUS couple of quick tweaks...

- replaced instances of `map()` followed by `flat()`, with `flatMap()` - for better browser compatibility
- minor UI tweak for /potential spinner on larger mobiles to prevent jank

### follow-up PR will actually display these estimated credits throughout the flow

### DIFF
--- a/app/server/holidayStopApiHandlers.ts
+++ b/app/server/holidayStopApiHandlers.ts
@@ -16,16 +16,17 @@ export const getHolidayStopApiHandler = (
         const hsrEnvConfig =
           testUserHeader === "true" ? hsrConfig.testMode : hsrConfig.normalMode;
         const isPotentialCall = req.params.sfId === "potential";
-        const maybeActualExistingSfId = !isPotentialCall
-          ? req.params.sfId
-          : undefined;
+        const basePathPart = isPotentialCall ? "potential" : "hsr";
+        const maybeActualExistingSfId = isPotentialCall
+          ? undefined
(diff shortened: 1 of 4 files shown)


app/server/holidayStopApiHandlers.ts
```diff
@@ -15,19 +15,20 @@ export const getHolidayStopApiHandler = (
...
-              (req.params && req.params.subscriptionName
-                ? req.params.subscriptionName === "potential"
-                  ? "/potential"
-                  : `/hsr/${req.params.subscriptionName}`
-                : "/hsr") +
-              (req.params && req.params.subscriptionName && req.params.sfId
-                ? `/${req.params.sfId}`
-                : ""),
+              req.params && req.params.subscriptionName
+                ? `/${isPotentialCall ? "potential" : "hsr"}/${
```
mario-galic: Terinary within terniary.

twrichards: ahh yes, good catch. fixed in https://github.com/guardian/manage-frontend/pull/274/commits/1c970a404c20dad91dbaa54341d3e0b885ffc52b

PR #278, opened 2019-09-24
Introduce new styles to the  CMP UI

GHaberis:
This PR begins the restyling work required for the CMP UI. A couple of changes to note:

- We're now using the `"@guardian/src-foundations` dependency which exports a number of useful/common design tokens common across the Guardian's digital products. We're using it for CSS properties such as colours from the `palette` and margins/padding from `spacing`.
- The Styles of the buttons and radio buttons have been copied from the https://github.com/guardian/source-components repo. It is anticipated that these will eventually be replaced by components imported directly from this library.
- We've had to update the webpack config to copy files from an `app/images` directory to `app/dist/static/images` when building the distributed assets, we do this so we can use `/static/images/consent-graphic.png`.


app/client/components/consent/OnOffRadio.tsx
```diff
@@ -1,86 +1,99 @@
...
+    Lucida Grande, sans-serif;
+  font-size: 17px;
+  color: ${disabled ? palette.brand.pastel : palette.neutral[100]};
+`;
 
 interface Props {
-  onChangeHandler: (value: boolean) => void;
-  selectedValue: ItemState;
+  onChangeHandler?: (value: boolean) => void;
+  selectedValue?: ItemState;
```
adamnfish: Is it true that these are both optional? Would an empty object be a valid `Props`?

ripecosta: Yes, both are optional. If one of them is missing we render the radio button as disabled.

PR #276, opened 2019-09-18
Migrate Account Details Tab to Manage Frontend

paddyo:
# Migrate Account Details Tab To Manage Frontend

## Description
Final tab to migrated from the frontend/identity project to the manage-frontend project. This is made with the links to the page missing so the feature can be activated in a much smaller pull request.

## Changes
* Add an `AccountDetails` page component and routing.
* Add an `AccountDetailsFormSection` for dealing with the account form.
* Add `phonenumber` idapi adapter for deleting phone numbers.
* Add `FormField`, `FormTextField`, `FormTextAreaField` and `FormSelectField` for wrapping Formik components. Existing Profile form updated to use these fields.
* Add a margin prop to the `Lines` component to allow custom definition of the margin.
* Allow `PageSection` to take a `ReactNode` as its description prop, to allow richer descriptions, such as including a link.
* Add missing top level error checking on the Profile form.
* Fix incorrect margins on the page description part of the Profile form.
* Add new fields to the `User model`
* Add CRSF utility function for fetching

## Dependencies
* #285 
* guardian/identity#1641 (AND with role attached to manage-frontend client access token)
* guardian/identity#1642

PR #285, opened 2019-10-02
Add user read proxy endpoint

paddyo:
## Description
Due to the way the `idapi` `/users/me` endpoint works, a proxy endpoint must be made for reading and writing to the `privateFields` section of the `user` object.

## Changes
* Add new `idapi` router for transformative requests to idapi
* Add `/idapi/user` endpoint that is capable of reading and writing user attributes allowed by the identity client token assigned to `manage-frontend`.
* Add CSURF for CRSF protection on the `/idapi/user` endpoints directly. Token must got from a User read first.
* Updated existing code referencing `/user/me` to use `idapi/user` instead.

## Dependents
* #276


twrichards:
what is the client-side bundle size impact? (if any)

paddyo:
There shouldn't be any significant bundle impact. The only library used client side is the cookie one, but that was introduced in previous work

PR #290, opened 2019-10-07
Move responsibility to save Cookie and post to Consent Log out of CMP UI

GHaberis:
This PR moves the responsibility to save the consent Cookie and post to the Consent Logs out of CMP UI and into `@guardian/consent-management-platform`. 

`@guardian/consent-management-platform` now accepts all the data required to save the cookie and post to the logs as part of the `CMP_SAVED_MSG` post message that originates from the CMP UI. We've therefore had to update the format of postMessages originating from the CMP UI, from `string` to objects, eg: 

```js
{
msgType: CMP_READY_MSG,
msgData: { .. }
}
```

More details on why we've moved these responsibilities out of the CMP UI can be found here: https://github.com/guardian/consent-management-platform/pull/37

This PR also fixed a bug when evaluating whether `isProd`, it was previously resolving to false in PROD, we now use the standard used elsewhere in `manage-frontend` to determine whether `isProd`.


app/package.json
```diff
@@ -121,18 +121,16 @@
   "dependencies": {
     "@emotion/core": "^10.0.14",
     "@emotion/styled": "^10.0.14",
-    "@guardian/consent-management-platform": "^0.3.0",
+    "@guardian/consent-management-platform": "1.0.0-beta.5",
```
GHaberis: This will be updated to `1.0.0` once it's been published.

app/client/components/consent/PrivacySettings.tsx
```diff
@@ -33,11 +32,15 @@ const PURPOSES_ID = "purposes";
 const SCROLLABLE_ID = "scrollable";
 const HEADER_ID = "header";
 
-const isProd = conf.DOMAIN === "theguardian.com";
+let domain: string;
```
adamnfish: Be easy to write this as a ternary operator to avoid the mutability, but this is a *super* minor point, no stressing at all.

PR #301, opened 2019-10-15
Fix Invariant Violation error

ripecosta:
Sentry has been reporting Invariant Violation errors ("Target container is not a DOM element") which we believe are being caused by calling `document.getElementById("app")` inside `run()` before the body is parsed.

Despite not being able to recreate the issue locally we believe this change will fix it.

Tested locally and on CODE to make sure there are no adverse effects.

### DIFF
--- a/app/client/consent.tsx
+++ b/app/client/consent.tsx
@@ -34,4 +34,10 @@ const run = (): void => {
   }
 };
 
-run();
+if (document.readyState !== 'loading') {
+  run();
+} else {
+  document.addEventListener('DOMContentLoaded', function () {
+    run();
+  });
+}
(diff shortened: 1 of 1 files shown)


app/client/consent.tsx
```diff
@@ -34,4 +34,10 @@ const run = (): void => {
   }
 };
 
-run();
+if (document.readyState !== 'loading') {
+  run();
+} else {
+  document.addEventListener('DOMContentLoaded', (): void => {
```
adamnfish: Good solution.

adamnfish: I'd consider also adding a note about this to the comment at the top of the run function's definition.

PR #311, opened 2019-10-24
Activate new account tab

paddyo:
## Description
Changes the navigation links to point to the new account tab. This essentially releases the feature on manage frontend.

## Dependencies
* #276

### DIFF
--- a/app/client/components/userNav.tsx
+++ b/app/client/components/userNav.tsx
@@ -101,7 +101,7 @@ export class UserNav extends React.Component {
     },
     {
       title: "Account details",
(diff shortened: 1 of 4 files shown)


app/client/components/userNav.tsx
```diff
@@ -101,7 +101,7 @@ export class UserNav extends React.Component {
     },
     {
       title: "Account details",
-      link: `${profileHostName}/account/edit`
+      link: `/account-details`
```
dominickendrick: Does this need backticks if its not doing any string interpolation any more ?

PR #319, opened 2019-11-13
prevent creation of holiday stops on non-auto-renewing subs

twrichards:
...with explainer message, although if there are any existing then it still displays those (we know of approx. 20 non-auto-renewing subs where users created holiday stops via self-service prior to this change).

### Holiday Overview screen for non-auto-renewing sub

|  ...with existing holiday stops | ...without existing holiday stops | 
| --- | --- |
| image | image |

... **NOTE**

- subtly different explainer ("cannot" vs "can no longer")
- no `Create Suspension` button on either


twrichards:
image
The actual change is only a few lines, most of this is just indentation since parts of the page are now wrapped in conditional rendering based on `productDetail.subscription.autoRenew` boolean.

PR #321, opened 2019-11-19
better message(s) when users go direct to a flow on a cancelled sub

twrichards:
If customers have multiple subscriptions of a given type (e.g. `guardianweekly`) and they enter any of the flows directly (for example enter the holiday stops flow direct from an email or FAQ page) then they're presented with a list of subs of that type so they can select the one they want to proceed with. _Note that if they come from the product tab (e.g. subscriptions) then it uses the 'browser history state' to denote the desired one to proceed with._

**Previously, 'cancelled' subs would be filtered out of this list**, likewise if the customer had just one subscription of that type which was cancelled and entered the flow directly they would just see the generic `NoProduct` page (which states for example `You do not currently have a newspaper voucher subscription.`) which is kind of misleading.

So now 'cancelled' subs are no longer filtered out of this list, and instead marked with a little `CANCELLED` badge...
image

If the user proceeds with a cancelled sub (or if they only have one of that type and so don't go via the list) they see the following message...

#### Holiday Stops Flow _(GW & Voucher only currently)_
image

#### Payment Update Flow
image

#### Cancellation Flow _(Membership & Contributions only currently)_
image

### DIFF
--- a/app/client/components/cancel/cancellationFlow.tsx
+++ b/app/client/components/cancel/cancellationFlow.tsx
@@ -140,6 +140,9 @@ export const CancellationFlow = (
     headingPrefix="Cancel"
     supportRefererSuffix="cancellation_flow"
     loadingMessagePrefix="Checking the status of your"
+    cancelledExplainer={`This ${
+      props.productType.friendlyName
+    } is already cancelled.`}
     singleProductDetailRenderer={reasonsRenderer(props)}
   />
 );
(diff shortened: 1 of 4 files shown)


JessL123:
@twrichards Could we say:
HS:  "This [product] has been cancelled. Any scheduled holiday suspensions have been removed. Please contact us if you would like to re-start your [product], make any amendments or need further help."  
Payment update: "This [product] has been cancelled. Please contact us if you would like to re-start your [product], make any amendments or need further help."  
Cancellation: "This [product] has been cancelled. Please contact us if you would like to re-start your [product], make any amendments or need further help."

jessroseking:
Agree with Jess L

PR #322, opened 2019-11-19
Adds a referer to IDAPI requests 

vlbee:
Adds a referer to IDAPI requests coming from manage-frontend, to facilitate on going security audit of IDAPI. Can be reverted once complete.


vlbee:
@twrichards  
We were going to use this to help track the source of requests to IDAPI but might have an alternative and not need to add this anymore. I will hold off merging and close the PR if so, otherwise I will alert you before merging. If we need this, it will only be a for a day or so.

PR #325, opened 2019-11-20
hide holiday stops row on product page for non-auto renewing subs

twrichards:
Follows on from https://github.com/guardian/manage-frontend/pull/319 - this hides holiday stops row on product page for to avoid disappointment when they then find out holiday stops are not supported for non-auto renewing subs.


twrichards:
✅ TESTED in CODE
@rBangay @frj please could I get a review for this small PR, if you select `Hide whitespace Changes` the diff is only image

image

PR #335, opened 2019-11-28
upgrade to Typescript 3.7 🎉🎉🎉

twrichards:
PLUS

- Babel & Prettier to latest
- add @babel/plugin-proposal-optional-chaining 
- refactor one file to use optional chaining 🎉
- one tiny tweak as TS has tightened something up


app/client/components/holiday/holidaysOverview.tsx
```diff
@@ -116,7 +116,7 @@ const renderHolidayStopsOverview = (
       value={{
         ...holidayStopsResponse,
         reload: reloadWhichAlsoClearsAnyExistingHolidayStopToAmend,
-        existingHolidayStopToAmend
+        existingHolidayStopToAmend: existingHolidayStopToAmend || undefined
```
twrichards: the type required is `HolidayStopRequest | undefined` but `HolidayStopRequest | null` is supplied. Pre TypeScript upgrade this was allowed (as it has virtually the same effect) however TypeScript have tightened something up, so this just coerces `null` into `undefined`

PR #337, opened 2019-12-03
renaming and restructuring, in anticipation of consuming `fulfilment-date-calculator`

twrichards:
Shortly we will need to be augmenting the `members-data-api` response with `deliveryAddressChangeEffectiveDate` from `fulfilment-date-calculator`, so making some preparatory changes to keep that PR easier to digest.

#### Changes
- move `isTestUser` annotation of the `members-data-api` response to the server-side 
- various renaming and restructuring

### DIFF
--- a/app/server/middleware/apiMiddleware.ts
+++ b/app/server/middleware/apiMiddleware.ts
@@ -12,13 +12,13 @@ import {
 
 export type BodyHandler = (res: Response, body: string) => void;
 
-const straightThroughBodyHandler: BodyHandler = (res, jsonString) =>
-  res.send(jsonString);
+export const straightThroughBodyHandler: BodyHandler = (res, body) =>
+  res.send(body);
 
 export const proxyApiHandler = (
   basePath: string,
   ...headersToForward: string[]
-) => (bodyHandler: BodyHandler = straightThroughBodyHandler) => (
+) => (bodyHandler: BodyHandler) => (
   path: string,
   forwardQueryArgs?: boolean,
   ...pathParamNamesToReplace: string[]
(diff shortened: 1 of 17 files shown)


app/server/middleware/apiMiddleware.ts
```diff
@@ -10,12 +10,15 @@ import {
   getCookiesOrEmptyString
 } from "./identityMiddleware";
 
-export type JsonHandler = (res: Response, jsonString: string) => void;
+export type BodyHandler = (res: Response, body: string) => void;
 
-export const apiHandler = (jsonHandler: JsonHandler) => (
+const straightThroughBodyHandler: BodyHandler = (res, jsonString) =>
```
rBangay:
```suggestion
const straightThroughBodyHandler: BodyHandler = (res, body) =>
```

rBangay:
```suggestion
const straightThroughBodyHandler: BodyHandler = (res, body) =>
```

app/server/middleware/apiMiddleware.ts
```diff
@@ -10,12 +10,15 @@ import {
...
+  res.send(jsonString);
```
rBangay:
```suggestion
  res.send(body);
```

app/server/middleware/apiMiddleware.ts
```diff
@@ -10,12 +10,15 @@ import {
...
+
+export const proxyApiHandler = (
   basePath: string,
   ...headersToForward: string[]
-) => (
+) => (bodyHandler: BodyHandler = straightThroughBodyHandler) => (
```
rBangay: As discussed earlier can we remove the default param so that the calls are more explicit

PR #338, opened 2019-12-05
'deliveryAddressChangeEffectiveDate' now available on 'members-data-api' mma response

twrichards:
_Preparatory refactor PR : https://github.com/guardian/manage-frontend/pull/337
Also relies on https://github.com/guardian/members-data-api/pull/418_

When proxying the `mma` endpoint of `members-data-api` before sending on the response to the client, in addition to adding the `isTestUser` (see https://github.com/guardian/manage-frontend/pull/337) on **each** item (i.e. each subscription) in the array it now adds a `deliveryAddressChangeEffectiveDate` on each response **IF**...

1. ... the `ProductType` inferred from the item/subscription has a `fulfilmentDateCalculatorProductFilenamePart` defined

2. AND there is a file **for today** in the `fulfilment-date-calculator-STAGE` S3 bucket **for that product**
_(it logs and fires a Sentry error if there isn't such a file but one is expected because of condition 1. above.)_

### DIFF
--- a/app/server/fulfilmentDateCalculatorReader.ts
+++ b/app/server/fulfilmentDateCalculatorReader.ts
@@ -1,7 +1,9 @@
+import Raven from "raven-js";
 import { ProductDetail } from "../shared/productResponse";
 import { ProductTypes } from "../shared/productTypes";
 import { s3FilePromise } from "./awsIntegration";
 import { conf } from "./config";
+import { log } from "./log";
 
 interface RawFulfilmentDateCalculatorDates {
   today: string;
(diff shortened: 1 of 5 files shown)


app/server/fulfilmentDateCalculatorReader.ts
```diff
@@ -0,0 +1,83 @@
...
+        datesByDayOfWeek[dayOfWeek].deliveryAddressChangeEffectiveDate;
+      if (
+        !earliestAcc ||
+        !deliveryAddressChangeEffectiveDateStr ||
+        new Date(earliestAcc) > new Date(deliveryAddressChangeEffectiveDateStr)
+      ) {
+        return deliveryAddressChangeEffectiveDateStr || "";
+      } else {
+        return earliestAcc;
+      }
```
rBangay: do you need the else here? could you just have:

`return earliestAcc;`

instead as that will only be called if the above if condition is not met

twrichards: done - https://github.com/guardian/manage-frontend/pull/338/commits/4f6df64f00b10e50b08ad35e711fc95959cfd3d0

PR #341, opened 2019-12-19
enable 'Home Delivery' Holiday Stops

twrichards:
This enables the Holiday Stops flow for 'Home Delivery' - 3 line change 🎉 

image

It uses `paper` for the issue keyword...
image
image


twrichards:
@JessL123 please can you confirm that you're happy with with `paper` as the issue keyword, the screenshots show some examples of where that comes out in the copy, one important one is the issue count on the right.

(FYI  the issue keyword is `voucher` for voucher subs)

PR #349, opened 2020-01-03
BUG - fix home delivery postcode validation

rBangay:
Fix issue where people without home delivery subscriptions were seeing the M25 area postcode validation error message.

M25validationError

### DIFF
--- a/app/client/components/delivery/address/formValidation.tsx
+++ b/app/client/components/delivery/address/formValidation.tsx
@@ -1,6 +1,6 @@
 import { DeliveryAddress } from "../../../../shared/productResponse";
-import { ukPhoneNumberWithoutPrefix } from "../../callCentreNumbers";
 import { ProductTypes } from "../../../../shared/productTypes";
+import { ukPhoneNumberWithoutPrefix } from "../../callCentreNumbers";
 
 interface ErrorState {
   isValid: boolean;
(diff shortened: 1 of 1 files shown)


app/client/components/delivery/address/formValidation.tsx
```diff
@@ -30,16 +30,18 @@ export const isFormValid = (
   const postcodeEnteredCheck =
     formData.postcode.length > 0 && formData.postcode.length < 20;
 
-  const withinM25Check =
-    subscriptionsNames.includes("home delivery subscription") &&
-    postcodeEnteredCheck &&
-    !isPostcodeOptional("GB") &&
-    isHomeDeliveryInM25(formData.postcode);
+  const withinM25Check = () => {
```
twrichards: not sure this needs to be a function

app/client/components/delivery/address/formValidation.tsx
```diff
@@ -30,16 +30,18 @@ export const isFormValid = (
...
+    return !subscriptionsNames.includes("home delivery subscription")
+      ? true
```
twrichards: this ternary can be eliminated with an `||`

app/client/components/delivery/address/formValidation.tsx
twrichards: this name should probably have something like `IfApplicable` on the end

app/client/components/delivery/address/formValidation.tsx
```diff
@@ -30,16 +30,18 @@ export const isFormValid = (
...
+      : postcodeEnteredCheck &&
+          !isPostcodeOptional("GB") &&
+          isHomeDeliveryInM25(formData.postcode);
```
twrichards: this function name is a touch confusing, if all it's checking is whether the postcode is in M25 the function name shouldn't mention home delivery since that part of the check is happening elsewhere

app/client/components/delivery/address/formValidation.tsx
twrichards: also the lookup string could be onto `ProductTypes`

PR #351, opened 2020-01-06
API Gateway Lambdas service discovery rather than config in S3

twrichards:
`manage-frontend` now looks up _host_ and _api-keys_ **by _name_ and _stage_** using `aws-sdk` rather than storing those as config in S3, for the following API Gateway Lambdas...

- [`holiday-stop-api` ](https://github.com/guardian/support-service-lambdas/tree/master/handlers/holiday-stop-api)
- [`delivery-records-api`](https://github.com/guardian/support-service-lambdas/tree/master/handlers/delivery-records-api)
- [`cancellation-sf-cases-api`](https://github.com/guardian/support-service-lambdas/tree/master/handlers/cancellation-sf-cases-api)

Rather than using the `S3` client to find manually maintained config files it now uses the `CloudFormation` and `APIGateway` clients to find the _host_ and _api-keys_ **by _name_ and _stage_** and selects the correct one based on whether the request is from a `test-user`, also adding the `x-identity-id` header using the value which came from IDAPI.

**See [wiki/Proxying-API-Gateway-Lambdas](https://github.com/guardian/manage-frontend/wiki/Proxying-API-Gateway-Lambdas) for more information.**

### MUST be released after https://github.com/guardian/support-service-lambdas/pull/540
... as this makes `cancellation-sf-cases-api` work like the others.

### Manual config deletion will follow release to PROD
The config in S3 that is eliminated by this PR will be cleaned up manually.


mario-galic:
Are return types on exported interfaces not a convention in TS?

app/server/apiGatewayHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+  stackResourceSummaries?: StackResourceSummaries
+) => {
+  const apiKeyPromises = stackResourceSummaries
+    ?.filter(byResourceType("AWS::ApiGateway::ApiKey"))
+    .map(toDefinedPhysicalResourceId(stackName))
+    .map(lookupApiKey);
+  if (apiKeyPromises && apiKeyPromises.length === 1) {
+    return apiKeyPromises[0];
+  }
+  log.error(
```
mario-galic: Why is `log.error` sufficient instead of  `throw`?

app/server/apiGatewayHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+    testUserApiStage
+  );
+
+  return async (req: express.Request, res: express.Response) => {
+    const isTestUser = req.header(MDA_TEST_USER_HEADER) === "true";
+    const { host, apiKey } = (await (isTestUser
+      ? testModeConfigPromise
+      : normalModeConfigPromise)) as any;
+    const stage = isTestUser ? testUserApiStage : normalUserApiStage;
+    if (!apiKey) {
```
mario-galic: Ah I see why, it depends on null for error handling.

app/server/apiGatewayHandler.ts
twrichards: this would just log the error, at launch (something like aws-sdk permission issues or no stack with that name existing).
We don't want to block the start-up of the entire of manage just because we can't get the host/api key for a service.

app/server/routes/api.ts
```diff
@@ -5,14 +5,14 @@ import {
   MDA_TEST_USER_HEADER,
   MembersDataApiItem
 } from "../../shared/productResponse";
+import { apiGatewayHandler } from "../apiGatewayHandler";
```
mario-galic: IMO, handler is a meaningless name which does not convey intention. How about `discoverAuthorizedResource` or something which conveys host+key intention.

app/server/apiGatewayHandler.ts
twrichards: yes we serve 500 for these services if we dont have the host or api key to call them, we'd then catch these with alarms.

app/server/routes/api.ts
```diff
@@ -91,11 +91,36 @@ router.post(
...
+    "cancellation-sf-cases-api",
+    pathParams => `/case/${pathParams.caseId || ""}`
+  )
+);
 
-router.patch(
-  "/case/:caseId",
-  sfCasesApiHandler("case/:caseId", false, "caseId")
+router.use(
+  "/holidays/:subscriptionName?/:sfId?",
```
mario-galic: Is it necessary to have logic in the router?

app/server/apiGatewayDiscovery.ts
```diff
@@ -0,0 +1,164 @@
...
+  const testModeConfigPromise = getHostAndApiKeyForStack(
+    apiName,
+    testUserApiStage
+  );
+
+  return async (req: express.Request, res: express.Response) => {
+    const isTestUser = req.header(MDA_TEST_USER_HEADER) === "true";
+    const { host, apiKey } = (await (isTestUser
+      ? testModeConfigPromise
+      : normalModeConfigPromise)) as any;
```
rBangay: can this be typed without using any?

twrichards: refactored to avoid this 😺

twrichards:
deleted the redundant `DEV` and `CODE` config files from `gu-reader-revenue-private` bucket

twrichards:
deleted the redundant `PROD` config files from `gu-reader-revenue-private` bucket - after a night of running without issue

app/server/routes/api.ts
twrichards: eliminated in https://github.com/guardian/manage-frontend/pull/373#discussion_r396399572

twrichards: eliminated in https://github.com/guardian/manage-frontend/pull/373/files#r396404506

twrichards: eliminated in https://github.com/guardian/manage-frontend/pull/373#discussion_r396404506

PR #353, opened 2020-01-13
Display delivery records

rBangay:
Self service display of delivery records.

[ticket](https://trello.com/c/OoKbySUs/641-goal-self-service-display-of-delivery-records-and-why-they-may-not-have-been-due-a-delivery)

#### Flow
01-viewDeliveryRecordsBtn

02-deliveryRecordsTableMobile

03-deliveryRecordsDesktop

##### delivery record address change
justAddressChange

##### delivery record instructions change
justInstructions

##### delivery record address and instructions change
withDeliveryandInstructions


app/client/components/delivery/records/deliveryRecordsTable.tsx
```diff
@@ -0,0 +1,251 @@
...
+  resultsPerPage: number;
+}
+
+export const RecordsTable = (props: RecordsTableProps) => {
+  const totalPages = Math.ceil(props.data.length / props.resultsPerPage);
+  const [currentPage, setCurrentPage] = useState(0);
+  const scrollToTop = () => {
+    window.scrollTo(0, 0);
+  };
+  const isOdd = (n: number) => Math.abs(n % 2) === 1;
```
twrichards: could this be done with css https://www.w3.org/Style/Examples/007/evenodd.en.html (probably with `nth-of-type` though to avoid react console warnings)

app/client/components/delivery/records/deliveryRecordsTable.tsx
```diff
@@ -0,0 +1,251 @@
...
+            <th>Date</th>
+            <th>Status</th>
+            <th>Delivery postcode</th>
+            <th>Delivery instructions</th>
+          </tr>
+        </thead>
+        <tbody css={tbodyCSS}>
+          {props.data
+            .filter(
+              (element, index) =>
```
twrichards: could we pull this function into a well-named `const` so it's easier to understand

app/client/components/flowStartMultipleProductDetailHandler.tsx
```diff
@@ -23,25 +23,32 @@ import { minWidth } from "../styles/breakpoints";
...
-  if (subscription.card) {
-    return <CardDisplay margin="0" {...subscription.card} inline />;
-  } else if (subscription.mandate) {
-    return <DirectDebitDisplay {...subscription.mandate} inline />;
-  } else if (subscription.payPalEmail) {
-    return <span>via PayPal</span>;
-  }
-  return null;
-};
+type NewLayoutWrapperFunc = (children: ReactElement | null) => ReactElement;
```
twrichards: this type could probably be inlined in the one place it's referenced

app/client/components/flowStartMultipleProductDetailHandler.tsx
```diff
@@ -23,25 +23,32 @@ import { minWidth } from "../styles/breakpoints";
...
+interface ConditionalNewLayoutWrapperProps {
+  condition: boolean | undefined;
+  wrapper: NewLayoutWrapperFunc;
+  children: ReactElement | null;
+}
+const ConditionalNewLayoutWrapper = ({
```
twrichards: please can we avoid the word 'new' and instead use something the new style of wrapper - 'WithLeftNav' or something

app/client/components/flowStartMultipleProductDetailHandler.tsx
```diff
@@ -97,102 +115,122 @@ const getProductDetailSelector = (
...
-                        background: palette.neutral["5"],
-                        marginRight: "10px",
-                        borderRadius: "5px",
-                        padding: "2px 5px 0 6px",
-                        fontWeight: "bold"
-                      }}
-                    >
-                      CANCELLED
-                    </span>
+        <ConditionalNewLayoutWrapper
```
twrichards: any chance we could get a screenshot of the visual effect of these changes (in reply to this comment is fine)

app/shared/productTypes.tsx
```diff
@@ -88,6 +88,11 @@ export interface HolidayStopFlowProperties {
   };
 }
 
+export interface DeliveryDetails {
```
twrichards: 💯 nice

app/shared/productResponse.ts
```diff
@@ -31,6 +33,21 @@ export const alertTextWithoutCTA = (productDetail: ProductDetail) =>
 export const sortByJoinDate = (a: ProductDetail, b: ProductDetail) =>
   b.joinDate.localeCompare(a.joinDate);
 
+export interface DeliveryRecordsDetail {
```
twrichards: this and the new type above probably should move to `app/client/components/delivery/records/deliveryRecordsApi.ts` since they don't need to be shared with server-side (as far as I can tell) and this `productResponse.ts` just has interfaces relating to `members-data-api` responses IIRC

app/client/components/delivery/records/deliveryRecordsTable.tsx
rBangay: The isOdd rule that I have used here is slightly more involved than the evenodd or nth-type css selectors as in some circumstances we have grouped tr elements (Where there is a delivery record problem or alert message). In this scenario instead of the following:

`even
odd
even
odd`

the correct rule might be:

`even
odd
odd
even
odd`

app/client/components/delivery/records/deliveryRecordsAddress.tsx
```diff
@@ -0,0 +1,61 @@
...
+            {props.addressLine2 && <li>{props.addressLine2}</li>}
+            {props.town && <li>{props.town}</li>}
+            {props.region && <li>{props.region}</li>}
+            {props.country && <li>{props.country}</li>}
+          </ul>
+        )}
+      </div>
+      <span
+        css={css`
+          display: block;
```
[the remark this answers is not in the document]
rBangay: I spoke to Cloe about this. Her preference was to keep the `Read more` link below the postcode as it should match up with the `Read more` link for the delivery instructions (which I've just finished)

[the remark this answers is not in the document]
twrichards: fair enough, would it make sense to line it up with the address column though?

app/client/components/delivery/records/deliveryRecordsTable.tsx
twrichards: ahh make sense!

app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -0,0 +1,104 @@
...
+  for (let i = filteredData.results.length - 1; i >= 0; --i) {
+    if (filteredData.results[i].hasHolidayStop) {
+      continue;
+    }
+    if (
+      currentAddress &&
+      getRecordAddressAsString(filteredData.results[i]) !== currentAddress
+    ) {
+      // tslint:disable-next-line: no-object-mutation -- maybe this should live in the API?
+      filteredData.results[i].isChangedAddress = true;
```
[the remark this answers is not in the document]
twrichards: https://github.com/guardian/support-service-lambdas/pull/548

PR #366, opened 2020-02-26
Self service delivery problem reporting

rBangay:
Allow users to view delivery records and report problems.

[Ticket](https://trello.com/c/AqhxzO63/860-goal-release-self-service-delivery-reporting)
step 1:
step 1

step 2:
step 2

step 3:
step 3

step 3 validation error:
step 3b

step 4:
step 4

step 4b (clicking on "Continue to step 2" button):
step 4b

aside, opening of the contact us components:
step 4c

step 5:
step 5

step 5b (add/update phone number):
step 5b

aside, opening of the contact us component in review stage:
step 5c

step 6:
step 6

step 7:
party


app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -224,6 +224,20 @@ export const DeliveryRecordsFC = (props: DeliveryRecordsFCProps) => {
...
+              margin: ${space[6]}px 0 ${space[12]}px;
+            `}
+          >
+            <ProductDetailsTable
+              productName={capatalize(
+                props.routeableStepProps.productType.shortenedFriendlyName ||
+                  props.routeableStepProps.productType.friendlyName
+              )}
+              subscriptionId={props.productDetail.subscription.subscriptionId}
+              isGift
```
twrichards: **I think this is always setting it to gift,** rather than taking that from the `productDetail` (you probably want to check the other places where `[GIFT]` is added to see how they work out if the product is a gift)

app/client/components/delivery/records/productDetailsTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+        Subscription details
+        {props.isGift && (
+          <span
+            css={css`
+              display: inline-block;
+              margin-left: ${space[3]}px;
+            `}
+          >
+            [GIFT]
+          </span>
```
twrichards: Not sure this should be following `Subscription details` probably better beside the `{props.productName}` on line 89

app/shared/productTypes.tsx
```diff
@@ -94,16 +103,35 @@ export interface HolidayStopFlowProperties {
...
 }
 
+export interface DeliveryProblemType {
+  label: string;
+  messageIsMandatory: boolean;
+}
+
+export const commonDeliveryProblemTypes: DeliveryProblemType[] = [
+  { label: "Damaged Paper", messageIsMandatory: true },
+  { label: "Delivered Despite Holiday", messageIsMandatory: false },
```
twrichards: it doesn't look like this option allows the user to tick holiday stop delivery records which would make sense
speaking to @JessL123 we should just drop this option for the time being and she's raising a ticket to reinstate it once we've figured out how it should work (inc. w.r.t. crediting etc.)

app/shared/productTypes.tsx
```diff
@@ -94,16 +103,35 @@ export interface HolidayStopFlowProperties {
...
+  { label: "No Delivery", messageIsMandatory: false },
+  { label: "Other", messageIsMandatory: true }
+];
+
+interface DeliveryRecordsProperties {
+  showDeliveryInstructions?: true;
+  numberOfProblemRecordsToShow: number;
```
twrichards: I can't actually see this being used to limit the items available for selection in 'report problem' mode?

app/shared/productTypes.tsx
```diff
@@ -94,16 +103,35 @@ export interface HolidayStopFlowProperties {
...
     subscription: Subscription
   ) => subscription is SubscriptionWithDeliveryAddress;
-  showRecords?: true;
-  showDeliveryInstructions?: true;
+  records?: DeliveryRecordsProperties;
 }
 
 export interface ProductType {
   friendlyName: ProductFriendlyName;
+  shortenedFriendlyName?: ProductShortenedFriendlyName;
```
twrichards: can't actually see this being used?

app/client/components/delivery/records/deliveryRecordsProblemContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+  subscription: Subscription;
+  subscriptionCurrency: string;
+  productName: string;
+  apiProductName?: string;
+  problemType?: DeliveryRecordsProblemType;
+  affectedRecords: DeliveryRecordApiItem[];
+  deliveryProblemMap: DeliveryProblemMap;
+  isTestUser: boolean;
+  showProblemCredit?: boolean;
+  repeatDeliveryProblem?: boolean;
```
twrichards: for home delivery, I can't seem to see this being set based on if they've reported a problem already in the last 14 days? 
nor it entering the 'escalation' flow for the same reason, similarly I can't seem to see where it enters the 'escalation' flow when `cancelledAt` is true

twrichards: we have another requirement, that it enters the 'escalation' flow when it's `autoRenew = false`

twrichards: did something go wrong with your rebase/commits that wouldn't include some of the `cancelledAt` stuff we did

app/shared/productTypes.tsx
```diff
@@ -31,6 +31,15 @@ export type ProductFriendlyName =
...
   | "subscription";
+export type ProductShortenedFriendlyName =
+  | "membership"
+  | "recurring contribution" // TODO use payment frequency instead of 'recurring' e.g. monthly annual etc
+  | "newspaper"
+  | "newspaper voucher"
+  | "home delivery"
+  | "digital"
+  | "Guardian Weekly"
+  | "subscription";
```
twrichards: this type definition need only include the ones you're specifying below, and tbh probably needn't exist if we're not restricting the type at the point it's used (just adds noise/confusion)
_To clarify, keep the `shortenedFriendlyName` property of `ProductType` just don't think it needs at type_

app/shared/productTypes.tsx
twrichards: [MINOR] could be called just `shortFriendlyName`

app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -1,77 +1,591 @@
...
+          props.routeableStepProps.productType.fulfilmentDateCalculator
+            ?.productFilenamePart,
+        problemType: deliveryProblem,
+        affectedRecords: props.data.results.filter(record =>
+          selectedProblemRecords.includes(record.id)
+        ),
+        deliveryProblemMap: props.data.deliveryProblemMap,
+        isTestUser: props.productDetail.isTestUser,
+        showProblemCredit:
+          !props.productDetail.subscription.cancelledAt &&
```
twrichards: We have a requirement that missed along the way...
```suggestion
          !props.productDetail.subscription.cancelledAt &&
          props.productDetail.subscription.autoRenew &&
```

app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -1,77 +1,591 @@
...
+  const isRecordInCurrentPage = (
+    index: number,
+    currentPageStartIndex: number,
+    currentPageEndIndex: number
+  ) => index >= currentPageStartIndex && index <= currentPageEndIndex;
+
+  const hasExistingDeliveryProblem =
+    props.data.results.findIndex(
+      deliveryRecord => deliveryRecord.problemCaseId
+    ) > -1;
```
twrichards: i think we need to change this implementation to make it explicitly last 14 days...
```suggestion
  const hasExistingDeliveryProblem =
    props.data.results.findIndex(
      deliveryRecord => (deliveryRecord.problemCaseId && deliveryRecord.date.minusDays(14) > TODAY)
    ) > -1;
```

app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -1,77 +1,591 @@
...
+            <ProductDetailsTable
+              productName={capatalize(
+                props.routeableStepProps.productType.shortenedFriendlyName ||
+                  props.routeableStepProps.productType.friendlyName
+              )}
+              subscriptionId={props.productDetail.subscription.subscriptionId}
+              isGift={isGift(props.productDetail.subscription)}
+            />
+          </div>
+          {props.data.results.find(record => !record.problemCaseId) && (
```
twrichards: In order to do the release only to me, as the last thing we change before releasing this, the following change should do it...
```suggestion
          {props.data.results.find(record => !record.problemCaseId) && window && window.identityDetails.userId==="15849095" (
```

app/client/components/delivery/records/deliveryRecords.tsx
twrichards: Just realised this suggestion is wrong... it should be more like ```deliveryRecord.deliveryDate > TODAY.minusDays(15)```

twrichards: If it's using Moments for the date substraction, make sure to clone() if referring to an existing variable for 'TODAY' otherwise it mutates the moment 😱

app/client/components/cancel/stages/executeCancellation.tsx
```diff
@@ -72,14 +72,15 @@ const getCancellationSummaryWithReturnButton = (
 const getCaseUpdatingCancellationSummary = (
   caseId: string,
   productType: ProductTypeWithCancellationFlow
-) => (withSubscription: WithSubscription | {}) => {
-  const subscription =
-    (withSubscription as WithSubscription).subscription || {};
+) => (productDetail: ProductDetail) => {
```
twrichards: looks like this has been broken a while, but it should be like this...
```suggestion
) => (productDetails: ProductDetail[]) => {
const productDetail = productDetails[0];
```
... because the API returns an array of one item.

PR #369, opened 2020-03-04
Report delivery problems update

rBangay:
- navigate up to parent if context information is missing eg user refreshes the confirmation page
- design review: button style consolidation, show more link styling, text area width, delivery record card alignment
- update component tests to reflect markup/component updates
- remove 'select all' checkbox
- update paragraphs to textSans
- add problem type form component tests
- don't show future records past read only mode
- add 'Delivered despite suspension' back into the problem options and link it to holiday stop records

begone!
futureRecords
holidayProblem


twrichards:
Not sure if there is still a spacing issue here 
image

PR #378, opened 2020-03-19
[RB] - problem reporting bug when reporting where no valid records to report

rBangay:
**After clicking on the 'Report a problem' button**,  if none of their delivery records can be reported, show info message and stop people from continuing down the problem reporting journey.

This may be because they have reported problems on all of the delivery records they have, or the records are too old.

youshallnotpasss

### DIFF
--- a/app/client/components/svgs/giftIcon.tsx
+++ b/app/client/components/svgs/giftIcon.tsx
@@ -38,8 +38,8 @@ export const GiftIcon = (props: GiftIconProps) => (
         fill="#333333"
       />
       <path
-        fill-rule="evenodd"
-        clip-rule="evenodd"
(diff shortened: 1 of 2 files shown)


twrichards:
Are we not disabling/hiding the button??

app/client/components/svgs/giftIcon.tsx
```diff
@@ -38,8 +38,8 @@ export const GiftIcon = (props: GiftIconProps) => (
         fill="#333333"
       />
       <path
-        fill-rule="evenodd"
-        clip-rule="evenodd"
+        fillRule="evenodd"
+        clipRule="evenodd"
```
twrichards: is this supposed to be in this PR?

PR #388, opened 2020-04-16
temporarily stop escalation of HD delivery problems based on repeat problem detection

twrichards:
…until call-centre has capacity to investigate.


twrichards:
TESTED ✅ 

Both the deliveries (reported separately) are credited...
image

... and notice how both of the cases are normal priority and both in the `DtoC One Off Issues` queue...

image

... yet SF correctly identifies them as **_Repeat Delivery Problem_** (for data correctness in the long term).

PR #400, opened 2020-04-27
[RB] - remove uneeded 'to' from delivery suspension message

rBangay:
readgood

### DIFF
--- a/app/client/components/delivery/records/deliveryRecordStatus.tsx
+++ b/app/client/components/delivery/records/deliveryRecordStatus.tsx
@@ -108,7 +108,7 @@ export const RecordStatus = (props: RecordStatusProps) => {
             <p>
               Unfortunately due to {props.bulkSuspensionReason}, we are unable
               to deliver your {props.productName}. You will be credited for this
-              issue off your next payment.
+              issue off the subsequent payment.
             </p>
           )}
         </>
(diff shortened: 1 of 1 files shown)


app/client/components/delivery/records/deliveryRecordStatus.tsx
```diff
@@ -107,8 +107,8 @@ export const RecordStatus = (props: RecordStatusProps) => {
           {props.bulkSuspensionReason && props.productName && (
             <p>
               Unfortunately due to {props.bulkSuspensionReason}, we are unable
-              to deliver to your {props.productName}. You will be credited for
-              this issue off your next payment.
+              to deliver your {props.productName}. You will be credited for this
+              issue off your next payment.
```
twrichards, line 111: @JessL123 I wonder if this should be...
```suggestion
              to deliver your {props.productName}. You will be credited for this
              issue off the next payment.
```
... or ...
```suggestion
              to deliver your {props.productName}. You will be credited for this
              issue off the subsequent payment.
```
... just because the user might have a payment between now and the invoice date the imposed suspension is locked in for (based on the lead times) - without further API changes we can't explicitly say which payment it will be credited on.

twrichards: Jian suggested going for the `subsequent` one, so committing it.

PR #373, opened 2020-03-12
detailed logging of all API proxying

twrichards:
_The general proxying approach used by `manage-frontend` is written up on [this wiki page](https://github.com/guardian/manage-frontend/wiki/Server-side-Proxying)._

### Consolidated proxying logic
In order to have a consistent logging mechanism, the proxying of API Gateway Lambdas now uses the `proxyApiHandler` used elsewhere (for the likes of `members-data-api` for example) with a few tweaks so the `x-api-key` and `x-identity-id` headers can still be injected.

### Detailed logging of all API proxying
All proxying is now logged in a detailed fashion, each endpoint now having a unique `loggingCode`. See example log entry below...
image

---
_Some PRs will follow, which will use these `loggingCode`s for Metric Filters, so we can build a dashboard for all these critical interactions._

PR #404, opened 2020-05-12
custom CloudWatch Metric for every `loggingCode`

twrichards:
Following https://github.com/guardian/manage-frontend/pull/373 (detailed logging) and https://github.com/guardian/manage-frontend/pull/402 (50% mechanism for Account Overview release) we can now put in place metrics to track all our key events for better visibility/monitoring.

The initial purpose is for extra visibility as we roll out Account Overview, but once that's out to 100% **we can begin using these metrics to set alarm thresholds (e.g. _no card updates in the last hour_).**

It uses the [`CloudWatch.putMetricData`](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudWatch.html#putMetricData-property) (part of AWS SDK - already a dependency). The metric namespace is fixed as `manage-frontend` and uses the `loggingCode` (e.g. `MDA_UPDATE_PAYMENT_CARD`) as the metric name, and since we're just counting here the `value` is `1` with the `Unit` as `Count`. We use dimensions to parameterise/segregate the metric...
- `Stage` - `DEV`/`CODE`/`PROD`
- `outcome`  - `SUCCESS`/`ERROR` (based on the `isOK` logging field)
- `variant` - `ACCOUNT OVERVIEW`/`LEGACY` (based on the `isInAccountOverviewTest` logging field)

Here's a (non-exhaustive) example of how they look...
image

### DIFF
--- a/app/server/apiProxy.ts
+++ b/app/server/apiProxy.ts
@@ -7,7 +7,7 @@ import { X_GU_ID_FORWARDED_SCOPE } from "../shared/identity";
 import { MDA_TEST_USER_HEADER } from "../shared/productResponse";
 import { isInAccountOverviewTest } from "./accountOverviewRelease";
 import { conf } from "./config";
-import { log } from "./log";
+import { log, putMetric } from "./log";
 import {
   augmentRedirectURL,
   getCookiesOrEmptyString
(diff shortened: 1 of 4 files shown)


twrichards:
Worth noting that I did experiment doing this with MetricFilters on the logs, but decided against because...
- Metric Filters don't support 'dimensions' so would need a naming convention for the metric name, which given there were 3 dimension, multiplied up very quickly
- the CloudFormation got very verbose, very quickly (would've needed to install the [count CloudFormation macro](https://stackoverflow.com/a/59636523/1868921) to the whole membership AWS account)
- lastly, as we add new logging codes we'd need to remember to update the metric filters in CloudFormation, now we get it automatically

PR #409, opened 2020-05-27
upgrade Sentry client from Raven "^2.5.1" to Sentry's own lib "^5.15.5"

twrichards:
Almost two years to the day from when Raven was added (https://github.com/guardian/manage-frontend/pull/14 & https://github.com/guardian/manage-frontend/pull/19) it's time to finally upgrade to Sentry's own library.

This is a long overdue ticket - https://trello.com/c/A4xHwp93/112-upgrade-javascript-library-on-manage-frontend-raven-sentry.

Raven is no more...
image


app/server/server.ts
```diff
@@ -49,8 +49,5 @@ server.use(routes.productsProvider("/api/"));
 // ALL OTHER ENDPOINTS CAN BE HANDLED BY CLIENT SIDE REACT ROUTING
 server.use(routes.frontend);
 
-if (conf.SERVER_DSN) {
-  server.use(Raven.errorHandler());
-}
```
rBangay, line 54: Is this part implicit in the new Sentry version?

twrichards: I thought it was, because when I tested and faked some errors it seemed to catch them, however according to https://sentry.io/for/express/ I should still include them, will make some changes accordingly.

twrichards: More detail : https://docs.sentry.io/platforms/node/express/

twrichards: done

PR #417, opened 2020-06-03
[RB TR] - Account overview product card refactor

rBangay:
## What does this change?
rename product.tsx to accountOverviewCard.tsx. Move away from string comparisons for branching logic and use properties in productType object (config)

## TODO

- [ ] replace this TODO section with a link to Wiki page about `ProductTypes` once written

### DIFF
--- a/app/shared/productTypes.tsx
+++ b/app/shared/productTypes.tsx
@@ -503,7 +503,7 @@ export const ProductTypes: { [productKey in ProductTypeKeys]: ProductType } = {
     productPage: "subscriptions"
   },
   guardianweekly: {
-    productTitle: calculateProductTitle("Guardian Weekly"),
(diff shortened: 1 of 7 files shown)


app/shared/productTypes.tsx
```diff
@@ -485,6 +503,7 @@ export const ProductTypes: { [productKey in ProductTypeKeys]: ProductType } = {
     productPage: "subscriptions"
   },
   guardianweekly: {
+    productTitle: calculateProductTitle("Guardian Weekly"),
```
twrichards, line 506: this should be...
```suggestion
    () => "Guardian Weekly",
```
... as we never want to display the plan name as part of the product title - I felt sure we did this though when we removed the `isSixForSix` from this call 🤔

PR #420, opened 2020-06-04
[RB TR] - consolidate manage components

rBangay:
## What does this change?
consolidate manageSubscription, manageMembership and manageContribution into a new singular component manageProduct

Fewer lines of repeated code :tada:

## Images
No! .. it shouldn't look any different :trollface:


app/shared/accountOverviewRelease.ts
```diff
@@ -5,4 +5,4 @@ export const isIdentityInAccountOverviewTest = (
 ) =>
   TEST_IDENTITY_ID_SUFFIXES.filter(
     suffix => identityID && identityID.endsWith(suffix)
-  ).length > 0;
+  ).length > 0 || true;
```
twrichards, line 8: 😱

PR #425, opened 2020-06-09
[RB TR] - refactor/rename FlowStart component and all of it's usages

rBangay:
_Part of the account overview release clean-up._ 

This PR is to refactor the old `FlowStartMultipleProductDetailHandler` component, to a much simpler and better named component **`ProductDetailWrapper`**. 
 
If you navigate **directly** to the start of a flow (from an email or help article for example), it will call `members-data-api` to fetch all the products of that type (based on the product type in the URL, e.g. `guardianweekly`), if you have...

- zero products of that type, you will now be redirected back to account overview (previously you would get a 'Sorry you don't have XXX' page with up-sell)
- **exactly one of that type**, as before it will pass that on to the child component so it's available throughout the flow
- more than one of that type, you will now be redirected back to account overview (previously you would be displayed a list of products and buttons for you to choose from, but this represented lots of duplication/complexity from the main display(s) of products)

... the above will ignore cancelled subs unless `allowCancelledSubscription` prop is set to true.

As before, if you navigate from the 'account overview' or 'manage product' pages then the desired `productDetail` is passed in the 'browser history state', so this component skips the `members-data-api` call and passes the `productDetail` direct to the child component.


app/client/components/productDetailWrapper.tsx
```diff
@@ -0,0 +1,75 @@
...
+  visuallyNavigateToParent
+} from "./wizardRouterAdapter";
+
+interface ProductDetailWrapperProps extends RouteableStepProps {
+  children: (productDetail: ProductDetail) => JSX.Element;
+  allowCancelledSubscription?: true;
+  loadingMessagePrefix: string;
+}
+
+export const ProductDetailWrapper = (props: ProductDetailWrapperProps) => {
```
twrichards, line 20: i'm having second thoughts about this name, as it's not wrapping ProductDetail it's wrapping flows, so it should it go back to...
```suggestion
export const ProductDetailProvider = (props: ProductDetailProviderProps) => {
```
... as this represents its current purpose/behaviour (despite the slight naming clash with Contexts).

**Then if/when we refactor the flows to have this component add the page furniture (nav & header) then we can rename it to just `FlowWrapper`.**

app/client/components/delivery/address/deliveryAddressForm.tsx
```diff
@@ -794,25 +793,12 @@ export const DeliveryAddressForm = (props: RouteableStepProps) => {
     )(props.location.state.allProductDetails);
   }
   return (
-    <FlowStartMultipleProductDetailHandler
+    <ProductDetailProvider // TODO: refactor to remove this wrapper (not needed)
```
twrichards, line 796: let's do this as our next refactor PR 😸

PR #428, opened 2020-06-09
[RB TR] - removal of account overview a/b switch and redundant code

rBangay:
image


twrichards:
🎉 
image

app/client/components/wizardRouterAdapter.tsx
```diff
@@ -131,7 +120,7 @@ const ThisStageContent = (props: WizardStepProps) => (
 export interface WizardStepProps extends CommonProps {
   routeableStepProps: RouteableStepProps;
   children: any;
-  fullWidth?: true;
+  fullWidth: true;
```
twrichards, line 123: I think we can eliminate this prop now, since it appears to be present on all usages.

PR #438, opened 2020-06-15
[RB] - navigation refactor

rBangay:
## What does this change?
Create a config file for both the left hand nav and the dropdown nav to run off. 

Seperate the navs into 2 spereate files within a nav
directory. 

Reintroduce icons at smaller (mobile) breakpoints for the
dropdown nav.

## Images

dropdown before             |  dropdown after
 |  

desktopDropdown

noChnage


app/client/components/svgs/profileIcon.tsx
```diff
@@ -3,14 +3,21 @@ import { palette } from "@guardian/src-foundations";
 import React from "react";
 
 interface ProfileIconProps {
-  css?: SerializedStyles;
+  additionalCss?: SerializedStyles;
```
twrichards, line 6: is this actually being used? (it looks like the old `css?` prop wasn't used)

rBangay: It's used in dropdownNav for the little profile roundel next the the title "My Account" in order to position it within the circle

PR #439, opened 2020-06-15
simplify first time set-up and day-to-day running of manage-frontend

twrichards:
add nginx configs to proxy CODE instances of `identity-frontend` and `members-data-api` **as a fallback** if not running locally (to develop those services in parallel one has to simply run them locally and it will use instead of CODE).

The set-up instructions were quite involved for new starters previously - now just one set-up script for nginx, plus a couple of simple commands for manage itself.

Also day-to-day (for all devs) having to start various other services locally was not only a faff but a distraction when debugging (if one crashed or didn't have some config etc).

README changes cover this

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
-# Manage Frontend - manage.theguardian.com
+# manage.theguardian.com
 
-Frontend of `manage.theguardian.com`
+This README contains mainly set-up information.
+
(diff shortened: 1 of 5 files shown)


twrichards:
will do a final test with @SabrinaCode (as an equivalent to new dev starter)

PR #440, opened 2020-06-16
Ie11 fixes

rBangay:
## Images
before

after


app/client/components/page.tsx
```diff
@@ -122,18 +122,19 @@ export const PageHeaderContainer: React.SFC<PageHeaderContainerProps> = (
   const gridBasev2 = () => {
     return `
       display: grid;
-      ms-grid-columns: (minmax(0, 1fr))[${gridColumns.default}];
+      display: -ms-grid;
+      -ms-grid-columns: (minmax(0, 1fr))[${gridColumns.default}];
```
twrichards, line 126: doesn't Emotion do the per-browser CSS options for you?

PR #442, opened 2020-06-17
[RB] - move page furniture into productDetailProvider

rBangay:
## What does this change?
Creation of `FlowWrapper` and `PageContainer` components to avoid duplication across the product flow journeys.

- `FlowWrapper` component is now used as the wrapping component for multi step/page journeys and contains both `ProductDetailProvider` and `PageContainer` components

- `PageContainer` component now replace remaining usages of `PageHeaderContainer` and `PageNavAndContentContainer` components in standalone pages.

- Phased out 'in page footer' components as they weren't adding anything useful and represented unnecessary complexity

- Now that everything is wrapped in the `PageContainer` (directly or indirectly) we were able to simplify section of the page that used to have a max-width without losing the top margin (now represented by `WithStandardTopMargin` component)

tenor


app/client/components/cancel/stages/executeCancellation.tsx
```diff
@@ -184,28 +182,9 @@ export const ExecuteCancellation = (
...
-                  <PageNavAndContentContainer
-                    selectedNavItem={NAV_LINKS.accountOverview}
-                  >
-                    {innerContent(productDetail, props, reason, caseId)}
-                  </PageNavAndContentContainer>
-                </>
-              )}
+              {productDetail => {
+                return innerContent(productDetail, props, reason, caseId);
+              }}
```
twrichards, line 187:
```suggestion
              {productDetail => innerContent(productDetail, props, reason, caseId)}
```

app/client/components/page.tsx
```diff
@@ -8,79 +8,14 @@ import { minWidth } from "../styles/breakpoints";
...
-);
-
-interface PageContainerProps {
+interface EmptyPageContainerProps {
   noVerticalMargin?: true;
   children: React.ReactNode;
   fullWidth?: true;
 }
-export const PageContainer = (props: PageContainerProps) => (
+export const EmptyPageContainer = (props: EmptyPageContainerProps) => (
```
twrichards, line 18: really think we need a different name for this, perhaps `PagePart`?

app/client/components/wizardRouterAdapter.tsx
```diff
@@ -65,32 +53,23 @@ export const ReturnToAccountOverviewButton = () => (
...
 );
 
-const RootComponent = (props: RootComponentProps) => (
-  <>
-    <PageContainer fullWidth>{props.thisStageChildren}</PageContainer>
-    {props.extraFooterComponents}
-  </>
-);
+interface CurrentStepProps {
+  path: string;
```
twrichards, line 57: Let's constrain this as a string literal type
```suggestion
  path: "/";
```

twrichards: to help reinforce the idea that this component is a holder for 'the content of this step'

PR #450, opened 2020-07-02
[RB] - migratation to react stripe js

rBangay:
## What does this change?
migrate from old react-stripe-elements to React Stripe.js. 

Introducing hooks along the way (new react stripe implementation also adds hooks functionality)

_**Related historical PR:** https://github.com/guardian/manage-frontend/pull/241_

## Images
Nothing should change visually 

Screenshots are from edge browser. (cross browser tested in ie11, edge, chrome win, chrome mac, firefox mac)

edge-1
edge-2
edge-3
edge-4
edge-5
edge-6


app/client/components/payment/update/card/stripeCardInputForm.tsx
```diff
@@ -11,143 +11,43 @@ import { Spinner } from "../../../spinner";
...
+  const isLoaded = () => {
+    return stripe && readyElements.length === 3 && props.stripeSetupIntent;
+  };
 
-  private markElementReady = (element: string) => () =>
-    this.setState(prevState => ({
-      readyElements: prevState.readyElements.concat(element)
-    }));
+  const markElementReady = (element: string) => () =>
+    setReadyElements(readyElements.concat(element));
```
twrichards, line 47: this could use spread these days

twrichards: in fact, I suspect this the cause of the hanging on image as previously it used the overload of setState which was a function from old state to new, whereas this new hooks approach is just using the state available on the last render, which if these elements mark themselves as ready in quick succession between renders, then we'd get stuck

twrichards: as discussed IRL, maybe better to pull the readiness out into three little pieces of state to avoid this whole fiddlyness and code will be more readable too 🙂

twrichards: referenced this problem here https://github.com/guardian/manage-frontend/wiki/Major-TODOs#replace-all-react-class-based-components-with-hooks-based-functional-components as a gotcha should the team ever decide to migrate any/all the class based components to hooks

app/client/components/payment/update/card/stripeCardInputForm.tsx
```diff
@@ -34,18 +38,23 @@ interface StripeInputFormError {
 
 export const StripeCardInputForm = (props: StripeCardInputFormProps) => {
   const [isValidating, setIsValidating] = useState<boolean>(false);
-  const [readyElements, setReadyElements] = useState<string[]>([]);
+  const [cardNumberReady, markCardNumberReady] = useState<boolean>(false);
+  const [cardExpiryReady, markCardExpiryReady] = useState<boolean>(false);
+  const [cardCVCReady, markCardCVCReady] = useState<boolean>(false);
```
twrichards, line 43: Minor but implies boolean a bit better
```suggestion
  const [isCardNumberReady, markCardNumberReady] = useState<boolean>(false);
  const [isCardExpiryReady, markCardExpiryReady] = useState<boolean>(false);
  const [isCardCVCReady, markCardCVCReady] = useState<boolean>(false);
```

app/client/components/payment/update/card/flexCardElement.tsx
```diff
@@ -33,7 +35,7 @@ export const FlexCardElement = (props: FlexCardElementProps) => (
             style: baseStyle,
             placeholder: "Card Number"
           }}
-          onReady={props.markElementReady("CardNumberElement")}
+          onReady={() => props.markCardNumberReady(true)}
```
twrichards, line 38: just a though but we could change these new bits of state to hold the actual element given by the `onReady` function (with type `undefined | StripeElementBase`) and appropriate names (e.g. `cardNumberReady` -> `cardNumberElement`) which would allow us to do something like
```suggestion
          onReady={props.setCardNumberElement}
```

app/client/components/payment/update/card/stripeCardInputForm.tsx
twrichards: or changing them to...
```suggestion
  const [cardNumberElement, setCardNumberElement] = useState<undefined | StripeElementBase>();
  const [cardExpiryElement, setCardExpiryElement] = useState<undefined | StripeElementBase>();
  const [cardCVCElement, setCardCVCElement] = useState<undefined | StripeElementBase>();
```
... as per my other comment.

Just a thought, not sure which is nicer.

app/client/components/payment/update/card/stripeCardInputForm.tsx
```diff
@@ -110,9 +119,9 @@ export const StripeCardInputForm = (props: StripeCardInputFormProps) => {
         return;
       }
 
-      //
       const intentResult = await stripe.confirmCardSetup(
-        props.stripeSetupIntent.client_secret
+        props.stripeSetupIntent.client_secret,
+        { payment_method: createPaymentMethodResult.paymentMethod.id }
```
twrichards, line 124: on reflection this explicitness is better than the implicit way this worked before 👍

app/client/components/payment/update/card/stripeCardInputForm.tsx
```diff
@@ -173,18 +82,26 @@ export class StripeCardInputForm extends React.Component<
...
-        "card",
-        {
-          billing_details: {
-            name: this.props.userEmail,
-            email: this.props.userEmail
-          }
+  const startCardUpdate = (navigate: NavigateFn) => async () => {
+    setIsValidating(true);
+    const cardElement = elements?.getElement(CardNumberElement);
+    if (!cardElement) {
```
twrichards, line 88: have you simulated this (by faking the condition) to test the error looks right? (and that the `return;` rather than `return null;` is legit?

rBangay: This is the error you get:

Screenshot 2020-07-03 at 12 47 44

The return craps you out correctly yep :)

PR #452, opened 2020-07-06
Front page newsletter

rBangay:
## What does this change?
Add the ability to opt in/out of the front page newsletter for all print based subscriptions from within the corresponding manage subscription pages.

Reliant on the following api work by the identity team:
https://github.com/guardian/identity/pull/1777

## Images
Screenshot 2020-07-06 at 12 18 56


app/shared/productTypes.tsx
```diff
@@ -351,6 +353,7 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
     allProductsProductTypeFilterString: "Paper",
     urlPart: "paper",
     getOphanProductType: () => "PRINT_SUBSCRIPTION",
+    showRestrictedNewsletters: ["6009"],
```
arausuy, line 356: 6009 is the identifier of the voucher newsletter? Would it benefit from being extracted to a value?

PR #460, opened 2020-07-21
Add ophan tracking to newsletter preferences

buck06191:
## What does this change?

Tracking in Ophan is needed for the newsletters in order to measure impact of upcoming changes as part of the newsletters OKR.

We have added `data-link-name` attributes to the newsletter checkboxes and have used the `identityName` as the identifier. This also means that the `ConsentOptions` model has been updated to include this field.

Co-authored-by: Alex Dufournet <alex.dufournet@theguardian.com>

## How to test

On the marketing preferences page, checking and unchecking a newsletter should trigger an Ophan page click tracking event in the network inspector. This event shouldn't be triggered for normal marketing preferences.

Ophan tracking from the email preferences page.

I don't think there are any potential risks here, unless there needs to be an update to any privacy information about this being tracked?

### DIFF
--- a/app/client/components/identity/MarketingPreference.tsx
+++ b/app/client/components/identity/MarketingPreference.tsx
@@ -6,6 +6,7 @@ import { Checkbox } from "../checkbox";
 interface MarketingPreferenceProps {
   id: string;
   description: string;
+  identityName?: string;
   frequency?: string;
   title?: string;
   selected?: boolean;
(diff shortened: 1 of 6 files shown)


app/client/components/identity/MarketingPreference.tsx
```diff
@@ -101,6 +113,7 @@ export const MarketingPreference: FC<MarketingPreferenceProps> = props => {
           position: "relative"
         }
       ]}
+      data-link-name={identityName ? linkName : null}
```
twrichards, line 116: should this be...
```suggestion
      data-link-name={identityName ? linkName : undefined}
```
... I have no idea, just stood out to me? (i think if `undefined` it's the equiv. of not specifying the attribute)

twrichards:
might be worth a +1 from @paddyo (the original author of this side of MMA)

buck06191:
> LGTM 👍
> 
> Is it working as you expect in `CODE`?

Just checked and it works as expected. Thanks

PR #463, opened 2020-07-28
[RB] - dropdown nav update

rBangay:
## What does this change?
previously at larger (desktop) breakpoints the dropdown nav only had "Comments & replies" and "Sign out" in it. Now it has the full set of nav items at all breakpoints.

## Images
Before  |  After
  |

### DIFF
--- a/app/client/__tests__/__snapshots__/main.tsx.snap
+++ b/app/client/__tests__/__snapshots__/main.tsx.snap
@@ -195,16 +195,14 @@ exports[`Main renders something 1`] = `
           css={
             Object {
               "map": undefined,
-              "name": "gud8wi",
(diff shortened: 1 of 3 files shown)


twrichards:
Whilst I personally prefer it this new way, was their a change of heart or some user research/feedback that provoked this change?

rBangay:
I think the rational was to make the navigation more consistent with the dotcom 'my account' dropdown. I'll get a better answer from Jian tomorrow thought ;)

rBangay:
Yep so from JIan:

"We need to make it consistent, yes.
That was the aim of testing when UX ran the testing originally.
Current dropdown in MMA, which shows only comment and signout is not great UX experience and needs to change"

PR #467, opened 2020-08-11
Jest update

ripecosta:
## What does this change?
Updates Jest packages to the latest versions. Also updates Node to v10.22.0 and Typescript to v3.9.7 because the versions we were using weren't supported by the latest version of Jest. Fixes `useState` type issues that were created from the update.

Cross browser tested using browserstack.


twrichards:
you will also want to update the node version baked into the AMI, see https://amigo.gutools.co.uk/recipes/manage-frontend

app/client/components/identity/PublicProfile/index.tsx
```diff
@@ -112,7 +110,7 @@ export const PublicProfile = (_: { path?: string }) => {
       <WithStandardTopMargin>
         {error ? <GenericErrorMessage ref={errorRef} /> : null}
       </WithStandardTopMargin>
-      {loading ? loader : content()}
+      {user ? content(user) : loader}
```
twrichards, line 113: changes to this file seem a little extreme to resolve a little more type strictness, adding @paddyo to the reviewers

ripecosta: Thanks. I had to specify `user`'s type to be `User | undefined` and `loading`'s value is only true when `user` is `undefined`. This felt redundant so I refactored it a little bit. Leaving the campsite tidier type of thing. The remaining changes are just so the type-checker understands that `user` is not used unless it is not `undefined`.

ripecosta:
> you will also want to update the node version baked into the AMI, see https://amigo.gutools.co.uk/recipes/manage-frontend

Thanks! Didn't know about this. I'll update it just before I merge this.

app/client/components/identity/PublicProfile/index.tsx
rBangay: This might actually be helpful in the near future. As part of our planned migration from tslint to eslint, the default linting configuration is stricter. I guess we could remove some of that strictness in the config, but I think I prefer it (obviously open to discussion :) )

paddyo: I think these changes look good. There was a reason we had a separate loader/user thing but it may have become redundant as the pages took shape when being built because it doesn't look needed any more and the refactor makes sense.

PR #475, opened 2020-08-17
Dependancy updates

rBangay:
Update project lodash dependancy from `4.17.19` to `4.17.20` due to the following vulnerability: https://snyk.io/vuln/SNYK-JS-LODASH-590103.

Update .snyk file for peer dependencies that can't be updated yet (set to ignore for 30 days, hopefully when a viable update is available)


app/.snyk
```diff
@@ -1,6 +1,23 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.14.1
-ignore: {}
+version: v1.13.5
```
ripecosta, line 2: Why was the version decreased?

twrichards: I wonder if Richard was running a different version of Snyk CLI than the last committer?

rBangay: Yep .. this was generated using `yarn snyk wizard` which used the snyk version from package.json. Perhaps the last committer had a more recent version installed globally?

rBangay: What is the consensus? Shall I leave the older version as is or update the version we have in package.json to match the newer (or latest) version

ripecosta: I would prefer updating the snyk version as well, if possible. Either as part of this or a separate PR. It might be a simple thing to do since it seems like the newer version has been used previously.

What do you think @rBangay @twrichards ?

PR #474, opened 2020-08-17
Invoice api integration

rBangay:
## What does this change?
Add MMA proxy endpoints for listing all user invoices and specific invoice pdf's.

Making use of AWS signature functionality and reliant on the following invoicing-api functionality

- [get all invoices by identityId](https://github.com/guardian/invoicing-api/tree/e77d66c2db8e477ffac411969eeaa6e1c44285ab/src/main/scala/com/gu/invoicing/invoice)
- [get invoice PDF](https://github.com/guardian/invoicing-api/tree/d7ebf6b711f6161c2a2b7c316f0823af74a5d7f8/src/main/scala/com/gu/invoicing/pdf)

PR #477, opened 2020-08-20
Billing invoices

rBangay:
### Billing Invoices

Display invoices below product payment info in billing tab.

The starting point for the client is billing.tsx where the existing payment information was already being displayed. The Async loader has been amended to load the new invoices endpoint as well as the existing members-data-api call.

There is a new component `invoicesTable` to handle the display of the invoices (the layout is substantially different from the productDescriptionListTable component at different breakpoints).

There are 2 other new components, `pagination` which should be re-usable throughout the codebase and `invoiceTableYearSelect` to handle to the dynamic display and selection of invoice years at the top of the invoice table.

Makes use of 2 new MMA api endpoints: `/invoices` and `/invoices/{invoiceId}`
https://github.com/guardian/manage-frontend/pull/474
The first one for getting a list of all invoices per user and the second to get a specific invoice pdf

Includes minor changes to apiGateway and apiProxy to handle the forwarding of headers from the response on to the client. Useful in this case to handle the correct content-type for the pdf endpoint. Also allow the option of not logging the body response to the console ... useful in this case so as not to log the contents of a large pdf to the console.

### Images
<details>
 <summary>Billing tab with invoices</summary>

</details>

<details>
 <summary>invoice table with pagination and different payment methods</summary>

</details>

<details>
 <summary>billing tab at mobile breakpoint with invoice tables</summary>

</details>

<details>
 <summary>example invoice view</summary>

</details>

<details>
 <summary>example invoice download dialog box (firefox)</summary>

</details>

<hr />

<details>
 <summary>ie 11</summary>
Screenshot 2020-08-21 at 12 59 02
Screenshot 2020-08-21 at 12 59 37
<p>Both `view invoice` and the download icon prompt the user to save the pdf to disk. The download attribute of the anchor tag doesn't seem to be supported in ie, subsequently the filename of the invoice defualts to response from the api call eg the invoice id</p>
</details>

<details>
 <summary>chrome</summary>
Screenshot 2020-08-21 at 13 09 16

Screenshot 2020-08-21 at 13 10 40
</details>

<details>
 <summary>firefox</summary>

</details>


twrichards:
design thing, but if the `View invoice (PDF)` is going to have a line above it, then the line below it should be fatter so one knows which download link belongs to which entry (or remove the lined above `View invoice (PDF)` image

app/client/components/billing/billing.tsx
```diff
@@ -31,13 +37,27 @@ import { PaymentDetailsTable } from "../payment/paymentDetailsTable";
...
 };
 
-const BillingRenderer = (apiResponse: MembersDataApiItem[]) => {
-  const allProductDetails = apiResponse.filter(isProduct).sort(sortByJoinDate);
+class BillingDataAsyncLoader extends AsyncLoader<
+  [MembersDataApiItem[], { invoices: InvoiceDataApiItem[] }]
+> {}
+
+const BillingRenderer = (
+  apiResponse: [MembersDataApiItem[], { invoices: InvoiceDataApiItem[] }]
```
twrichards, line 51: how about de-structuring (by name) to avoid the `apiResponse[0]` and `apiResponse[1]` further down...
```suggestion
  [mdaResponse, invoiceResponse]: [MembersDataApiItem[], { invoices: InvoiceDataApiItem[] }]
```

app/client/components/billing/billing.tsx
```diff
@@ -100,6 +120,26 @@ const BillingRenderer = (apiResponse: MembersDataApiItem[]) => {
...
+                        productDetail.subscription.subscriptionId
+                    )
+                    .map(invoice => {
+                      return {
+                        ...invoice,
+                        pdfPath: `/api/${invoice.pdfPath}`,
+                        currency: paidPlan.currency,
+                        currencyISO: paidPlan.currencyISO
+                      };
+                    });
```
twrichards, line 139: expression is alway cleaner...
```suggestion
                    .map(invoice => ({
                        ...invoice,
                        pdfPath: `/api/${invoice.pdfPath}`,
                        currency: paidPlan.currency,
                        currencyISO: paidPlan.currencyISO
                      })
                   );
```

app/client/components/billing/billing.tsx
```diff
@@ -211,11 +264,25 @@ const BillingRenderer = (apiResponse: MembersDataApiItem[]) => {
...
+  Promise.all([
+    allProductsDetailFetcher(),
+    fetch("/api/invoices", {
+      credentials: "include",
+      mode: "same-origin",
+      headers: {
+        [X_GU_ID_FORWARDED_SCOPE]: getScopeFromRequestPathOrEmptyString(
+          window.location.href
+        )
+      }
```
twrichards, line 286: is this needed or just copy/pasta from some other `fetch` call? This is for payment failure magic links, so not relevant here, I'd be inclined to remove it...

app/client/components/billing/invoicesTable.tsx
```diff
@@ -0,0 +1,328 @@
...
+}
+
+export const InvoicesTable = (props: InvoicesTableProps) => {
+  const initialPage = 1;
+
+  const [currentPage, setCurrentPage] = useState<number>(initialPage);
+
+  const tableHeadings = ["Date", "Payment method", "Price", ""];
+  const invoiceYears = [
+    ...new Set(
```
twrichards, line 40: does IE support `Set`? (or are sure it's being polyfilled)

rBangay:
> design thing on mobile, but if the `View invoice (PDF)` is going to have a line above it, then the line below it should be fatter so one knows which download link belongs to which entry (alternatively remove the line above `View invoice (PDF)` image

Jian is in agreement. The solution is to remove the top border from `View invoice (PDF)` 👍

app/server/awsIntegration.ts
```diff
@@ -12,12 +12,12 @@ const PROFILE = "membership";
 
 const CREDENTIAL_PROVIDER = new AWS.CredentialProviderChain([
   () => new AWS.SharedIniFileCredentials({ profile: PROFILE }),
-  ...AWS.CredentialProviderChain.defaultProviders
+  ...AWS.CredentialProviderChain.defaultProviders,
```
twrichards, line 15: I feel like these have made it in due to a `no-verify` commit at some point and thus your local prettier (which has different rules) has added extra/noisy changes to this PR... I recommend...
1) removing this file from the commit (since the only changes look to be incorrect prettier ones from your machine)
2) once this PR is out, addressing https://github.com/guardian/manage-frontend/wiki/Major-TODOs#move-prettier-config-to-a-config-file-for-consistency

twrichards (approved):
Its looking good, I didn't really review the layout/React stuff too closely (can't spare the time, sorry) but the rest looks nice (see comments though).

I also recommend a thorough manual regression test in `CODE` of all the flows, given we changed the api integration a fair bit in the end.

app/client/components/billing/invoicesTable.tsx
rBangay: ie11 seems to have enough support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility

Screenshot 2020-08-24 at 11 20 47

app/client/components/billing/billing.tsx
rBangay: Yep, copy pasta I'm afraid ... removed now.
copypasta

rBangay:
> Its looking good, I didn't really review the layout/React stuff too closely (can't spare the time, sorry) but the rest looks nice (see comments though).
> 
> I also recommend a thorough manual regression test in `CODE` of all the flows, given we changed the api integration a fair bit in the end.

:+1 I've regression tested the payment details update, delivery address update, delivery record problem reporting and holiday suspension journeys in CODE

app/client/components/billing/billing.tsx
```diff
@@ -100,6 +114,24 @@ const BillingRenderer = (apiResponse: MembersDataApiItem[]) => {
                     null,
                     !!productDetail.alertText
                   );
+                  const paidPlan = getMainPlan(
+                    productDetail.subscription
+                  ) as PaidSubscriptionPlan;
+                  const productInvoiceData = invoiceData
+                    .filter(
+                      invoice =>
+                        invoice.subscriptionName ===
```
paulbrown1982, line 123: It's a shame the two APIs can't agree on a name for this field. In the Zuora UI it's Subscription Number, in their API it's subscriptionName, in our models and UX it's subscriptionId and Subscription ID respectively 🤦

app/client/components/billing/invoicesTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+  invoiceData: InvoiceInfo[];
+}
+
+export const InvoicesTable = (props: InvoicesTableProps) => {
+  const initialPage = 1;
+
+  const [currentPage, setCurrentPage] = useState<number>(initialPage);
+
+  const tableHeadings = ["Date", "Payment method", "Price", ""];
+  const invoiceYears = [
```
paulbrown1982, line 39: Does this rely on invoiceData being sorted by `invoice.date`?

app/client/components/billing/billing.tsx
rBangay: Hey Paul, yep you are right. It's been discussed and there is a plan to take a look at members data and MMA with the goal of renaming `subscriptionId` to `subscriptionName` it a separate ticket.

app/client/components/billing/invoicesTable.tsx
rBangay: Hey, yep invoiceData is sorted by date before being passed through to the invoicesTable component as a prop.

I can see what you are saying, the component will work without the invoiceYears being sorted, you may just get a list of years not in chronological order.

Do you think it is worth explicitly sorting the years in this component, or leave it knowing that the data has already been sorted elsewhere?

paulbrown1982 (approved):
👍  This looks good to me. Good catch not to log the PDF body. Thank you for including all the screenshots, it really helped contextualise the change.

app/client/components/billing/invoicesTable.tsx
paulbrown1982: It's a tiny change to run `.sort()` on the end of the array assignment,  so possibly safer to do that than relying on Set() not reordering or the API changing.

mario-galic (approved):

PR #479, opened 2020-08-24
[RB] - only load stripe SDK on intent

rBangay:
## What does this change?
only load the stripe sdk when the user explicitly decides to update their payment details:

Screenshot 2020-08-24 at 10 36 28

I've taken some screenshots in order to explain the approach here ...

The first image is the default route (account overview) on the existing master branch
01_master

The 2nd image is if we move the loadStripe call within the useEffect hook but keep the existing stripe imports (again on the default route - account overview):
02_loadstripeInUseEffect

The 3rd and 4th images are the approach taken in this pr (the 1st image being the default route and the 2nd being the update payment page after capthca is complete)
03_asyncLoadStripeImport

update payment page:
04_asyncLoadStripeImport

### DIFF
--- a/app/client/components/payment/update/card/cardInputForm.tsx
+++ b/app/client/components/payment/update/card/cardInputForm.tsx
@@ -74,9 +74,7 @@ export const CardInputForm = (props: CardInputFormProps) => {
   };
 
   const loadSetupIntent = (recaptchaToken: string) => {
-    if (stripePromise === null) {
-      loadStripe();
-    }
+    loadStripe();
 
     setDidCompleteRecaptcha(true);
 
--- a/app/client/components/payment/update/card/cardInputForm.tsx
+++ b/app/client/components/payment/update/card/cardInputForm.tsx
@@ -1,6 +1,6 @@
 import * as Sentry from "@sentry/browser";
 import { Elements } from "@stripe/react-stripe-js";
-import { loadStripe } from "@stripe/stripe-js";
+import { Stripe } from "@stripe/stripe-js";
 import React, { useEffect, useState } from "react";
 import {
   STRIPE_PUBLIC_KEY_HEADER,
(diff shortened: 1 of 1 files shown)


app/client/components/payment/update/card/cardInputForm.tsx
```diff
@@ -64,7 +67,17 @@ export const CardInputForm = (props: CardInputFormProps) => {
...
+    import("@stripe/stripe-js").then(StripeAsync => {
+      setStripePromise(StripeAsync.loadStripe(props.stripeApiKey));
+    });
+  };
+
   const loadSetupIntent = (recaptchaToken: string) => {
+    if (stripePromise === null) {
+      loadStripe();
+    }
+
```
twrichards, line 80: this is not synchronous? think you might have a race condition?

app/client/components/payment/update/card/cardInputForm.tsx
```diff
@@ -38,7 +38,10 @@ export const CardInputForm = (props: CardInputFormProps) => {
   >();
   const [stripeSetupIntentError, setstripeSetupIntentError] = useState<Error>();
 
-  const stripePromise = loadStripe(props.stripeApiKey);
```
twrichards, line 41: could this not have been moved into a `useEffect` instead?

twrichards (commented):
Is this approach followed from somewhere?

app/client/components/payment/update/card/cardInputForm.tsx
rBangay: > this is not synchronous? think you might have a race condition?

I've taken some screenshots in order to explain the approach here ...

The first image is the default route (account overview) on the existing master branch
01_master


The 2nd image is if we move the loadStripe call within the useEffect hook but keep the existing stripe imports (again on the default route - account overview):
02_loadstripeInUseEffect

The 3rd and 4th images are the approach taken in this pr (the 1st image being the default route and the 2nd being the update payment page after capthca is complete)
03_asyncLoadStripeImport

update payment page:
04_asyncLoadStripeImport

twrichards (approved):
Great work! Thanks for talking it through with me IRL. LGTM 👍 

@paulbrown1982 will be pleased.

app/client/components/payment/update/card/cardInputForm.tsx
twrichards: thanks for explaining in the other answer :)

PR #485, opened 2020-09-01
[RB] - add GA tracking for invoice interactions

rBangay:
## What does this change?
adds GA tracking for the view and download invoice links, clicks on the invoice pagination, and the selection of a year from the invoice year select.

The event labels are as follows:

pagination event: `invoice_pagination_select`
view pdf tracking event: `view_XXXX_pdf_invoice` eg `view_homedelivery_pdf_invoice`
download pdf tracking event: `download_XXXX_pdf_invoice` eg `download_homedelivery_pdf_invoice`
year select event: `invoice_year_select`

## Images
90876239-3ca2ea80-e39a-11ea-8a23-c6ab2bbfd407

### DIFF
--- a/app/client/components/billing/billing.tsx
+++ b/app/client/components/billing/billing.tsx
@@ -127,7 +127,8 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                       ...invoice,
                       pdfPath: `/api/${invoice.pdfPath}`,
                       currency: paidPlan.currency,
-                      currencyISO: paidPlan.currencyISO
+                      currencyISO: paidPlan.currencyISO,
+                      product: specificProductType.productTitle(mainPlan)
                     }));
                   const resultsPerPage = paidPlan.interval?.includes("year")
                     ? productInvoiceData.length
--- a/app/client/components/billing/billing.tsx
+++ b/app/client/components/billing/billing.tsx
@@ -128,7 +128,7 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                       pdfPath: `/api/${invoice.pdfPath}`,
                       currency: paidPlan.currency,
                       currencyISO: paidPlan.currencyISO,
-                      product: specificProductType.productTitle(mainPlan)
+                      productTitle: specificProductType.urlPart
                     }));
                   const resultsPerPage = paidPlan.interval?.includes("year")
                     ? productInvoiceData.length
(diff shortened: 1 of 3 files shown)


app/client/components/billing/invoicesTable.tsx
```diff
@@ -293,7 +301,20 @@ export const InvoicesTable = (props: InvoicesTableProps) => {
...
+                      css={invoiceLinkCss}
+                      href={tableRow.pdfPath}
+                      onClick={() => {
+                        return trackEvent({
+                          eventCategory: "invoice",
+                          eventAction: "click",
+                          eventLabel: `view_${tableRow.product?.replace(
+                            /\s+/g,
+                            ""
+                          )}_pdf_invoice`
```
twrichards, line 314: instead of doing a replace here, you could use `urlPart` of the `ProductType` (or rather `specificProductType`) - which would be more consistent with other GA events

app/client/components/billing/billing.tsx
```diff
@@ -127,7 +127,8 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                       ...invoice,
                       pdfPath: `/api/${invoice.pdfPath}`,
                       currency: paidPlan.currency,
-                      currencyISO: paidPlan.currencyISO
+                      currencyISO: paidPlan.currencyISO,
+                      product: specificProductType.productTitle(mainPlan)
```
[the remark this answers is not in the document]
twrichards: but based on my other comment, better would be to pass `urlPart` via a key named something like `productUrlPart`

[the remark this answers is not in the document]
twrichards: although, thinking more about it, does the `productTitle` function expose the rate plan name for certain products (e.g. Everyday, Sixday, Weekend etc.)

app/client/components/billing/billing.tsx
```diff
@@ -127,7 +127,8 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                       ...invoice,
                       pdfPath: `/api/${invoice.pdfPath}`,
                       currency: paidPlan.currency,
-                      currencyISO: paidPlan.currencyISO
+                      currencyISO: paidPlan.currencyISO,
+                      productTitle: specificProductType.urlPart
```
twrichards, line 131: if you're going to use `urlPart`, either call it 
```suggestion
                      productUrlPart: specificProductType.urlPart
```
...OR something like...
```suggestion
                      productCode: specificProductType.urlPart
```

app/client/components/billing/invoicesTable.tsx
```diff
@@ -31,6 +32,7 @@ interface InvoicesTableProps {
 }
 
 export const InvoicesTable = (props: InvoicesTableProps) => {
+  let trackingPaginationInteractionCount = 1;
```
twrichards, line 35: do we definitely need a mutable variable? should it not be a piece of state rather than a `let`

ripecosta (approved):
LGTM 👍

PR #496, opened 2020-09-11
Enforcing Yarn’s version

mario-galic:
## What does this change?

* Make sure everyone uses the same yarn version
* Recommended as per https://classic.yarnpkg.com/en/docs/cli/policies/#toc-policies-set-version
* Commits yarn to repository under `.yarn/releases`
* Fix synk https://app.snyk.io/org/the-guardian-cuu/project/49b0e67e-be08-4894-9c16-171e72117b99#issue-SNYK-JS-YARN-557182
* Similar PR: https://github.com/guardian/dotcom-rendering/pull/1322
* I had to execute `dos2unix .yarn/releases/yarn-1.22.5.js` before committing otherwise pre-hooks failed

## How to test

* execute yarn from within `app/`
* Run TC builld


twrichards:
This yarn policies thing is great, good find.

In that it's making the `yarn` binary part of the repo, could the changes to the build script instead be to use that yarn binary from the repo rather than downloading it each time (of a version that's hard-coded separately from the new `.yarnrc` file).

rBangay (approved):

mario-galic:
[SNYK-JS-NODEFETCH-674311](https://snyk.io/vuln/SNYK-JS-NODEFETCH-674311) was published today which requires updating node-fetch to 2.6.1. However [`react-daterange-picker`](https://github.com/onefinestay/react-daterange-picker) seems inactive and cannot be updated to appropriate version. I have delayed the vulnerability for 30 days via `.snyk` config.

build-tc
```diff
@@ -2,22 +2,7 @@
...
-YARN_LOCATION="$(pwd)/tools/${YARN_VERSION}"
-
-if [ ! -d "$YARN_LOCATION" ]; then
-	mkdir -p ${YARN_LOCATION}
-	cd ${YARN_LOCATION}/
-	wget -qO- https://github.com/yarnpkg/yarn/releases/download/v${YARN_VERSION}/yarn-v${YARN_VERSION}.tar.gz | tar zvx
-	cd ../..
-fi
-
-export PATH="$PATH:$YARN_LOCATION/yarn-v$YARN_VERSION/bin"
```
twrichards, line 17: I'd be surprised if the `yarn` commands further down this script are actually running the committed version (and not some random version that happens to be on the build agent)... I'd probably keep this `PATH` line and set the location to be the committed one.

mario-galic: According to https://classic.yarnpkg.com/en/docs/cli/policies/#toc-policies-set-version

> yarn policies set-version offers a simple way to check in your Yarn release within your repository. Once you run it, your configuration will be updated in such a way that anyone running a Yarn command inside the project will always use the version you set - and this transparently.

Since we do `cd app` before running `yarn`  I think we are inside the project.

twrichards: interesting, is there a way we could prove it? e.g. by running `yarn -v` and `type -a yarn` in the build script

mario-galic: TC build logs already indicates correct version (unless it is a coincidence agent has the same version):

image

rBangay (approved):

build-tc
mario-galic: It looks up location of yarn on the basis `yarn-path` in `.yarnrc`

```
yarn-path ".yarn/releases/yarn-1.22.5.js"
```

mario-galic: If we wanted to be rally sure we could do in `build-tc` script something like 

```
cd app
alias yarn ='node ../.yarn/releases/yarn-*.js'
```

twrichards:
was there a last check of this on CODE? my guess is that it's not build a valid artifact, so boxes are probably cycling?

mario-galic:
@twrichards It never deployed. Could not find yarn on agent.

build-tc
sndrs: any version of yarn <1 should actually use the file referenced in `.yarnrc` if it finds it

PR #497, opened 2020-09-14
Do not depend on yarn installed on TC agent even for bootstraping

mario-galic:
## What does this change?

Related PR: https://github.com/guardian/manage-frontend/pull/496

Some TC agents apparently did not have global yarn installed which is needed to boostrap local yarn under `.yarn/releases/` via `yarn-path` in `.yarnrc`. So now we do not depend on agent's installation at all (although this is non-standard way of doing it).

## How to test

- Try to run TC build
- Check build log output for correct version of yarn

### DIFF
--- a/build-tc
+++ b/build-tc
@@ -1,9 +1,14 @@
 #!/bin/bash
+shopt -s expand_aliases
 
 set -e
 
 # cd into the app directory where package.json resides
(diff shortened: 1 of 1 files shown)


ripecosta (approved):
👍

PR #502, opened 2020-09-24
Update iframe min-height for new sign up content.

buck06191:
## What does this change?

Change the min height of the iframe element used for newsletter signup in the footer. Needed because the iframe content has changed and now takes up more space.

## How to test

Run locally.

### Before
image

### After
image


ripecosta (approved):
👍

PR #504, opened 2020-09-28
Investigate users being re-directed back to root before flow is completed

mario-galic:
## What does this change?

* https://trello.com/c/zGUa33yd/1639-manage-frontend-redirects-to-account-overview-before-user-flow-is-completed-s
* This is just an exploratory draft PR to server as discussion point
* Currently `ProductDetailProvider` component requests `/me/mma`, which might return multiple products, and then has a guard `if (filteredProductDetails.length === 1) setSelectedProductDetail(filteredProductDetails[0])`. Perhaps instead we should be more specific, and somehow pass specific `subscriptionId` we are interested in thus `ProductDetailProvider` would request `/me/mma/A-S00060454`. [`createProductDetailFetcher`](https://github.com/guardian/manage-frontend/blob/fd7fb46cf73fab9737eca7b1c0517d10eea3ad8e/app/shared/productTypes.tsx#L215) already accounts for this possibility 
* How would we pass the `subscriptionId` down to all the components that need it? Currently just for cancellation flow I pass it via Router path parameter, which does seem to allow me to escape from the redirection loop.

### DIFF
--- a/app/client/components/accountoverview/manageProduct.tsx
+++ b/app/client/components/accountoverview/manageProduct.tsx
@@ -376,7 +376,12 @@ const CancellationCTA = (props: CancellationCTAProps) => {
         css={css`
           color: ${palette.brand["500"]};
         `}
-        to={"/cancel/" + props.specificProductType.urlPart}
+        to={
+          "/cancel/" +
+          props.specificProductType.urlPart +
+          "/" +
+          props.productDetail.subscription.subscriptionId
+        }
         state={props.productDetail}
       >
         {shouldContactUsToCancel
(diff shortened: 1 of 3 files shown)


twrichards:
I'm keen to better understand / witness this problem... happy to pair on it at some point later today or tomorrow.

This '_auto-selection_' (for when there is one sub of the given product type and someone is navigating direct to a flow, rather than when the user navigates from elsewhere in MMA - where browser history state is used to hold the selected sub) was first tackled/implemented a long time ago (although it has been reworked/simplified during the 'Account Overview' integration). At the time @JessL123 wanted to avoid having the sub id in the URL for privacy reasons, and I doubt that has changed.

mario-galic:
Closing in favour of https://github.com/guardian/manage-frontend/pull/506 Silent redirection was by-design to nudge the user to pick the correct subscription. The bug was simply due to router state being incorrectly accessed so `ProductDetailProvider` thought it does not have enough information in the router's `location.state`.

PR #506, opened 2020-09-30
Fix early redirection to root due to productDetail existing at different levels in location.state

mario-galic:
## What does this change?

`ProductDetails` is sometimes passed directly as the `state` object of Reach Router [Link](https://reach.tech/router/api/Link) component

```
<Link
  state={props.productDetail}
  ...
>
```

and other times nested inside another object 

```
<LinkButton
  state={
    {
      productDetail,
      flowReferrer: {
        title: NAV_LINKS.billing.title,
        link: NAV_LINKS.billing.link
      }
    }
  }
  ...
/>
``` 

so this PR accounts for both possibilities in `ProductDetailProvider` component. This fixes the bug where  `ProductDetailProvider` would think there is nothing in the state, and then re-request from MMA, however MMA could respond with less precise information, for example, there could be two subscriptions of the same kind, and `ProductDetailProvider` would not have enough information to know which the user has picked, so it would simply redirect back to the `Account Overview` with the intention of user picking the correct one. The user would then get stuck in redirection loop.

## How to test

To reproduce the bug acquire two subscriptions of the same kind, for example two separate subscriptions of Guardian Weekly, and then try to complete cancellation flow.

It is rare that users would have duplicate products, however this bug was also interfering with development as we often create many subscriptions during testing.

- There is no major behavioural change in this PR
- It was tested end-to-end

### DIFF
--- a/app/client/components/accountoverview/accountOverviewCard.tsx
+++ b/app/client/components/accountoverview/accountOverviewCard.tsx
@@ -281,7 +281,7 @@ export const AccountOverviewCard = (props: AccountOverviewCardProps) => {
             <LinkButton
               to={`/${groupedProductType.urlPart}`}
               text={`Manage ${groupedProductType.friendlyName}`}
-              state={{ productDetail: props.productDetail }}
+              state={props.productDetail}
               colour={palette.brand[800]}
               textColour={palette.brand[400]}
               fontWeight={"bold"}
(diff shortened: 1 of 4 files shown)


twrichards (dismissed):
Assuming we've had a quick check of all the flows in `CODE` - then 🎉

mario-galic:
@twrichards Does CODE check anything extra over completing the flow in DEV?

twrichards:
> @twrichards Does CODE check anything extra over completing the flow in DEV?

No not really, I more meant confirming that the other flows that we ended up tweaking all continue to work.

twrichards (approved):

PR #508, opened 2020-10-08
Form submission architecture decision

ripecosta:
## What does this change?
Adds an architecture decision made by the team regarding form submissions in this project.

### DIFF
--- a/docs/05-form-submission-via-server-side.md
+++ b/docs/05-form-submission-via-server-side.md
@@ -1,0 +1,27 @@
+# To submit all forms via server-side
+
+## Context
+
+Forms that do not require a login to access could be submitted directly to the service layer and bypass the server-side of this project. However, there are trade-offs with that decision.
+
+Benefits of not bypassing the server side:
+
+- Is consistent with manage-frontend’s standard approach
+- Lambda endpoints remains secret
+- Can use API key to validate requests at API Gateway level
+- Cheaper in case of DDoS attack
+- Easier to debug manage-frontend's logs VS API gateway's log
+- Allows for better room to grow in case requirements change
+
+Benefits of bypassing the server-side:
+
+- Simpler to maintain as there are less parts
+- Less latency due to 1 less hop (relatively negligible)
+
+## Decision
+
+We will submit all forms through the server-side of this project.
+
+## Status
+
+Accepted


arausuy (approved):

arausuy:
Thanks for doing that!

docs/05-form-submission-via-server-side.md
```diff
@@ -0,0 +1,27 @@
+# To submit all forms via server-side
+
+## Context
+
+Forms that do not require a login to access could be submitted directly to the service layer and bypass the server-side of this project. However, there are trade-offs with that decision.
+
+Benefits of not bypassing the server side:
+
+- Is consistent with manage-frontend’s standard approach
+- Lambda endpoints remains secret
```
kelvin-chappell, line 10: remain

docs/05-form-submission-via-server-side.md
```diff
@@ -0,0 +1,27 @@
...
+- Can use API key to validate requests at API Gateway level
+- Cheaper in case of DDoS attack
+- Easier to debug manage-frontend's logs VS API gateway's log
+- Allows for better room to grow in case requirements change
```
kelvin-chappell, line 14: Perhaps: Allows for more flexibility in case requirements change

docs/05-form-submission-via-server-side.md
```diff
@@ -0,0 +1,27 @@
...
+
+Benefits of bypassing the server-side:
+
+- Simpler to maintain as there are less parts
```
kelvin-chappell, line 18: fewer

kelvin-chappell (commented):
Just a couple of pernickety points.

Also is the fact that validation can be done on the manage-frontend server so that the lambda can assume it will receive valid input another benefit?

ripecosta:
> Just a couple of pernickety points.
> 
> Also is the fact that validation can be done on the manage-frontend server so that the lambda can assume it will receive valid input another benefit?

Thanks for you feedback, Kelvin. Apologies for the (very) late reply.
I would say that's more of an implementation choice. Although we did that for the Contact Us form it might not be the best choice in other cases. A project could choose to do all validation in the lambda so that it might serve requests from source that we don't own and therefor cannot ensure correct validation. Because this is a generalised architecture decision I would prefer to keep it as broad as possible.

PR #509, opened 2020-10-09
Update newsletter footer embed src

buck06191:
## What does this change?

Following a refactor of how the newsletter embeds will work, there won't be an iframe hosted at `profile.guardian.co.uk`. Instead I'm moving this to using `www.theguardian.com` directly, and we'll be implementing some changes to make sure that things like autofilling email addresses for logged in users will be implemented in a way that is suitable across platforms.

### Related PRs:
 - https://github.com/guardian/frontend/pull/23051 (Reverted pending some work on the router, but the code will be the same following these changes)

### DIFF
--- a/app/client/components/footer/footer.tsx
+++ b/app/client/components/footer/footer.tsx
@@ -1,19 +1,11 @@
 import { css } from "@emotion/core";
 import { from } from "@guardian/src-foundations/mq";
 import React, { SyntheticEvent, useEffect, useState } from "react";
-import { conf } from "../../../server/config";
 import palette from "../../colours";
 import { isInUSA as isUserInUSA } from "../../geolocation";
 import { headline } from "../../styles/fonts";
 import { SupportTheGuardianButton } from "../supportTheGuardianButton";
 import { footerLinks } from "./footerlinks";
-
-let domain: string;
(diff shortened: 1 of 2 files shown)


app/client/components/footer/footer.tsx
```diff
@@ -221,7 +213,7 @@ export const Footer = () => {
               <div css={emailSignUpStyles}>
                 <iframe
                   title="Guardian Email Sign-up Form"
-                  src={`https://profile.${domain}/email/form/footer/today-uk`}
+                  src={`https://www.theguardian.com/email/form/footer/today-uk`}
```
twrichards, line 216: just curious is the `www.` needed?

twrichards (approved):
LGTM 👍

app/client/components/footer/footer.tsx
buck06191: That's a good question. On frontend the source is typically just `/email/form/footer/today-uk`, which I think then defaults to using the domain of the page. 

Is there any detriment to including `www.`? It's not something I've thought about much before

buck06191: Having just checked against PROD, not including the `www.` works the same way

PR #510, opened 2020-10-14
[RB] - display cancelled products in the account overview page for 3 months

rBangay:
## What does this change?

Related PR: https://github.com/guardian/members-data-api/pull/478

Cancelled products should show in the account overview page for 3 months after they were cancelled.

A new `/cancelled` endpoint has been created (see [here](https://github.com/guardian/members-data-api/pull/478/commits/f350544476ecefaf0c9a240cc546a3ed014deabd)) which is called alongside the existing `/me/mma` members-data-api request.

A new `accountOverviewCancelledCard` component has been created to handle the response from the `/cancelled` endpoint. The response from the `/cancelled` endpoint is a slimmed down version of the `/me/mma` response and as such the component does not currently show payment information, a slightly slimmed down product title (`Newspaper Voucher` instead of `Newspaper Voucher - Everyday` for example) and no 6for6 explainer.

If the cancelled product was a subscription, then a button to subscribe again is displayed at the bottom of the product box. If the product was a contribution or a membership then a message is displayed at the bottom of the list of contributions or memberships with a custom message and cta.

## Images
Screenshot_2020-10-14 My Account The Guardian

### DIFF
--- a/app/client/components/accountoverview/accountOverview.tsx
+++ b/app/client/components/accountoverview/accountOverview.tsx
@@ -1,12 +1,13 @@
 import { css } from "@emotion/core";
 import { palette, space } from "@guardian/src-foundations";
-import { headline, textSans } from "@guardian/src-foundations/typography";
+import { headline } from "@guardian/src-foundations/typography";
 import { RouteComponentProps } from "@reach/router";
 import React from "react";
 import {
+  CancelledProductDetail,
+  isCancelledProduct,
   isProduct,
   MembersDataApiItem,
-  MembersDatApiAsyncLoader,
   ProductDetail,
   sortByJoinDate
 } from "../../../shared/productResponse";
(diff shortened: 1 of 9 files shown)


twrichards:
Is there a reason this is a separate endpoint (and thus another Ajax call, the results of which need to be spliced/merged in to the 3 sections of Account Overview) as opposed to a query param on the main `/mma` call to request N months back of cancelled subs?

app/shared/productResponse.ts
```diff
@@ -59,12 +59,23 @@ export interface ProductDetail extends WithSubscription {
...
+
 export function isProduct(
   data: MembersDataApiItem | undefined
 ): data is ProductDetail {
   return !!data && data.hasOwnProperty("tier");
 }
 
+export const isCancelledProduct = (
+  data: CancelledProductDetail | ProductDetail | undefined
+) => !!data && !data.hasOwnProperty("joinDate");
```
mario-galic, line 77: members-data-api should probably return explicit field for this, something like, `isCancelledAndEnded`.

rBangay: ^ that would be nice :)

app/client/components/accountoverview/accountOverview.tsx
```diff
@@ -68,28 +86,43 @@ const AccountOverviewRenderer = (apiResponse: MembersDataApiItem[]) => {
...
+                    />
+                  );
+                })}
+                {(groupedProductType.groupFriendlyName === "membership" ||
+                  groupedProductType.groupFriendlyName === "contribution") &&
+                  products.some(product => {
+                    if (isCancelledProduct(product)) {
+                      return true;
+                    } else {
+                      return isCancelled(
```
mario-galic, line 117: Having now both `isCancelledProduct` and `isCancelled` seems a bit confusing. Maybe `isCancelled` should be renamed to indicate this is not fully ended product, say,`cancellationIsPending`

mario-galic:
The reason cancelled and ended subscriptions cannot easily be returned as part of standard `me/mma` response is that membership-common is designed with "current" subscriptions in mind which means there are safeguards in place to check that. For example see [GetCurrentPlans]( https://github.com/guardian/membership-common/blob/72db910ddbaf8f45920d05effdc6264dac86ef5f/src/main/scala/com/gu/memsub/subsv2/Subscription.scala#L49). Calling `sub.plan` always goes via this `GetCurrentPlans`. It would be risky to loosen these safeguards to accommodate fully ended plans as well. Hence a separate endpoint was implemented which separates this concern, which has the advantage of simpler and less risky implementation server-side, however it has the disadvantage of duplication in MMA and extra call.

app/client/components/accountoverview/accountOverview.tsx
rBangay: yep .. I'm going to factor out `isCancelledProduct` and use the separation of the active and cancelled products from the respective api calls to distinguish them

mario-galic (approved):

PR #511, opened 2020-10-19
Contact us form client

rBangay:
## What does this change?
Release of the first version of the contact us form. 

Runs off of a local config object.

The contact us form routes have been disabled in this pr awaiting future prs and full release.

## Images
Screenshot_2020-10-19 My Account The Guardian
Screenshot_2020-10-19 My Account The Guardian(1)
Screenshot_2020-10-19 My Account The Guardian(2)
Screenshot_2020-10-19 My Account The Guardian(3)
Screenshot_2020-10-19 My Account The Guardian(4)


app/client/components/contactus/contactUsConfig.ts
```diff
@@ -0,0 +1,287 @@
...
+      {
+        id: "s9",
+        name: "Something else",
+        editableSubjectLine: true
+      }
+    ]
+  },
+  {
+    id: "vouchers",
+    name: "newspaper vouchers",
```
mario-galic, line 89: Other names seem to start with upper case.

app/client/components/contactus/contactUsConfig.ts
```diff
@@ -0,0 +1,287 @@
...
+    enquiryLabel: "your newspaper voucher issue",
+    subTopicsTitle: "sub topic title here",
+    subtopics: [
+      {
+        id: "s10",
+        name: "I've lost my vouchers"
```
mario-galic, line 95: Are apostrophe abbreviations stylistically ok?

app/client/components/contactus/contactUsForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+}
+
+interface FormPayload {
+  fullName: string;
+  email: string;
+  subjectLine: string;
+  details: string;
+}
+
+interface FormElValidationObject {
```
mario-galic, line 25: Maybe say `Element` instead of `El`.

app/client/components/contactus/contactUsForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+export const ContactUsForm = (props: ContactUsFormProps) => {
+  const [subjectLine, setSubjectLine] = useState<string>(props.subjectLine);
+  const [fullName, setFullName] = useState<string>(
+    window.guardian?.identityDetails?.displayName || ""
+  );
+  const [email, setEmail] = useState<string>(
+    window.guardian?.identityDetails?.email || ""
+  );
+  const [details, setDetails] = useState<string>("");
```
mario-galic, line 46: What does empty string `""` do here?

app/client/components/contactus/contactUsForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+    }
+  });
+
+  const validateForm = () => {
+    const isFullNameValid = !!fullName.length;
+    const isEmailValid = !!email.length;
+    const isSubjectLineValid = !!subjectLine.length;
+    const isDetailsValid = !!details.length;
+    const isFormInValidState =
+      isFullNameValid && isEmailValid && isDetailsValid;
```
mario-galic, line 82: Is it not necessary to include `isSubjectLineValid`?

app/client/components/contactus/contactUsForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+    },
+    details: {
+      isValid: true,
+      message: mandatoryFieldMessage
+    }
+  });
+
+  const validateForm = () => {
+    const isFullNameValid = !!fullName.length;
+    const isEmailValid = !!email.length;
```
mario-galic, line 78: Should email validation be stronger than just length?

app/shared/pathsToAllowThroughWithoutSignin.ts
```diff
@@ -0,0 +1,22 @@
+export const pathsToAllowThroughWithoutSignin = [
+  "/contact-us-form",
+  "/contact-us-form/*",
+  "/contact-us-form/*/*",
+  "/contact-us-form/*/*/*"
+];
+
+export const allowPathThroughWithoutSignin = (path: string) => {
```
mario-galic, line 8: This seems critical? How do you know it works? Could we write some tests?

app/client/components/accountoverview/contributionUpdateAmountForm.tsx
```diff
@@ -14,7 +14,7 @@ import { trackEvent } from "../analytics";
 import AsyncLoader from "../asyncLoader";
 import { Button } from "../buttons";
 import { SuccessMessage } from "../delivery/address/deliveryAddressEditConfirmation";
-import { Input } from "../delivery/address/input";
+import { Input } from "../input";
```
mario-galic, line 17: Is it possible to give it more meaningful name than than `input`?

app/client/components/contactus/contactus.tsx
```diff
@@ -0,0 +1,388 @@
...
+interface ContactUsFormStateSnapshot {
+  selectedTopic: string | undefined;
+  selectedSubTopic: string | undefined;
+  selectedSubSubTopic: string | undefined;
+}
+
+interface ContactUsProps extends RouteComponentProps {
+  topicId?: string;
+  subTopicId?: string;
+  subSubTopicId?: string;
```
mario-galic, line 26: Seems strange to have everything optional.

app/client/components/contactus/contactUsConfig.ts
rBangay: I'll update the copy for the config file here and in a subsequent pr that moves this config over to the server

mario-galic (approved):
The biggest change seems to me `pathsToAllowThroughWithoutSigning`, so hopefully it works as intended.

app/.snyk
```diff
@@ -4,40 +4,40 @@ version: v1.14.1
 ignore:
   SNYK-JS-LODASH-590103:
     - '@emotion/core > @emotion/css > babel-plugin-emotion > @babel/helper-module-imports > @babel/types > lodash':
-        reason: No upgrade available
-        expires: '2020-10-17T17:00:51.078Z'
+        reason: No remediations available
```
mario-galic, line 7: I feel synk has become noisy and should be removed as blocker for our PRs.

app/client/components/contactus/contactUsConfig.ts
ripecosta: This is not the final copy so never mind this for now. We will update to the final copy in the next PR.

app/client/components/contactus/contactUsForm.tsx
ripecosta: It is indeed. 👍

app/.snyk
ripecosta: Agree. We have processes to check it regularly, so I believe it's best to change to to warn us instead of block us.

app/client/components/accountoverview/contributionUpdateAmountForm.tsx
rBangay: This is for an input component and follows design systems naming convention

app/client/components/contactus/contactus.tsx
rBangay: Hey Mario, we've renamed these props to better reflect what they do. They are only used if the user is deep linking into the form via the url and as such all of them are optional as you can use the form without deep linking.

app/.snyk
rBangay: this will be done in a separate pr

app/client/components/contactus/contactUsForm.tsx
rBangay: we've added some more limited validation to the client

app/shared/pathsToAllowThroughWithoutSignin.ts
rBangay: we have written some tests :)

mario-galic (approved):

app/shared/requiresSignin.ts
```diff
@@ -0,0 +1,8 @@
+import * as pathLib from "path";
+
+const publicPaths = ["/contact-us-form"];
+
+export const requiresSignin = (path: string) => {
```
mario-galic, line 5: Could percent-encoding `%2e%2e` maybe trick it? I guess it gets decoded by MMA before it reaches that point.

mario-galic (approved):

app/shared/requiresSignin.ts
ripecosta: Tested it locally and `%2e%2e` does not trick it. Seem like `normalize` takes care of it.

mario-galic: `path.normalize` does not handle `%2e%2e`. If it was decoded then it must have happened higher up.

ripecosta: Looked into it a bit closer and you are right. I used a jest test and the reason it doesn't trick it there is because `%2e%2e` is read as a string instead of being replaced by `..`. Tried to use it to access an authorised path from the browser but it also doesn't trick it. It seems like it is being decoded somewhere higher up as you said.

PR #512, opened 2020-10-21
Load Contact Us form config from an endpoint

ripecosta:
## What does this change?
Changes how the Contact Us form loads its config. Previously it read it from a variable that was bundled with the client-side. Now it fetches that config from an endpoint that is exposed in the server-side.

This is part of a larger piece of work to release the [Contact Us form](https://trello.com/c/aCi04YJT/1632-integrate-frontend-and-lambda-work-for-contact-us-form-s) and it expands on the code already merged in #511.

The Contact Us form routes are disabled for now until this feature is ready for release.

## Motivation
Currently the config is bundled with the client-side, meaning it is loaded every time a user accesses manage-frontend. This carries an unnecessary hit to performance for all other areas on the website. With this change, only users who access the Contact Us form will load the config, avoiding the performance cost for the rest of the website.

## Example
https://github.com/guardian/manage-frontend/blob/878350c8188e10498863fd584b8213ddbea4b983/app/client/components/contactus/contactUsConfig.ts#L26-L287


app/client/components/contactUs/contactUs.tsx
```diff
@@ -233,156 +246,166 @@ export const ContactUs = (props: ContactUsProps) => {
 
   return (
     <>
-      <ContactUsHeader />
-      <ContactUsPageContainer>
-        <div
+      <div
```
ripecosta, line 249: This seems like a huge change but it was just a move of `<ContactUsHeader />` and `<ContactUsPageContainer>` from here into `ContactUsPage ` that caused a re-identation of the code inside this return. It apparently confused GitHub in the comparison.

twrichards:
Having used `yarn bundle-check-client`...
| `main` | `rc/contact-us-config-endpoint` (this branch) |
| --- | --- |
| image | image 0.3% reduction |
... as you can see moving this to an endpoint is a minimal saving in bundle size (although great you've been keeping that in mind 🥇 ). So unless you think this config will become huuuge in future, personally I don't think it's worth making this an endpoint given the complexity & reliability challenges of making AJAX calls etc... as such I don't feel comfortable giving a 👍, sorry.

twrichards:
If you think bundle size is becoming an issue more generally (it's certainly not lightweight right now), I would recommend a wider rethink and a more wholesale approach to addressing it... something like code splitting, such that there is say a 'core' bundle then additional bundles for different areas of the site (e.g. 'contact us', 'holiday stops', 'delivery records' etc. Informed by the output of `yarn bundle-check-client`...
image

There is https://github.com/guardian/manage-frontend/wiki/Major-TODOs#bundle-size-review-purge-things-plus-code-splitting-and-lazy-loading (apologies I never got round to leaving any notes for it)... but would make a good health task.

ripecosta:
Following @twrichards comments and further discussions this PR is being closed in favour of #515. Bundle size will be addressed more generally in the project by using code splitting techniques.

PR #513, opened 2020-10-23
Contact Us form submits to an endpoint

ripecosta:
## What does this change?
This PR creates an endpoint (`/api/contact-us/`) and changes the Contact Us form so it submits to that endpoint.

This is part of a larger piece of work to release the [Contact Us form](https://trello.com/c/aCi04YJT/1632-integrate-frontend-and-lambda-work-for-contact-us-form-s) and it expands on the code in #511 and #512.

The Contact Us form routes are disabled for now until this feature is ready for release.

### DIFF
--- a/app/client/components/contactUs/contactUs.tsx
+++ b/app/client/components/contactUs/contactUs.tsx
@@ -3,7 +3,7 @@ import { Button } from "@guardian/src-button";
 import { palette, space } from "@guardian/src-foundations";
 import { headline, textSans } from "@guardian/src-foundations/typography";
 import { RouteComponentProps } from "@reach/router";
-import * as Sentry from "@sentry/browser";
+import { captureException } from "@sentry/browser";
 import React, { useEffect, useState } from "react";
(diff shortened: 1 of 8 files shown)


app/client/components/contactUs/contactUs.tsx
```diff
@@ -3,12 +3,13 @@ import { Button } from "@guardian/src-button";
 import { palette, space } from "@guardian/src-foundations";
 import { headline, textSans } from "@guardian/src-foundations/typography";
 import { RouteComponentProps } from "@reach/router";
+import * as Sentry from "@sentry/browser";
```
rBangay, line 6:
```suggestion
import { captureException } from "@sentry/browser";
```

rBangay: it looks like `captureException` is the only part of Sentry that you are using here, so perhaps we could just import that and call it like this:
`captureException(errorMsg);`

instead of:
`Sentry.captureException(errorMsg);`

ripecosta:
Closing in favour of #515

PR #514, opened 2020-10-23
Standardise event tracking info for Contact Us form

ripecosta:
## What does this change?
This PR standardises the information submitted to GA in the Contact Us form for improved clarity and easier reporting.

This is part of a larger piece of work to release the [Contact Us form](https://trello.com/c/aCi04YJT/1632-integrate-frontend-and-lambda-work-for-contact-us-form-s) and it expands on the code in #511, #512 and #513.

PR #515, opened 2020-10-29
 Contact Us form submits 

ripecosta:
## What does this change?
Creates an endpoint (`/api/contact-us/`) that the Contact Us form will submit into. The endpoint will validate the request and forward it to the [contact-us-api lambda](https://github.com/guardian/support-service-lambdas/tree/main/handlers/contact-us-api), wait for the reply, and forward that back to the client.

It also:
- Renames some folder/file names to comply with the project's standard
- Makes the `requiresSignin` more strict to avoid unintended behaviours that would create vulnerabilities if certain conditions were met 
- Standardises the information submitted to GA in the Contact Us form for improved clarity and easier reporting.
- Upgrades the `src-button` library to 2.5.0 to avoid a React complaining about lack of unique keys inside lists. This meant a small change to how this component is used in our project (removal of the `showIcon` property)
- Creates a new `FormError` for displaying errors that can be reused in other parts of the project (PR with those changes will follow shortly)

This is part of a larger piece of work to release the [Contact Us form](https://trello.com/c/aCi04YJT/1632-integrate-frontend-and-lambda-work-for-contact-us-form-s) and it expands on the code already merged in #511. It replaces PR #512, #513 and #514.

The new `/api/contact-us/` endpoint expects requests with the the following format:
```
{
    "topic": string,
    "subtopic"?: string,
    "subsubtopic"?: string,
    "name": string,
    "email": string,
    "subject": string,
    "message": string,
}
```

#### Submit error message
CU submit error

#### Self-service message with no form
CU no form self service box


app/shared/contactUsTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+
+interface SelfServiceBox {
+  text: string[];
+  linkText: string;
+  href: string;
+}
+
+export interface SubTopic extends BaseTopic {
+  subsubtopics?: BaseTopic[];
+  subsubTopicsTitle?: string;
```
kelvin-chappell, line 17: Why is the title not part of the `BaseTopic`?

app/server/contactUsApi.ts
```diff
@@ -0,0 +1,135 @@
...
+import fetch from "node-fetch";
+import { contactUsConfig } from "../shared/contactUsConfig";
+import { ContactUsReq } from "../shared/contactUsTypes";
+import { getContactUsAPIHostAndKey } from "./apiGatewayDiscovery";
+import { log } from "./log";
+
+export const contactUsFormHandler = async (req: Request, res: Response) => {
+  const validBody = parseAndValidate(req.body);
+  if (!validBody) {
+    return res.status(400).send();
```
kelvin-chappell, line 12: Can you give more information about why it wasn't valid or is that deliberate?

app/server/contactUsApi.ts
```diff
@@ -0,0 +1,135 @@
...
+    return false;
+  }
+
+  return true;
+};
+
+const isEmail = (email: string): boolean => {
+  const emailSplit = email.split("@");
+  return emailSplit.length === 2 && emailSplit[1].includes(".");
+};
```
kelvin-chappell, line 121: Seems quite like a liberal definition of an email address

app/server/contactUsApi.ts
```diff
@@ -0,0 +1,135 @@
...
+  body.topic &&
+  validateTopics(body.topic, body.subtopic, body.subsubtopic) &&
+  body.name &&
+  body.email &&
+  isEmail(body.email) &&
+  body.subject &&
+  body.message;
+
+const validateTopics = (
+  reqTopic: string | undefined,
```
kelvin-chappell, line 71: Isn't that the same type as a `reqTopic?: string`

kelvin-chappell (approved):
I had a few trivial questions but it all looks good to me 👍 

(If it could have been broken down a bit, it would be easier to spot potential problems and get a quick review!)

app/server/contactUsApi.ts
ripecosta: This was deliberate. If the body isn't valid it means we have a bug in our code so there's nothing the user can do about it.

ripecosta: We did however forgot to log this error to Sentry. I will add it now.

app/server/contactUsApi.ts
ripecosta: Yes, it is. However the question mark implies that the argument is optional when it's not quite the case here. It's more that we don't know at this point what the contents of these variables are (or their types). But I suppose a better representation of that is to make this of type `unknown`. I will make this change now.

app/server/contactUsApi.ts
ripecosta: I'll change it to a better one!

app/shared/contactUsTypes.ts
ripecosta: This variables stores the copy to be shown on the title of the box surrounding the subtopic and subsubtopic options. The topic does not have such a box and so it's not present in BaseTopic.

kelvin-chappell (approved):
Looks good to me! 👍

PR #516, opened 2020-10-30
[RB] - enforce blocking validation of the contribution update amount …

rBangay:
## What does this change?
Enforce blocking validation of the contribution update amount form for the minimum amount so that you can't submit the form with a value less than the minimum

## Images
Screenshot 2020-10-30 at 15 33 20

Screenshot 2020-10-30 at 15 39 43


twrichards:
there were tests around the original contribution amount changer, perhaps they were discarded in the re-write (part of Account Overview) changes - they probably need to be reinstated to cover these scenarios.

_whilst I'm not a huge fan of tests this a text-book use for them, and is the reason the original tests for this were among the the fairly minimal set of tests in this project_

ripecosta (approved):
Loving those tests!

PR #517, opened 2020-11-02
Update minimum contribution update amounts

kelvin-chappell:
We now accept minimum contributions of 1 GBP and 2 AUD.


kelvin-chappell:
We don't accept minimum recurring contributions at these levels.

PR #518, opened 2020-11-05
Contact us form captcha

rBangay:
## What does this change?
Adds a re-captcha gate to the contact us form.

## Images
Screenshot 2020-11-05 at 10 07 46

Screenshot 2020-11-05 at 10 03 29

### DIFF
--- a/app/server/contactUsApi.ts
+++ b/app/server/contactUsApi.ts
@@ -6,6 +6,7 @@ import { ContactUsReq } from "../shared/contactUsTypes";
 import { isEmail } from "../shared/validationUtils";
 import { getContactUsAPIHostAndKey } from "./apiGatewayDiscovery";
 import { log } from "./log";
+import { recaptchaConfigPromise } from "./recaptchaConfig";
 
 export const contactUsFormHandler = async (req: Request, res: Response) => {
   const validBody = parseAndValidate(req.body);
@@ -30,10 +31,10 @@ export const contactUsFormHandler = async (req: Request, res: Response) => {
     body: JSON.stringify(validBody),
     headers: {
       "Content-Type": "application/json",
-      "x-api-key": apiConfig.apiKey
-    }
+      "x-api-key": apiConfig.apiKey,
+    },
   })
-    .then(contactUsAPIResponse => {
+    .then((contactUsAPIResponse) => {
       if (!contactUsAPIResponse.ok) {
         const errorMessage = `Unexpected error from contact-us-api endpoint. ${contactUsAPIResponse.status} ${contactUsAPIResponse.statusText}`;
         log.error(errorMessage);
@@ -41,7 +42,7 @@ export const contactUsFormHandler = async (req: Request, res: Response) => {
       }
       res.status(contactUsAPIResponse.status).send();
     })
-    .catch(error => {
+    .catch((error) => {
       const errorMessage =
         "Unexpected error when trying to contact contact-us-api endpoint.";
       log.error(errorMessage, error);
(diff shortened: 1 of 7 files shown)


ripecosta (approved):
👍

ripecosta (changes requested):
The reCaptcha key needs to be submitted with the rest of the form so it can be validated server-side. Otherwise we will be vulnerable to spamming if someone sends messages directly to our endpoint.

rBangay:
> The reCaptcha key needs to be submitted with the rest of the form so it can be validated server-side. Otherwise we will be vulnerable to spamming if someone sends messages directly to our endpoint.

The form should now pass along the captcha token `captchaToken` along with the rest of the form data

app/server/contactUsApi.ts
```diff
@@ -65,7 +65,9 @@ const validateContactUsFormBody = (body: any): boolean =>
   body.email &&
   isEmail(body.email) &&
   body.subject &&
-  body.message;
+  body.message &&
+  body &&
```
ripecosta, line 69: Line 61 already checks `body` so this one is not needed.
```suggestion
```

app/server/contactUsApi.ts
```diff
@@ -65,7 +65,9 @@ const validateContactUsFormBody = (body: any): boolean =>
...
+  body.captchaToken;
```
ripecosta, line 70: The token itself also need to be validated here.

ripecosta (changes requested):

app/server/contactUsApi.ts
```diff
@@ -50,18 +51,39 @@ export const contactUsFormHandler = async (req: Request, res: Response) => {
...
 
-const parseAndValidate = (body: any): ContactUsReq | undefined => {
+const parseAndValidate = async (
+  body: any
+): Promise<ContactUsReq | undefined> => {
   try {
     const bodyAsJson = body ? JSON.parse(body) : "{}";
 
-    return validateContactUsFormBody(bodyAsJson)
+    const passedCaptchaValidation = await validateCaptchaToken(
```
ripecosta, line 60: At this point in the execution it still hasn't been confirmed that `bodyAsJson.captchaToken` exists. `validateContactUsFormBody` is the validate step of this function, so it makes more sense to validate the captcha inside that function.

app/server/contactUsApi.ts
```diff
@@ -50,18 +51,39 @@ export const contactUsFormHandler = async (req: Request, res: Response) => {
...
+    return validateContactUsFormBody(bodyAsJson) && passedCaptchaValidation
       ? buildContactUsReqBody(bodyAsJson)
       : undefined;
   } catch (error) {
     return undefined;
   }
 };
 
+const validateCaptchaToken = async (token: string) => {
+  const recaptchaSecret = (await recaptchaConfigPromise)?.secretKey;
```
ripecosta, line 72: Can we fail the check before making the validation fetch if `recaptchaSecret` is undefined?

app/server/recaptchaConfig.ts
```diff
@@ -0,0 +1,10 @@
+import { s3ConfigPromise } from "./awsIntegration";
+
+interface ReCaptchaKeys {
+  publicKey: string;
+  secretKey: string;
+}
+
+export const recaptchaConfigPromise = s3ConfigPromise<ReCaptchaKeys>()(
```
ripecosta, line 8: Should we log an error in Sentry if the promise comes back with 'undefined'?

ripecosta (changes requested):

ripecosta (approved):
👍

PR #519, opened 2020-11-06
[RB] - input component should only return a number on update if the i…

rBangay:
## What does this change?
input component should only return a number on update if the input type is number, otherwise it simply returns target.value

## How to test
The input fields in the contact us form (Name, email) should now be editable. Similarly the contribution "other" amount input field should also be editable (example where the input update should return a number) 

## Images
bugpr


ripecosta (approved):
Nice catch!

PR #520, opened 2020-11-06
Refactor Contact Us form components

ripecosta:
## What does this change?
This PR refactors the Contact Us React components. The goal is to make the main Contact Us component more lean.

## Why?
The main Contact Us component was getting quite big. This made it more difficult to work on. It also affects the ease with which someone new to the code would be able to pick it up. Both of these negatively impact the team's speed of delivery.

## How?
The goal was achieved mainly by delegating minor responsibilities from the main component to its children. This opened up the possibility for some abstraction at the main component level, which in turn allowed the simplification of a big chunk of its logic.

## Results
The main Contact Us component went from **_~450 lines_** of code to only **_~285 lines_**. Components are better encapsulated and provide better abstraction.

## Other improvements in this PR
- Removes the use of indexes and non-unique values as React component keys according to [React recommendations](https://reactjs.org/docs/lists-and-keys.html#keys)
- Some variables now have more meaningful names

### DIFF
--- a/app/client/components/contactUs/contactUs.tsx
+++ b/app/client/components/contactUs/contactUs.tsx
@@ -3,7 +3,7 @@ import { palette, space } from "@guardian/src-foundations";
 import { headline, textSans } from "@guardian/src-foundations/typography";
 import { RouteComponentProps } from "@reach/router";
 import { captureException } from "@sentry/browser";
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

rBangay (approved):

PR #523, opened 2020-11-10
[RB] - use https instead of http

rBangay:
## What does this change?
use https instead of http for the contact us api url. The load balancer only supports https.

No more timeouts when submitting the contact us form


twrichards:
As communicated over chat...
> I strongly recommend you refactor to use the existing mechanism. As I said you get a lot of good stuff for free (logging, metrics etc.) and it's fairly battle tested.

> More than happy to pair on the refactors (to relax the identity requirements such that it can be used for this more anonymous contact-us-api use case)

ripecosta (approved):
👍

PR #524, opened 2020-11-11
Contact us form attachment

rBangay:
## What does this change?
add the ability to attach an image to the contact us form submission

## Images
Screenshot 2020-11-11 at 15 18 18

Screenshot 2020-11-11 at 15 19 08

Screenshot 2020-11-11 at 15 18 47

### DIFF
--- a/app/client/components/contactUs/contactUs.tsx
+++ b/app/client/components/contactUs/contactUs.tsx
@@ -5,7 +5,6 @@ import { navigate, RouteComponentProps } from "@reach/router";
 import { captureException } from "@sentry/browser";
 import React from "react";
 import { contactUsConfig } from "../../../shared/contactUsConfig";
-import { base64FromFile } from "../../../shared/fileUploadUtils";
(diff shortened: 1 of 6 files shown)


app/client/components/contactUs/contactUs.tsx
```diff
@@ -105,7 +106,13 @@ export const ContactUs = (props: ContactUsProps) => {
       email: formData.email,
       subject: formData.subjectLine,
       message: formData.details,
-      captchaToken: formData.captchaToken
+      captchaToken: formData.captchaToken,
+      ...(formData.fileAttachment && {
+        attachment: {
+          name: formData.fileAttachment.name,
+          contents: await base64FromFile(formData.fileAttachment)
```
ripecosta, line 113: Can we move this logic into the `ContactUsForm` component to improve the separation of concerns? That way all the `ContactUs` component needs to concern itself with is to submit whatever it receives from `ContactUsForm`.

ripecosta (approved):
👍

PR #525, opened 2020-11-11
Rework page flow tweak

ripecosta:
## What does this change?
This PR improves a tweak that was introduced in #133 to make all page navigations scroll to the top (as they would if it wasn't a single page app). It is now smoother. It also tweaks `@reach/router` settings so we can make the most of its accessibility features.

## Details
When a user navigates to a different page within the SPA `@reach/router` will focus the new component. This needs to happen for accessibility reasons (eg: not confusing screen readers). When this components is focused the browsers scrolls to it.

Because we're adding our header outside the client side routing (`Main` component) when the router focuses the element the scroll will hide part of the page. PR #133 fixes this by scrolling the page back to top shortly after navigation.
In cases where the user comes from a long page that has been scrolled down, the jump to the new component in the router and then to the top of the page is noticeable for the user. Removing the delay and moving the code to the end of the page fixes this visual glitch. I also added a way to easily declare paths as exceptions to this behaviour.

Because we have two client side routers we also weren't taking advantage of the accessibility features that `@reach/router` offers by default. This is because upon page load the `@reach/router` will focus on the component returned by the first router (our actual content) and then immediately on the component returned by the second router (our consent banner). If there is no consent banner this focuses a non-visible component at the bottom of the page, misleading accessibility software. Setting the `primary` property on the second router to `false` will prevent it from focusing on the component it returns.

The problem then becomes that screen readers will not focus on the consent banner when it is displayed anymore. Having the consent banner manage its own focus and focus itself when the banner is displayed fixes that too.

## Screenshots

| Before        | After           |
| ------------- |:-------------:|
| mmajumpglitch | mmajumpglitchfixed |

## Other issues
We currently do not focus the consent banner when we show a Sourcepoint banner (US users, for now). This should be looked at more closely when we update to using only the Sourcepoint banner.

### DIFF
--- a/app/client/components/analytics.tsx
+++ b/app/client/components/analytics.tsx
@@ -130,10 +130,6 @@ export class AnalyticsTracker extends React.PureComponent<{}> {
       <Location>
         {({ location }) => {
           if (location && typeof window !== "undefined") {
-            // tslint:disable-next-line:no-object-mutation
-            document.body.scrollTop = 0; // For Safari
-            // tslint:disable-next-line:no-object-mutation
-            document.documentElement.scrollTop = 0; // For Chrome, Firefox, IE and Opera
             if (
               window.guardian &&
               window.guardian.ophan &&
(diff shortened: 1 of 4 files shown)


twrichards:
Great work 🙌

Awesome PR description!!

rBangay (approved):

PR #526, opened 2020-11-12
[RB] - don not hide the captha box after the user has ticked it

rBangay:
## What does this change?
Remove timeout function that removed the recaptha form 1 second after the user had clicked on it.

Now the recaptcha should always be visible.

## Images
Screenshot 2020-11-12 at 09 28 54


ripecosta (approved):
👍

PR #527, opened 2020-11-12
Contact Us form deeplinking

ripecosta:
## What does this change?
This PR makes changes to the `ContactUs` component such that it uses the URL parameters instead of an internal state. This causes the URL to update as the uses chooses their topic/subtopic/subsubtopic and allows passing that URL to users to link them directly to a specific choice.

### DIFF
--- a/app/client/components/contactUs/contactUs.tsx
+++ b/app/client/components/contactUs/contactUs.tsx
@@ -226,8 +226,8 @@ export const ContactUs = (props: ContactUsProps) => {
                   showContacts={!showForm}
                   topicReferer={
                     `${currentTopic?.id} - ` +
-                    `${currentSubSubTopic?.id} - ` +
-                    `${currentSubSubTopic?.id}`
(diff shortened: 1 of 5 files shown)


app/client/components/contactUs/contactUs.tsx
```diff
@@ -237,9 +225,9 @@ export const ContactUs = (props: ContactUsProps) => {
                   linkAsButton={!showForm}
                   showContacts={!showForm}
                   topicReferer={
-                    `${contactUsState.topicId} - ` +
-                    `${contactUsState.subTopicId} - ` +
-                    `${contactUsState.subSubTopicId}`
+                    `${currentTopic?.id} - ` +
+                    `${currentSubSubTopic?.id} - ` +
+                    `${currentSubSubTopic?.id}`
```
rBangay, line 230:
```suggestion
                    `${currentSubSubTopic?.id || "N/A"} - ` +
                    `${currentSubSubTopic?.id || "N/A"}`
```

rBangay (approved):

PR #531, opened 2020-11-16
[RB] - update synk wizard config

rBangay:
Updates the synk wizard config.

Most of the dependencies  don't have a viable upgrade or patch available.

The dependencies and setting snyk to warn instead of breaking the build are to be looked at in subsequent health tasks.


ripecosta (approved):
GG NO RE

PR #532, opened 2020-11-16
Contact Us form length validation

ripecosta:
## What does this change?
Adds length validation to the Contact Us form and endpoint. [Trello card](https://trello.com/c/9UUZDKq9/1778-alarm-5xx-rate-from-contact-us-api-prod).

On the client-side it makes it so the user is unable to go over the allowed length for the name, email and subject fields.

On the endpoint it trims the name, email, subject and message fields so they do no go over the allowed length.


kelvin-chappell (approved):
👍

PR #536, opened 2020-11-20
Contact Us form fixes and tweaks

ripecosta:
## What does this change?
This PR makes a few changes to the Contact Us form.
- Renames variables so they are common across all the Contact Us components (client-side, json requests, server-side) and consolidates the types.
- Ensures selective field forwarding happens for attachment fields. Currently a malicious user can manipulate our code to add any number of of fields (spam) inside the attachment request and this would be forwarded to the contact-us-api lambda.
- increases the character limit for the subject. Longer subjects were getting clipped leading to some confusion for our CSRs.

### DIFF
--- a/app/client/components/contactUs/contactUsForm.tsx
+++ b/app/client/components/contactUs/contactUsForm.tsx
@@ -3,6 +3,7 @@ import { Button } from "@guardian/src-button";
 import { palette, space } from "@guardian/src-foundations";
 import { textSans } from "@guardian/src-foundations/typography";
 import React, { ChangeEvent, FormEvent, useEffect, useState } from "react";
+import { ContactUsFormPayload } from "../../../shared/contactUsTypes";
 import {
   base64FromFile,
   MAX_FILE_ATTACHMENT_SIZE_KB,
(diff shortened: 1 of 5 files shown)


app/client/components/contactUs/contactUsForm.tsx
```diff
@@ -70,8 +62,8 @@ export const ContactUsForm = (props: ContactUsFormProps) => {
       window.guardian?.identityDetails?.email) ||
       ""
   );
-  const [details, setDetails] = useState<string>("");
-  const [detailsRemainingCharacters, setDetailsRemainingCharacters] = useState<
+  const [message, setDetails] = useState<string>("");
```
rBangay, line 65: if we are renaming the state to `message` should we also rename the set state to `setMessage`?

app/client/components/contactUs/contactUsForm.tsx
```diff
@@ -70,8 +62,8 @@ export const ContactUsForm = (props: ContactUsFormProps) => {
...
+  const [messageRemainingCharacters, setDetailsRemainingCharacters] = useState<
```
rBangay, line 66: same again here, perhaps the set state and state names ought to match

app/client/components/contactUs/contactUsForm.tsx
```diff
@@ -285,25 +277,23 @@ export const ContactUsForm = (props: ContactUsFormProps) => {
...
+        {props.editableSubject ? (
           <Input
             label="Subject of enquiry"
             type="text"
-            width={50}
-            changeSetState={newSubjectLine =>
-              setSubjectLine(newSubjectLine.substr(0, 50))
-            }
-            value={subjectLine}
+            width={100}
```
rBangay, line 286: This change has changed not only the character limit but the max-width of the input field, resulting in the following on desktop breakpoints:
Screenshot 2020-11-20 at 13 35 24

rBangay (approved):

PR #537, opened 2020-11-23
Make contact us page furniture reusable

rBangay:
## What does this change?
The header and page container components in the contact us work are to be used as a template for future public facing routes on MMA. 

To accommodate this, the `ContactUsHeader` and `ContactUsPageContainer` components have been renamed to reflect there more general usage.

The left hand side of the new `SectionPageContainer` component is now narrower in keeping with the latest designs. 

There are a few more css amends to accommodate the new content widths.

## Images
Screenshot 2020-11-23 at 15 55 02


app/client/components/contactUs/contactUs.tsx
```diff
@@ -159,10 +159,38 @@ export const ContactUs = (props: ContactUsProps) => {
     }
   };
 
+  const HeaderTitle = (
+    <span>
+      Need to contact us{" "}
```
kelvin-chappell, line 164: What does the `{" "}` do?

rBangay: Hey Kelvin ... it forces a space between `Need to contact us` and `about something?` on desktop and higher breakpoints (the "about something?" part is only visible for the larger breakpoints)

kelvin-chappell: I see. Thanks!

kelvin-chappell (approved):
Looks good to me! 👍

PR #540, opened 2020-12-01
[RB] - help centre questions tracking

rBangay:
Add tracking for the help centre most popular topic questions

## Images
Screenshot 2020-12-01 at 14 47 03

### DIFF
--- a/app/client/components/helpCentre/HelpTopicBox.tsx
+++ b/app/client/components/helpCentre/HelpTopicBox.tsx
@@ -98,57 +98,55 @@ const seeAllAnchorStyle = css`
   }
 `;
 
-export const HelpTopicBox = (props: HelpTopicBoxProps) => {
-  return (
-    <div css={boxHolderStyle}>
-      <h2 css={boxTitleStyle}>
-        <i css={iconStyle}>{getHelpSectionIcon(props.topic.id)}</i>
-        {props.topic.title}
-      </h2>
-      <ul css={linksListStyle}>
-        {props.topic.links.map((faqLink, questionIndex) => (
-          <li
-            key={`${props.topic.id}Question-${questionIndex}`}
-            css={linkListItemStyle}
-          >
-            <a
-              href={faqLink.link}
-              target="_blank"
-              css={linkAnchorStyle}
-              onClick={() => {
-                trackEvent({
-                  eventCategory: "href",
-                  eventAction: "click",
-                  eventLabel: faqLink.link
-                });
-              }}
-            >
-              {faqLink.title}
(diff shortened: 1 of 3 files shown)


app/client/components/helpCentre/HelpTopicBox.tsx
```diff
@@ -98,57 +98,55 @@ const seeAllAnchorStyle = css`
...
+    >
+      <a
+        href={props.topic.seeAllLink}
+        target={"_blank"}
+        css={seeAllAnchorStyle}
+        onClick={() => {
+          trackEvent({
+            eventCategory: "help-centre",
+            eventAction: "popular-topic-see-all-click",
+            eventLabel: `${props.topic.id}-see-all`
```
ripecosta, line 144: We don't need the "-see-all" part of this, do we? Since that is already specified in the `eventAction`.

app/client/components/helpCentre/helpCentreConfig.ts
```diff
@@ -1,7 +1,7 @@
 export interface HelpCentreTopic {
   id: string;
   title: string;
-  links: Array<{ title: string; link: string }>;
+  links: Array<{ title: string; link: string; id: string }>;
```
ripecosta, line 4: Could we extract this into it's own type, to improve readability?

ripecosta (changes requested):
Do we not want to add tracking to the "Take me to the form" button at the bottom of the page?

app/client/components/helpCentre/helpCentre.tsx
```diff
@@ -86,6 +87,13 @@ export const HelpCentre = (_: HelpCentreProps) => (
           fontWeight={"bold"}
           text="Take me to the form"
           to="/contact-us/"
+          onClick={() =>
+            trackEvent({
+              eventCategory: "help-centre",
+              eventAction: "cta-click",
+              eventLabel: "contact-us-form"
```
ripecosta, line 94: To be more inline with the other events we're tracking I suggest that the `eventAction` is set to `contact-us-cta-click` and that we remove the `eventLabel`. What do you think?

app/client/components/helpCentre/helpCentreConfig.ts
```diff
@@ -1,7 +1,13 @@
+interface Link {
+  title: string;
+  link: string;
+  id: string;
```
ripecosta, line 4: Can we move `id` to the top of this interface? So that we have our unique identifiers at the start of the type.

ripecosta (approved):
👍

PR #541, opened 2020-12-04
[RB] - remove old help page content. Replace with help centre context

rBangay:
## What does this change?
Now that the help centre is live there is no need to duplicate it's content in the help page in MMA. 

This pr replaces the old content with links to popular questions, a link to the help centre and links to the call centre contact details and contact us form

## Images
Before            |  After
  |


ripecosta (approved):
Looks good! 👍 
Having the the highlighted questions defined here (instead of extracted from the main config) means we will need to remember to edit them in two places if there are any changes. But because this is a temporary solution I don't think that's a problem.

PR #542, opened 2020-12-07
[RB] - minor copy amend

rBangay:
## What does this change?
minor copy change to one of the subscription topic questions.


ripecosta (approved):
👍

PR #543, opened 2020-12-07
[RB] - update to uk call centre opening times

rBangay:
## What does this change?
Update to the UK call centre opening times


ripecosta (approved):
👍

PR #544, opened 2020-12-07
[RB] - help centre config link change and question copy change

rBangay:
## What does this change?
Updates to the help centre config:

- update the link for the `I have an Apple/iTunes subscription that I need help with` option
- update the copy for the `I need to report a bug with your app/website` option


ripecosta (approved):
👍

PR #546, opened 2020-12-11
Refactor + update contributions amount update form

TomPretty:
## What does this change?
The main point of this PR was to pull out the contributions amount update form into a separate component so we can use it during the cancellation flow. To do that `ContributionUpdateAmountForm` was renamed to `ContributionUpdateAmount` and the form part of that was pulled out into a new (albeit with an old name!) `ContributionUpdateAmountForm`. 

I also took the oportunity to update the form itself. We've now got 'frequency buttons' (e.g $3 per month) and we're using source for choice cards + text input. We've also got the weekly breakdown (e.g. Contributing £3 works out as £0.69 each week). There were also some small tweaks to padding.

## Images
Screenshot 2020-12-11 at 09 58 46
Screenshot 2020-12-11 at 09 59 03
Screenshot 2020-12-11 at 09 59 23

### DIFF
--- a/app/client/components/accountoverview/contributionUpdateAmount.tsx
+++ b/app/client/components/accountoverview/contributionUpdateAmount.tsx
@@ -32,12 +32,15 @@ export const ContributionUpdateAmount = (
   }
 
   const [status, setStatus] = useState<Status>(Status.OVERVIEW);
-  const [, setConfirmedAmount] = useState<number | null>(null);
+  const [confirmedAmount, setConfirmedAmount] = useState<number | null>(null);
+
+  const currentAmount = confirmedAmount || props.mainPlan.amount / 100;
 
   if (status === Status.EDITING) {
     return (
       <ContributionUpdateAmountForm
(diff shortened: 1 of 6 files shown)


app/client/components/accountoverview/contributionUpdateAmount.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const ContributionUpdateAmount = (
+  props: ContributionUpdateAmountProps
+) => {
+  enum Status {
+    OVERVIEW,
+    EDITING,
+    CONFIRMED
+  }
+
+  const [status, setStatus] = useState<Status>(Status.OVERVIEW);
```
TomPretty, line 34: before this was done with a couple of bools, but I thought this may be a little clearer because those bools weren't allowed to both be true

tomrf1 (approved):
This all looks great to me, but I'm new to this codebase!
@rBangay @ripecosta are you happy with this change?

PR #548, opened 2020-12-11
[RB] - 'My Account' header display

rBangay:
## What does this change?
Don't show the 'My Account' header for 'public' routes (routes that don't go through the identity sign in process)

## Images
Screenshot 2020-12-11 at 16 45 25


ripecosta (approved):
👍

PR #549, opened 2020-12-14
[RB] - update to snyk config via snyk wizard

rBangay:
## What does this change?
Update to snyk config via snyk wizard


ripecosta (approved):
👍

PR #550, opened 2020-12-14
Implement CMP lib for all regions

ripecosta:
## What does this change?
- Updates the CMP lib to v6.9.1.
- Makes the necessary to the Webpack config so that the new version the CMP lib is transpiled. See why [here](https://github.com/guardian/recommendations/blob/master/client-side.md#guardian-packages)
- Expands the use of the CMP library to all regions (previously only being used for USA).
- Removes the exception that prevented showing the CMP on the `payment/*` path.
- Renames `consentsBanner` to `CMPBanner` and converts it from class component to functional components that uses hooks.
- Removes the old banner code.
- Updates the snapshots.

Tested locally and on CODE in several different regions.


app/client/components/footer/footer.tsx
```diff
@@ -224,15 +232,25 @@ export const Footer = () => {
...
+                    {linkList.map(({ title, titleUSA, link, cmp }) => {
+                      const actualTitle = cmp && isInUSA ? titleUSA : title;
+                      const extraParams = {
+                        ...(cmp
+                          ? {
+                              href: "#",
+                              onClick: () => {
+                                if (importedCmp) {
+                                  importedCmp.showPrivacyManager();
+                                }
```
rBangay, line 244: not a biggy, but could use optional chaining and avoid the if:

`importedCmp?.showPrivacyManager();`

rBangay (dismissed):

app/client/components/footer/footer.tsx
ripecosta: Done!

rBangay (approved):

PR #552, opened 2020-12-17
Put GA behind consent

ripecosta:
## What does this change?
Puts Google Analytics behind consent. It also converts the `AnalyticsTracker` components to use hooks.

Most of the changes in the PR are related to the conversion to hooks. The parts that are specifically about putting GA behind consent are the parts inside the `useEffect` and the `cmpIsInitialised`. The behaviour now will be to wait until the first consent signal is obtained (meaning the CMP is initialised) before starting to track the page views (bit inside the component's return) and that's what the `cmpIsInitialised` state variable is used for. The way we suppress tracking events when consent is not present is by using the opt out mechanism described in [GA's documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/user-opt-out) on line 135.

Tested locally and on CODE from IPs in GB, US and AUS.

## How to test
Give or deny consent with the appropriate privacy manager ('privacy settings link at the bottom of the page'). Check the browser's network tab. When consent is denied there should no longer be any calls to the `https://www.google-analytics.com/j/collect` endpoint.

## Notes
As part of this task a bug with `@guardian/consent-management-platform` was discovered where, if a user is shown the TCFv2 banner on page load,  the `onConsentChange` callbacks are only called after a user interacts with the banner. This bug was reported in detail to comm-dev in guardian/consent-management-platform#382 who have very quickly addressed it with guardian/consent-management-platform#383. However the fix was yet live at the time for this PR. This means that until a fix is released and our version updated, GA and Ophan to be initialised only after the user has interacted with the banner, on pages where a TCFv2 banner is shown on page load. This is considerable an acceptable impact given that the fix is on its way and no tracking events happen before the user interacts with this banner, as it prevents interaction with the page until dismissed.

### DIFF
--- a/app/client/components/analytics.tsx
+++ b/app/client/components/analytics.tsx
@@ -1,5 +1,5 @@
 import { Location } from "@reach/router";
-import React, { ReactNode } from "react";
+import React, { useEffect, useState } from "react";
 import parse from "url-parse";
 import { OphanProduct } from "../../shared/ophanTypes";
 import { ProductDetail } from "../../shared/productResponse";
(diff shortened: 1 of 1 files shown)


rBangay (approved):

mxdvl:
You’ll need to upgrade to `v6.10.1`!

https://github.com/guardian/consent-management-platform/pull/383#issuecomment-748068965

app/client/components/analytics.tsx
```diff
@@ -77,85 +78,102 @@ export const applyAnyOptimiseExperiments = () => {
...
-            variant: abVariant
-          };
-        }
+      if (abName && abVariant) {
+        // tslint:disable-next-line:no-object-mutation
+        window.guardian.abTest = {
+          name: abName,
+          variant: abVariant
+        };
       }
```
paulbrown1982, line 102: Why is this code needed?

twrichards: it comes from https://github.com/guardian/manage-frontend/pull/197 (and identity thing)

PR #555, opened 2020-12-22
Add contribution amount update into the 'financial reasons' cancellation flow

TomPretty:
## What does this change?

When a user selects the 'I can no longer afford it' action, they will be presented with the option to update there contribution amount within the flow. If they update their amount, they are then taken to a new "/saved" endpoint (instead of the current "/confirmed" endpoint) that displays a thank you message + some info about their updated contribution. If they are already contributing the minimum amount, they aren't presented with the option of updating the amount.

We have a different strategy with the amount update form here compared to the general manage version (hence the new `mode` prop + corresponding type). After some discussion we decided the easiest away to avoid presenting users with choices that are larger than their current contribution would be to present them with only the other amount text field (left blank by default). 

We also made some refactors to the form (validation result is now a `string | null`) and changed the UX slightly - errors now show on change/on submit.

### review step, before hitting 'Reduce'
Screenshot 2021-01-04 at 14 15 28

### review step, after hitting 'Reduce'
Screenshot 2021-01-04 at 14 15 36

### review step, already paying min amount
Screenshot 2021-01-04 at 13 44 33

### confirmed step
Screenshot 2020-12-22 at 15 21 55


app/client/components/cancel/contributions/contributionsCancellationAmountUpdatedSaved.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const ContributionsCancellationAmountUpdatedSaved: React.FC<SavedBodyProps> = ({
+  amount
+}: SavedBodyProps) => {
+  return (
+    <MembersDataApiItemContext.Consumer>
+      {productDetail => {
+        if (!isProduct(productDetail)) {
+          Sentry.captureMessage(
+            "MembersDataApiItem is not a productDetail in ContributionsCancellationAmountUpdateSaved"
```
TomPretty, line 20: These messages may be a little verbose, perhaps we don't need the `in ContributionsCancellationAmountUpdateSaved` part?

app/client/components/cancel/contributions/contributionsCancellationFlowFinancialSaveAttempt.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const onCancelClicked = () => {
+    navigate(`mma_financial_circumstances/confirmed`);
+  };
+
+  const onReturnClicked = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+  ) => {
+    event.preventDefault();
+    navigate("/");
+  };
```
TomPretty, line 40: This felt a little hacky - I was trying to use the `LinkButton` component from `source` to get the right styles, but I don't think there's a great way to interate it with client side routing.

app/client/components/cancel/stages/savedCancellation.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { RouteableStepProps, WizardStep } from "../../wizardRouterAdapter";
+import { CancellationReason } from "../cancellationReason";
+
+const innerContent = (reason: CancellationReason, updatedAmount: number) => {
+  const onReturnClicked = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+  ) => {
+    event.preventDefault();
+    navigate("/");
+  };
```
TomPretty, line 20: Same hacky solution from before :/

app/client/components/cancel/contributions/contributionsCancellationFlowFinancialSaveAttempt.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <div css={container}>
+            <div>
+              Did you know that you can reduce the size of your contribution
+              rather than cancelling it? Simply select a new amount and we’ll do
+              the rest.
+            </div>
+
+            {showAmountUpdateForm ? (
+              <ContributionUpdateAmountForm
+                currentAmount={mainPlan.amount / 100}
```
paulbrown1982, line 73: Should this be 99.9% so that the actual current amount doesn't appear in the option set?

TomPretty: Currently we don't actually do anything to customise the option set a user sees - it's just hardcoded in the form. This also means e.g. a user contributing 2.00 a month will only see options to increase the amount (apart from `other`). We should probably come up with a better solution to this!

app/client/components/cancel/contributions/contributionsCancellationAmountUpdatedSaved.tsx
rBangay: I personally don't mind the extra context that this gives us

app/client/components/cancel/contributions/contributionsCancellationFlowFinancialSaveAttempt.tsx
```diff
@@ -0,0 +1,118 @@
...
+    navigate(`mma_financial_circumstances/saved`, { state: { updatedAmount } });
+  };
+
+  const onReduceClicked = () => {
+    setShowUpdateForm(true);
+  };
+
+  const onCancelClicked = () => {
+    navigate(`mma_financial_circumstances/confirmed`);
+  };
```
rBangay, line 35: Not a biggie at all, and not something to hold the review up for. These functions don't need block scope.

e.g:
```
const onReduceClicked = () => {
  setShowUpdateForm(true);
}
```

could become:

```
const onReduceClicked = () => setShowUpdateForm(true);
```

app/client/components/cancel/contributions/contributionsCancellationFlowFinancialSaveAttempt.tsx
rBangay: Perhaps this can be raised in a planning session and worked on separately

rBangay (approved):

app/client/components/cancel/contributions/contributionsCancellationFlowFinancialSaveAttempt.tsx
TomPretty: Yeah I'll change those!

rBangay (approved):

TomPretty:
I think there shouldn't be any more changes now! Unless something comes up during testing!

tomrf1 (approved):
great work!!

PR #557, opened 2021-01-05
Add reminder sign up at end of contribution cancellation flow

TomPretty:
## What does this change?
Update the cancellation confirmed screen for recurring contributions to give users the option to sign up for a reminder in the future. 

## Images
Screenshot 2021-02-15 at 14 06 54

Screenshot 2021-02-15 at 14 06 59

### DIFF
--- a/app/client/components/cancel/cancellationSummary.tsx
+++ b/app/client/components/cancel/cancellationSummary.tsx
@@ -12,6 +12,7 @@ import { SupportTheGuardianButton } from "../supportTheGuardianButton";
 import { WithStandardTopMargin } from "../WithStandardTopMargin";
 import { hrefStyle } from "./cancellationConstants";
 import { CancellationReasonContext } from "./cancellationContexts";
+import { CancellationContributionReminder } from "./cancellationContributionReminder";
 
 const actuallyCancelled = (
   productType: ProductType,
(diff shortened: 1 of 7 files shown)


app/client/components/cancel/cancellationSummary.tsx
```diff
@@ -12,18 +12,23 @@ import { SupportTheGuardianButton } from "../supportTheGuardianButton";
...
 const actuallyCancelled = (
   productType: ProductType,
   productDetail: ProductDetail
 ) => {
   const deliveryRecordsLink: string = `/delivery/${productType.urlPart}/records`;
   const subscription = productDetail.subscription;
+
+  const productIsContribution =
+    productType.friendlyName === "recurring contribution";
+
```
TomPretty, line 26: I wasn't sure if there was a better way to find out if the cancellation was for a recurring contribution?

twrichards: great that you've asked this question, its best to keep the flows generic and move the differing stuff into the `productTypes.tsx` (as per https://github.com/guardian/manage-frontend/wiki/Product-Types) - you'll see that convention everywhere. try to come up with a new optional property that encapsulates what you're doing (perhaps `shouldShowReminder` on `CancellationFlowProperties ` which is the type for the `cancellation` field of `ProductType`, then set it to true for just the `contributions` ProductType https://github.com/guardian/manage-frontend/blob/cfbe5f420fcf7e6312fae8b619a9ad84e7c979d4/app/shared/productTypes.tsx#L280-L298)

TomPretty: Oh nice yeah I see. I was sort of using `productIsContribution` for two things:

1 - to hide the existing summary paragraph
2 - to show the reminder instead of the resub thrasher 

I guess I can break these out into two properties?

`hideSummaryMainPara` and `shouldShowReminder`/`shouldShowReminderInsteadOfThrasher`. Alternatively that second one could be `mainBodyContent: "THRASHER" | "REMINDER" = ...`

TomPretty: Just to update here, I split it out into
`shouldHideThrasher` and `shouldShowReminder`

tomrf1 (approved):
looks good!

PR #558, opened 2021-01-07
Update feedback intro copy and remove standard save body

TomPretty:
## What does this change?
Update the copy for the feedback intro for contributions cancellations and remove the standard save body. The copy was repeated between the two. 

## Images
Screenshot 2021-01-07 at 15 43 31


paulbrown1982 (approved):
👍

PR #559, opened 2021-01-08
Add API Gateway key resolution to healthcheck

mario-galic:
## What does this change?

* https://trello.com/c/lNk6s2s8/1766-hook-up-authentication-to-healthcheck-in-manage-frontend
* Occasionally, although with seemingly increasing frequency, upon deployment we fail to fetch API Gateway credentials due to AWS rate limiting which results in inability to talk to any of the APIs in the "service layer"

    ```
    {"message":"ERROR loading host and api key for ...Rate exceeded","code":"Throttling", ...}
    ```
* This PR tries to hook up credentials resolution to the healthcheck such that instances do not spin up in the first place. Previously despite failure to load api keys it would still deploy the boxes which would result in users likely facing errors in MMA.

## How to test

My flow is usually

1. Fresh janus
1. sudo nginx
2. start-api.sh members-data-api
2. cd app
3. yarn watch
4. failure injection by direct code mutation

If the rate limit problem is hit, hopefully deployment fails, which should prompt the developer to redeploy, which should likely work, as hitting this haphazardous transient issue twice in relatively quick succession should be rather unlikely.

Without memoisation the AWS periodic healthcheck mechanism would fetch the key each time so the call is memoised such that it happens only once on deployment.

### DIFF
--- a/app/server/apiGatewayDiscovery.ts
+++ b/app/server/apiGatewayDiscovery.ts
@@ -213,3 +213,29 @@ export const getContactUsAPIHostAndKey = async () => {
 
   return { host: `https://${host}/${stage}/`, apiKey };
 };
+
+/**
+ * Used to fail riff raff deployment. Uses an arbitrarily selected service layer api as a check.
+ * Memoised in the sense of credentials fetching happens only once on deployment.
+ */
+export const memoisedApiGatewayAuthForHealthcheck = () => {
+  const memoisedHostKeyPair = getHostAndApiKeyForStack(
+    "membership",
+    "holiday-stop-api",
+    "PROD"
+  );
+  return async (
+    _: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) => {
+    const { host, apiKey } = await memoisedHostKeyPair;
+    const errMsg = `Failed to fetch authentication credentials for API Gateway service layer. Healthcheck failed!`;
+    if (!apiKey && host) {
+      next();
+    } else {
+      log.error(errMsg);
+      res.status(500).send(errMsg);
+    }
+  };
+};
--- a/app/server/apiGatewayDiscovery.ts
+++ b/app/server/apiGatewayDiscovery.ts
@@ -231,7 +231,7 @@ export const memoisedApiGatewayAuthForHealthcheck = () => {
   ) => {
     const { host, apiKey } = await memoisedHostKeyPair;
     const errMsg = `Failed to fetch authentication credentials for API Gateway service layer. Healthcheck failed!`;
-    if (!apiKey && host) {
+    if (apiKey && host) {
       next();
     } else {
       log.error(errMsg);
--- a/app/server/routes/core.ts
+++ b/app/server/routes/core.ts
@@ -1,4 +1,5 @@
-import { Response, Router } from "express";
+import { Request, Response, Router } from "express";
+import { memoisedApiGatewayAuthForHealthcheck } from "../apiGatewayDiscovery";
 import { withIdentity } from "../middleware/identityMiddleware";
 
 const router = Router();
@@ -6,7 +7,8 @@ const router = Router();
 router.get(
   "/_healthcheck",
   withIdentity(200), // healthcheck needs identity redirect service to be accessible (returns 200 if redirect required)
-  (_, res: Response) => {
+  memoisedApiGatewayAuthForHealthcheck(),
+  (_: Request, res: Response) => {
     res.send("OK - signed in");
   }
 );


app/server/routes/core.ts
```diff
@@ -1,12 +1,14 @@
...
+import { memoisedApiGatewayAuthForHealthcheck } from "../apiGatewayDiscovery";
 import { withIdentity } from "../middleware/identityMiddleware";
 
 const router = Router();
 
 router.get(
   "/_healthcheck",
   withIdentity(200), // healthcheck needs identity redirect service to be accessible (returns 200 if redirect required)
-  (_, res: Response) => {
+  memoisedApiGatewayAuthForHealthcheck(),
```
paulbrown1982, line 10: Just checking that express does definitely allow for an async function to be provided in the router chain?

paulbrown1982: Seems like it: https://itnext.io/using-async-await-to-write-cleaner-route-handlers-7fc1d91b220b

paulbrown1982 (dismissed):
👍  LGTM

app/server/apiGatewayDiscovery.ts
```diff
@@ -213,3 +213,29 @@ export const getContactUsAPIHostAndKey = async () => {
...
+/**
+ * Used to fail riff raff deployment. Uses an arbitrarily selected service layer api as a check.
+ * Memoised in the sense of credentials fetching happens only once on deployment.
+ */
+export const memoisedApiGatewayAuthForHealthcheck = () => {
+  const memoisedHostKeyPair = getHostAndApiKeyForStack(
+    "membership",
+    "holiday-stop-api",
+    "PROD"
+  );
```
twrichards, line 226: one refactor that could be made (probably not for this PR) would be to refactor the `apiNames` const further up in the list to be an array of tuples/objects containing the `stackPrefix` and `apiName` (which are currently params to the `getHostAndApiKeyForStack` function, that way for extra confidence we could perform this  check for all the known APIs (rather than an arbitrary one).

twrichards (commented):
**Nice one for picking this up - great work!**

A possible/probable explanation for the increased frequency of these is the fact that prism is under quite active development at the moment, and these rate limits apply to the whole aws account, as such if we're unlucky prism's frequent indexing has already eaten up the quota for the second that its trying to obtain the credentials.

Also worth mentioning, that because the health check fails, the riff-raff autoscaling deploy will likely make a few re-attempts itself as it cycles the boxes a few times before giving up, and as such it won't always result in a failed deploy that devs will have to manually resolve.

A complementary approach to using the SDK for these things, but still avoiding manual config would be to use [CloudFormation Exports](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html) to expose the `host` in the cfn for each of the api lambdas, which [once consumed](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html) in manage's cloudformation also creates a binding that prevents the lambdas being accidentally removed (which our current solution doesn't guard against). It probably isn't a good idea to expose the api key this way, so would need to keep most of the sdk logic for that purpose - so probably extra complexity without much benefit, but thought its worth mentioning. Also not sure if manage would require re-cloudforming to bring in those changes. _If you're interested in that pattern we used it in https://github.com/guardian/editorial-tools-platform/pull/477 to expose some stuff, then consumed it here for example https://github.com/guardian/editorial-tools-pinboard/pull/11/files#diff-8968eef82d9228dd8df9cd0c8ef6e2f0387822166ca6bd09fad8f155cef14be6R56-R58 (that's CDK btw)._

app/server/apiGatewayDiscovery.ts
```diff
@@ -213,3 +213,29 @@ export const getContactUsAPIHostAndKey = async () => {
...
+  return async (
+    _: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) => {
+    const { host, apiKey } = await memoisedHostKeyPair;
+    const errMsg = `Failed to fetch authentication credentials for API Gateway service layer. Healthcheck failed!`;
+    if (!apiKey && host) {
```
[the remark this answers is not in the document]
mario-galic: Oh dear. Just a testing artefact. Good catch!

app/server/apiGatewayDiscovery.ts
mario-galic: If one fails, I think it is likely all of them would fail. But having a list of all the api names might be generally useful even beyond the healthcheck. Separate PR.

mario-galic:
> A complementary approach to using the SDK for these things, but still avoiding manual config would be to use [CloudFormation Outputs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html)...

@twrichards This is a great idea! I do not see how exposing key in cf output increases risk. User with access to cf stacks can anyways clickthrough API Gateway and see it there. How would would the app itself (not cf) actually consume the values?

twrichards:
> > A complementary approach to using the SDK for these things, but still avoiding manual config would be to use [CloudFormation Outputs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html)...
> 
> @twrichards This is a great idea! I do not see how exposing key in cf output increases risk. User with access to cf stacks can anyways clickthrough API Gateway and see it there. How would the app itself (not cf) actually consume the values?

@mario-galic probably fair point about it not actually being any **more** risky, but we probably want to check that those outputs are not exposed anywhere obvious - they're certainly not designed for secrets (i.e. they're in plain text, where as the API Keys view in the UI is at least starred out, with a 'reveal' button).
I have no idea about consuming directly in the application (rather than ImportValue in the CFN), I suspect that would require  SDK to retrieve and in which case would likely fall foul of the same rate limiting as it would be using the describe style endpoints/actions.

mario-galic:
Might be possible to consume directly in the app by combining ImportValue with environmental variables for the EC2 instance.

app/server/apiGatewayDiscovery.ts
```diff
@@ -213,3 +213,29 @@ export const getContactUsAPIHostAndKey = async () => {
...
+
+/**
+ * Used to fail riff raff deployment. Uses an arbitrarily selected service layer api as a check.
+ * Memoised in the sense of credentials fetching happens only once on deployment.
+ */
+export const memoisedApiGatewayAuthForHealthcheck = () => {
+  const memoisedHostKeyPair = getHostAndApiKeyForStack(
+    "membership",
+    "holiday-stop-api",
+    "PROD"
```
twrichards, line 225: you should be able to get the stage variable here from `config.STAGE`

PR #561, opened 2021-01-11
Help centre known issues

rBangay:
## What does this change?
Display known issues to the user at the top of the help centre page. Based off of config in s3.

## Images
Screenshot 2021-01-11 at 15 20 47


app/client/components/helpCentre/knownIssues.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useEffect(() => {
+    (async () => {
+      const knownIssuesResponse = await fetch("/api/known-issues/");
+      const unfilteredKnownIssues: Issue[] = knownIssuesResponse.ok
+        ? await knownIssuesResponse.json()
+        : [];
+      const unfilteredDateSortedIssues = unfilteredKnownIssues.sort(
+        (a, b) => Date.parse(a.date) - Date.parse(b.date)
+      );
+      const responseContainsProductIssues = unfilteredKnownIssues.some(
```
ripecosta, line 32: `unfilteredKnownIssues` will always be an extremely small array. For that reason you could remove `responseContainsProductIssues` and have a cleaner code. I believe the benefits of cleaner code outweigh the performance benefits of having this var.

ripecosta (changes requested):

app/client/components/helpCentre/knownIssues.tsx
rBangay: `responseContainsProductIssues` here is used to check if we have any product specific issues as opposed to the generic global issues.

If `responseContainsProductIssues` is true then an additional api call to members-data-api is necessary in order to match the product specific issues up with the products that the user has.

`responseContainsProductIssues` is therefore a gate between making an unnecessary api call if there are only global issues.

ripecosta (approved):
Looks good!

PR #562, opened 2021-01-11
Better wording for the form submission architecture decision

ripecosta:
## What does this change?
Changes the wording of the form submission architecture decision following @kelvin-chappell 's suggestions on #508 (which I only recently noticed)


kelvin-chappell (approved):
A huge improvement!

PR #563, opened 2021-01-12
[RB] - re-escalate existing newspaper delivery problems

rBangay:
## What does this change?
Over the Christmas period we allowed users to go through the delivery reporting journey without escalation to csr's.

This pr revert that change so that existing delivery problems get escalated as before.


ripecosta (approved):
👍

PR #565, opened 2021-01-13
Remove react server rendering

ripecosta:
## What does this change?
This PR removes the React server-side rendering. It still server-side renders the HTML that React renders now so it's easier to introduce the extra information we want on it (such as global variables, scripts, etc.)

## Why are we doing this?
We are working on code splitting the codebase as a way to alleviate the large bundle size it currently has. Code splitting becomes quite complex if you couple it with server-side rendering. However manage-frontend does not get have much need of server-side rendering at this point, so we are trialing removing it to more easily introduce code splitting.

Removing the server-side rendering from the project aggravates the problem of having a large bundle size, as it takes longer for users to see content. For that reason **this PR is not to be merged until the code splitting PR is ready** and we are confident that our approach adequately addresses all the performance concerns. In the meantime this PR will remain open to gather feedback.

### DIFF
--- a/app/client/components/main.tsx
+++ b/app/client/components/main.tsx
@@ -4,11 +4,7 @@ import { serif } from "../styles/fonts";
 import { Footer } from "./footer/footer";
 import Header from "./header";
 
-export interface WithOptionalServerPathWithQueryParams {
-  serverPathWithQueryParams?: string;
-}
-
-interface MainProps extends WithOptionalServerPathWithQueryParams {
+interface MainProps {
   children: any;
 }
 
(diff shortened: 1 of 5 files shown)


twrichards (commented):
Looks good, nice work everyone.

- [ ] There's loads of snippets all over the place, which check for `window` undefined etc which I think are only there for SSR reasons (since there is no `window` on the server) and could be removed.

- [ ] Also means you can delete all the references from the wiki once this is merged https://github.com/guardian/manage-frontend/search?q=server-side+rendering&type=wikis

Also for posterity, what I said on chat...

> on the face of it this seems sensible, I long considered removing SSR (less for code splitting purposes, although great that you're doing that 🙌) given it does very little (since we do so much ajax), adds complexity and is the cause of various bugs from time to time (dare to mention the occasions where the  entire page has ended up rotating around the loading spinner, see https://github.com/guardian/manage-frontend/commit/5cb8c7a84399185339a057d0961b09b5123a3d1e#diff-80c45e47af81baf4ac77ffcff3aed2ab and https://github.com/guardian/manage-frontend/pull/427)

arausuy (approved):

PR #567, opened 2021-01-15
Update "A payment issue" cancellation flow

TomPretty:
## What does this change?
Update "A payment issue" cancellation flow, as per [this card](https://trello.com/c/BbWsdbz6/2491-mma-cancellation-a-payment-issue-flow). There's some duplication here, but it didn't seem worth the effort to fight to get the current feedback form working for me so I created a new one (which will also use on the "I would rather do a [single/monthly/annual] contribution]" flow.

### review + <= min amount

Screenshot 2021-01-15 at 10 52 47

### review + > min amount
Screenshot 2021-01-15 at 10 54 12

### confirmed (saved)
Screenshot 2021-01-15 at 10 54 28


tomrf1 (approved):
looks great!

PR #568, opened 2021-01-15
Change Fastly to by default not cache unless route is explicitly added to allow-list

mario-galic:
## What does this change?

_**Note this PR is just for audit purposes as we currently do not version control VCL, that is, the change was done directly via Fastly UI.**_

There exists a risk of serving personal data to wrong user due to having Fastly caching in-front of manage-frontend. This PR strengthens the safeguard by changing from deny-list principle to allow-list principle in Fastly, that is, for a route to become cacheable from now own it will have to be explicitly allowed. Previously, for example, adding a new route under say `/foo/bar` and accidentally placing this route above middleware that disables caching in `server.ts` might have resulted in caching of personal data.

VCL Before

```
 if ( req.url ~ "^/api/" ) {
      # Force PASS on all /api/* endpoints
      set beresp.ttl = 0s;
      set beresp.grace = 0s;
      return(pass);
  }
``` 

VCL After

```
 if ( req.url !~ "^/(static)" ) {
      # do not cache
      set beresp.ttl = 0s;
      set beresp.grace = 0s;
      return(pass);
  }
```

image

From now one, to make the route `newroute` cachable 

1. Modify VCL condition to `req.url !~ "^/(static|newroute)"`
2. Add the route above `disableCache` middleware in `server.ts` or override `cache-control` headers 

[Fastly-&-Caching](https://github.com/guardian/manage-frontend/wiki/Fastly-&-Caching) corresponding wiki has been updated. (TODO)

## How to test

The change was tested in CODE by observing `x-cache` headers

image

and Fastly stats dashboards 

image

to see if `/static` routes are still cached whilst routes with personal data such as `/api/invoices` are not (`x-cache: MISS`)

`x-cache: MISS` on all endpoints returning personal data.

* Changing default from allow-all to deny-all should strengthen the safety net in Fastly.
* There is a question of the value of having Fastly in place at all for something like manage-frontend since the purpose of the app is to manage personal accounts.


ripecosta (approved):
👍  Looks good. Thanks Mario.

Regarding your question about the value of using Fastly in MMA:
We currently use Fastly to provide us the geolocation code we need for the Consent Management Platform. In the near future we will need to cache the server-side rendered Help Centre Article pages.

twrichards (approved):
Great work and excellent write-up - nice one!

paulbrown1982:
👍  Good spot and good question to raise about whether Fastly is necessary. It might be worth doing a deep dive into what performance / security protection it really is giving us (e.g. TCP slow-start / DDoS), and whether features like a web application firewall is what an application like this really ought to have.

- https://docs.fastly.com/products/web-application-firewall
- https://docs.aws.amazon.com/waf/latest/developerguide/cloudfront-features.html

PR #569, opened 2021-01-15
Redirect straight to update payment method page

TomPretty:
## What does this change?
Redirect the user directly to the update payment method page, instead of to the account overview page.


tomrf1 (dismissed):
👍

tomrf1 (approved):

PR #571, opened 2021-01-19
Don't show e.g "I would rather make an annual contribution" cancellation reason to annual contributors 

TomPretty:
## What does this change?
The two user facing changes are:
- Don't show "I would rather make an annual contribution" to annual contributors
- Don't show "I would rather make an monthly contribution" to monthly contributors 

To do this, I've added an optional (seemed less painfull to make it optional) `shouldShow` function to the `cancellationReason` model. I'm not super familiar with all the data types, but `ProductDetail` seemed like a reasonable choice for the argument here. It worked for this particular use case and I think it would work for potential future ones? (Very much open to suggestions here though!)
 
### annual contributor - "I would rather make an annual contribution" is hidden
Screenshot 2021-01-19 at 09 16 32

### DIFF
--- a/app/client/components/cancel/cancellationFlow.tsx
+++ b/app/client/components/cancel/cancellationFlow.tsx
@@ -21,7 +21,6 @@ import { WithStandardTopMargin } from "../page";
 import { ProgressIndicator } from "../progressIndicator";
 import { RadioButton } from "../radioButton";
 import {
-  MultiRouteableProps,
   ReturnToAccountOverviewButton,
   RouteableStepProps,
   WizardStep
@@ -36,7 +35,9 @@ import {
   cancellationDateFetcher,
   CancellationDateResponse
 } from "./cancellationDateResponse";
+import { CancellationReason } from "./cancellationReason";
 import { ContactUsToCancel } from "./contactUsToCancel";
+import { GenericSaveAttemptProps } from "./stages/genericSaveAttempt";
 
 export interface RouteableStepPropsWithCancellationFlow
   extends RouteableStepProps {
@@ -52,6 +53,9 @@ interface ReasonPickerState {
   reasonPath: string;
   cancellationPolicy?: string;
 }
+
+const shouldShow = (reason: CancellationReason, productDetail: ProductDetail) =>
+  reason.shouldShow ? reason.shouldShow(productDetail) : true;
 
 class ReasonPicker extends React.Component<
   ReasonPickerProps,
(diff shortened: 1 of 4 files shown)


app/client/components/cancel/cancellationFlow.tsx
```diff
@@ -53,6 +54,9 @@ interface ReasonPickerState {
   cancellationPolicy?: string;
 }
 
+const shouldShow = (reason: CancellationReason, productDetail: ProductDetail) =>
+  reason.shouldShow ? reason.shouldShow(productDetail) : true;
+
```
TomPretty, line 59: Seemed easiest to make `shouldShow` optional and use it like this, rather than add e.g

```js
shouldShow = () => true
```

to all the  cancellation reasons

app/client/components/cancel/cancellationFlow.tsx
```diff
@@ -75,7 +79,7 @@ class ReasonPicker extends React.Component<
 
     // we extract the options from the children rather than direct from the ProductType to guarantee the routes are setup correctly
     const options = this.props.children.props.children.map(
-      (child: { props: MultiRouteableProps }) => child.props
+      (child: { props: GenericSaveAttemptProps }) => child.props
```
TomPretty, line 82: This line seems slightly wild to me, but the children are `GenericSaveAttempt`s so this change is just so Typescript lets me access the `reason` prop.

app/client/components/cancel/contributions/contributionsCancellationReasons.tsx
```diff
@@ -100,7 +104,14 @@ export const contributionsCancellationReasons: CancellationReason[] = [
         </p>
       </>
     ),
-    alternateFeedbackIntro
+    alternateFeedbackIntro,
+    shouldShow: (productDetail: ProductDetail) => {
+      const mainPlan = getMainPlan(productDetail.subscription);
+      if (!isPaidSubscriptionPlan(mainPlan)) {
+        return false;
+      }
```
TomPretty, line 112: `isPaidSubscriptionPlan` should always be `true`, as this option is only rendered for a recurring contribution, but we need this for the type guard

app/client/components/cancel/contributions/contributionsCancellationReasons.tsx
```diff
@@ -114,7 +125,14 @@ export const contributionsCancellationReasons: CancellationReason[] = [
...
```
TomPretty, line 133: same as above

tomrf1 (approved):
makes sense 👍

PR #572, opened 2021-01-19
[RB] - pass credentials explicitly through with fetch calls for older versions of safari

rBangay:
## What does this change?
Older versions of Safari (versions 11 and 12) need the `credentials` and `mode` properties to be passed through with fetch calls.

## Images
Screenshot_2021-01-15_at_13 35 53

### DIFF
--- a/app/client/components/delivery/records/deliveryRecordsApi.ts
+++ b/app/client/components/delivery/records/deliveryRecordsApi.ts
@@ -85,6 +85,8 @@ export const createDeliveryRecordsFetcher = (
   isTestUser: boolean
 ) => () =>
   fetch(`/api/delivery-records/${subscriptionId}`, {
+    credentials: "include",
+    mode: "same-origin",
     headers: {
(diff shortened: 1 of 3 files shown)


ripecosta (approved):
SQUASHED!

paulbrown1982:
Good catch!

app/client/components/delivery/records/deliveryRecordsApi.ts
```diff
@@ -85,6 +85,8 @@ export const createDeliveryRecordsFetcher = (
   isTestUser: boolean
 ) => () =>
   fetch(`/api/delivery-records/${subscriptionId}`, {
+    credentials: "include",
+    mode: "same-origin",
     headers: {
       [MDA_TEST_USER_HEADER]: `${isTestUser}`
     }
```
twrichards, line 92: notice these headers, they might well need adding to the other fetches, so they 'test-users' works correctly, see https://github.com/guardian/manage-frontend/wiki/test-users#test-users

PR #574, opened 2021-01-21
Introduce code splitting

marialani:
## What does this change?
This PR introduces route based code splitting using React Lazy and Suspense. A Suspense fallback component has also been added which acts as an intermediary view whilst each separate code chunk is loading. React Lazy works with files that have a default export so the relevant pages have been updated.

The tsconfig `module` value has been changed from `commonjs` to `ESnext` in order to get the code splitting to work as expected, with the previous setting webpack had trouble separating chunks. As a result of changing the tsconfig the jest configuration has also been updated to ensure it uses commonjs.

- the parsed bundle size was 1.32mb and is now 1.06mb
- the gzipped bundle size was 362kb and is now 303.19kb

There was a risk of users seeing a blank whilst the page was loading. We have addressed this by including an intermediary view (pageSkeleton) whilst chunks are loading.

## Images
This is what it looks like on a fast 3G connection

Successfully tested on CODE

Co-authored by @ripecosta @rBangay

### DIFF
--- a/app/client/components/pageSkeleton.tsx
+++ b/app/client/components/pageSkeleton.tsx
@@ -1,0 +1,177 @@
+import { css } from "@emotion/core";
+import { space } from "@guardian/src-foundations";
+import { Location } from "@reach/router";
+import React from "react";
+import {
+  GROUPED_PRODUCT_TYPES,
+  PRODUCT_TYPES,
+  shouldHaveHolidayStopsFlow
+} from "../../shared/productTypes";
+import { MenuSpecificNavItem, NAV_LINKS } from "./nav/navConfig";
+import { PageContainer, WithStandardTopMargin } from "./page";
+import { SectionHeader } from "./sectionHeader";
+import { SectionPageContainer } from "./sectionPageContainer";
+import { Spinner } from "./spinner";
+
+interface LocationObject {
+  title: string;
+  path: string;
+  selectedNavItem: MenuSpecificNavItem;
+}
+
+const manageProductLocationObjects: LocationObject[] = Object.values(
+  GROUPED_PRODUCT_TYPES
+).map(groupedProductType => ({
+  title: `Manage ${groupedProductType.shortFriendlyName ||
+    groupedProductType.friendlyName}`,
+  path: `/${groupedProductType.urlPart}`,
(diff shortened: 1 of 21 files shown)


app/client/components/user.tsx
```diff
@@ -15,181 +15,197 @@ import {
...
+const Settings = lazy(() => import("./identity/Settings"));
+const Help = lazy(() => import("./help"));
+const HelpCentre = lazy(() => import("./helpCentre/helpCentre"));
+const ContactUs = lazy(() => import("./contactUs/contactUs"));
+
 const User = () => (
   <Main>
     <Global styles={css(`${global}`)} />
     <Global styles={css(`${fonts}`)} />
+    {/* replace with a real fallback component */}
```
rBangay, line 62: we can probably remove this comment

app/client/components/pageSkeleton.tsx
```diff
@@ -0,0 +1,177 @@
...
+        )[0];
+
+        if (selectedNonMMALocationObject) {
+          return (
+            <>
+              <SectionHeader title={selectedNonMMALocationObject.title} />
+              <SectionPageContainer sectionTitle="&nbsp;">
+                <div
+                  css={css`
+                    margin-bottom: ${space[24]}px;
```
arausuy, line 147: How safe is this access by index?

app/client/components/user.tsx
arausuy: 👍

app/client/components/pageSkeleton.tsx
marialani: I believe there will be some refactoring done here once there is a bit more standardisation across the codebase. This works for now with what we have, on safety I will defer to @rBangay and @ripecosta.

arausuy (approved):

app/client/components/pageSkeleton.tsx
ripecosta: Very safe. It is made available by the Guardian Design System (Source components). Documentation for `space` [here](https://www.theguardian.design/2a1e5182b/p/449bd5-space)

ripecosta (approved):
Re-Approving, following rebase!

PR #575, opened 2021-01-25
Refactor deliveryRecords.tsx to use our FormError component

marialani:
## What does this change?
The FormError component replaces the existing error message code in deliveryRecords with relevant content passed down in the form of props. This has been done to further simplify and standardise our codebase.

### DIFF
--- a/app/client/components/delivery/records/deliveryRecords.tsx
+++ b/app/client/components/delivery/records/deliveryRecords.tsx
@@ -269,19 +269,16 @@ export const DeliveryRecordsFC = (props: DeliveryRecordsFCProps) => {
       : "Some information is missing";
 
   const formErrorMessages = [
-    ...(!step1FormValidationDetails.isValid &&
-    step1FormValidationDetails.message
-      ? [step1FormValidationDetails.message]
-      : []),
-    ...(!step2FormValidationDetails.isValid &&
-    step2FormValidationDetails.message
-      ? [step2FormValidationDetails.message]
-      : []),
-    ...(!step3FormValidationDetails.isValid &&
-    step3FormValidationDetails.message
-      ? [step3FormValidationDetails.message]
-      : [])
-  ];
+    step1FormValidationDetails,
+    step2FormValidationDetails,
+    step3FormValidationDetails
+  ].reduce(
+    (acc: string[], validationDetails) =>
+      !validationDetails.isValid && validationDetails.message
+        ? [...acc, validationDetails.message]
+        : acc,
+    []
+  );
 
   return (
     <DeliveryRecordsProblemContext.Provider
--- a/app/client/components/delivery/records/deliveryRecords.tsx
+++ b/app/client/components/delivery/records/deliveryRecords.tsx
@@ -624,12 +624,7 @@ export const DeliveryRecordsFC = (props: DeliveryRecordsFCProps) => {
                 {(step1formValidationState ||
                   step2formValidationState ||
                   step3formValidationState) &&
-                  ((!step1FormValidationDetails.isValid &&
-                    step1FormValidationDetails.message) ||
-                    (!step2FormValidationDetails.isValid &&
-                      step2FormValidationDetails.message) ||
-                    (!step3FormValidationDetails.isValid &&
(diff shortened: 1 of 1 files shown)


app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -260,6 +260,28 @@ export const DeliveryRecordsFC = (props: DeliveryRecordsFCProps) => {
...
+      : []),
+    ...(!step2FormValidationDetails.isValid &&
+    step2FormValidationDetails.message
+      ? [step2FormValidationDetails.message]
+      : []),
+    ...(!step3FormValidationDetails.isValid &&
+    step3FormValidationDetails.message
+      ? [step3FormValidationDetails.message]
+      : [])
+  ];
```
ripecosta, line 283: Would using `formErrorMessages.push()` help simplify this code?

ripecosta (changes requested):
Can we also extract the condition in lines 626-634 from the return so the component is easier to read?

app/client/components/delivery/records/deliveryRecords.tsx
marialani: As the array contains 3 elements with repeated logic this method seems feasible. The alternative of a chain of if statements with `formErrorMessages.push()` would make the code more verbose. Though, this method does require JS knowledge on how the spread operator works on an array within an array.

twrichards: this looks ripe for a function which takes in `validationDetails` does the ternary fun, which you could either have as a named function, which you call three times, or better IMO would be something like...
```suggestion
  const formErrorMessages = [
       step1FormValidationDetails,
       step2FormValidationDetails,
       step3FormValidationDetails
    ].filter(validationDetails => 
        !validationDetails.isValid && validationDetails.message
    );
```

marialani: Ooh that looks great! 👍  
As I'm new here, may I commit your suggestion here? Or do I simply praise it whilst making another commit locally and pushing it to this branch

twrichards: Hi Maria, and welcome!

Good question, in my experience for small typo style suggestions, committing directly from GitHub suggestion (using the `Commit suggestion`). However for something like this I would probably reimplement the change yourself to test it works as expected and to ensure prettier does its thing etc.

Worth saying though, its totally up to you there is no rule (or even convention) and this is your PR (that you could rebase if you committed a suggestion you later wished you hadn't.

Only thing worth nothing, is that committing suggestions adds the suggester as a[ 'Co-Author'](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) which you might want to bear in mind when using this in future (I'm not mentioning because I want credit for this suggestion, for my sins my name is all over this repo already 🙈 )

marialani: FormErrorMessages has been refactored

marialani:
> Can we also extract the condition in lines 626-634 from the return so the component is easier to read?

Shall we separate that refactoring task into another PR?

twrichards:
> > Can we also extract the condition in lines 626-634 from the return so the component is easier to read?
> 
> Shall we separate that refactoring task into another PR?

Now you've done that lovely `reduce`, could the condition that @ripecosta mention be simplified to display `FormError` if `formErrorMessages.length > 0', since the logic that makes items appear in that list is virtually identical to what the reduce is now doing.

ripecosta:
> > Can we also extract the condition in lines 626-634 from the return so the component is easier to read?
> 
> Shall we separate that refactoring task into another PR?

@marialani That bit of a code still pertains to the formError so it still makes sense to do it as part of this refactor PR. @twrichards 's suggestion of replacing the condition with `formErrorMessages.length > 0` is a very good one.

ripecosta (approved):
LGTM 👍

twrichards:
Awesome work @marialani - well done for sticking with it. Love a PR that is a net reduction in CODE... 
image
👏

PR #576, opened 2021-01-25
[RB] - import specific colours from pallete instead of everything

rBangay:
## What does this change?
Import only the specific colours that we use from the palette package instead of all of them

Should be a pretty small decrease in our client bundle size

## Images
nothing visual should change with this pr


ripecosta (approved):
I didn't know we could do that! Pretty cool! 👍 
Did you notice a change in the bundle size as a consequence?

rBangay:
> I didn't know we could do that! Pretty cool! 👍
> Did you notice a change in the bundle size as a consequence?

tiny tiny:

Before             |  After
  |

PR #579, opened 2021-01-29
Add functionality for links in Known Issues banner

marialani:
## What does this change?
This PR adds an optional link to the an issue alert. This means we can now provide further information to the user about the issue should we choose to.

## Images
Before

After


ripecosta (approved):
Great work! 👍

PR #580, opened 2021-02-01
Split MMA and HelpCentre bundles

ripecosta:
## What does this change?
This PR splits our single bundle (user.js) into an MMA bundle and Help Centre bundle to improve bundle size. MMA's bundle size has a minimal reduction but the Help Centre bundle shrinks significantly. This will provide a better experience for users accessing the Help Centre and the Contact Us form.

It also adds [magic comments](https://webpack.js.org/api/module-methods/#magic-comments) to the dynamic imports that instruct Webpack on how to name the chunk produced byt that import. This will make it easier for us to debug any issues that involve the chunks.

Note: More can be done to further decrease the bundle sizes, especially for the Help Centre. As it stands there is still unnecessary code being imported into each bundle because of how our code e organised.I will raise a separate PR with these changes and both can be merged to master at the same time.

### Bundle analysis

Before | After
------------ | -------------
Screenshot 2021-02-01 at 18 14 35 | Screenshot 2021-02-01 at 18 17 47

#### Summary

&nbsp; | parsed | gziped
------------ | ------------- | -------------
Single bundle | 1.06 MB | 302.85 KB
MMA bundle | 1.06 MB | 302.04 KB
HelpCentre bundle | 460.13 KB | **139.89 KB**

### Browser analysis
How much code do we actually need to load before the user can interact with the page?
<sub>This includes the main bundle plus any chunks needed to render each of the landing page</sub>

&nbsp; | Before | After | % Change
------------ | ------------- | ------------- | -------------
MMA | 331.2 KB | 331.3 KB | +0.03%
Help Centre | 323.4 KB | **207.3 KB** | **-35.9%**
Contact Us | 328.1 KB | **171.2 KB** | **-47.82%**

### DIFF
--- a/app/client/components/HelpCentrePage.tsx
+++ b/app/client/components/HelpCentrePage.tsx
@@ -9,6 +9,10 @@ import HelpCentrePageSkeleton from "./HelpCentrePageSkeleton";
 import { Main } from "./main";
 import { ScrollToTop } from "./scrollToTop";
 
+// The code below uses magic comments to instruct Webpack on
+// how to name the chunks these dynamic imports produce
+// More information: https://webpack.js.org/api/module-methods/#magic-comments
+
 const HelpCentre = lazy(() =>
   import(/* webpackChunkName: "HelpCentre" */ "./helpCentre/helpCentre")
 );
--- a/app/client/components/HelpCentrePage.tsx
+++ b/app/client/components/HelpCentrePage.tsx
@@ -1,0 +1,49 @@
+import { css, Global } from "@emotion/core";
+import { Redirect, Router } from "@reach/router";
+import React, { lazy, Suspense } from "react";
+import { fonts } from "../styles/fonts";
+import global from "../styles/global";
+import { AnalyticsTracker } from "./analytics";
+import { CMPBanner } from "./consent/CMPBanner";
+import { Main } from "./main";
+import PageSkeleton from "./pageSkeleton";
+import { ScrollToTop } from "./scrollToTop";
+
+const HelpCentre = lazy(() =>
+  import(/* webpackChunkName: "HelpCentre" */ "./helpCentre/helpCentre")
+);
+const ContactUs = lazy(() =>
+  import(/* webpackChunkName: "ContactUs" */ "./contactUs/contactUs")
+);
+
+const HelpCentreRouter = () => {
+  return (
+    <Main>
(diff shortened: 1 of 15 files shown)


app/client/components/HelpCentrePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { fonts } from "../styles/fonts";
+import global from "../styles/global";
+import { AnalyticsTracker } from "./analytics";
+import { CMPBanner } from "./consent/CMPBanner";
+import HelpCentrePageSkeleton from "./HelpCentrePageSkeleton";
+import { Main } from "./main";
+import { ScrollToTop } from "./scrollToTop";
+
+const HelpCentre = lazy(() =>
+  import(/* webpackChunkName: "HelpCentre" */ "./helpCentre/helpCentre")
```
rBangay, line 13: can we write a sneaky comment somewhere explaining that the webpackChunkName comments here are not just comments but functional

👍

ripecosta: Yup, I will add those. Also added information about magic comments to the PR description.

ripecosta: Done!

app/client/components/HelpCentrePageSkeleton.tsx
```diff
@@ -0,0 +1,58 @@
...
+    title: "Help Centre",
+    path: "/help-centre"
+  },
+  {
+    title: "Need to contact us about something?",
+    path: "/contact-us"
+  }
+];
+
+const MMAPageSkeleton = () => {
```
rBangay, line 26: no biggie ... MMAPageSkeleton doesn't need to be a block level component

app/client/components/MMAPageSkeleton.tsx
```diff
@@ -113,49 +109,13 @@ const MMALocationObjectArr: LocationObject[] = [
...
-  {
-    title: "Need to contact us about something?",
-    path: "/contact-us",
-    selectedNavItem: NAV_LINKS.help
-  }
-];
-
-const PageSkeleton = () => {
+const MMAPageSkeleton = () => {
   return (
```
rBangay, line 113: while we're at it, this also doesn't need to be a block level component

rBangay (approved):

PR #581, opened 2021-02-02
[RB] - load fonts from assets.guim

rBangay:
## What does this change?
Manage statically hosted it's fonts. The aim of this pr is to consolidate the loading of the fonts in line with dotcom (from https://assets.guim.co.uk/static/frontend/fonts/) and hopefully making use of caching so that if the user comes from the main site into manage they don't have to re-download the same fonts.

## Consequences
The removal of all of the statically hosted fonts from manage-frontend totaling `1.7mb`, however much of which should now be cached for users that have come from the main Guardian site into manage 🎉 .

..... and a whopping 2.83kb reduction in the parsed sized of our client bundle!!

Before             |  After
before  |  after

cashback


ripecosta (dismissed):
Looks good!

ripecosta (approved):
👍

PR #583, opened 2021-02-03
Remove the changeTier field from ProductType

rupertbates:
## What does this change?
We are no longer allowing members to change their tier via membership.theguardian.com so this PR removes the 'change tier' button from the manage product page. 

As membership was the product which had a change tier option I have also removed it from the ProductType type.

## How to test
Log into manage with an account that has a membership and click the 'manage membership' button, you should no longer see the 'change tier' button on the manage product page.

This is just a tidy up exercise


ripecosta (approved):
Looks good 👍

PR #586, opened 2021-02-09
Update help centre topics

marialani:
## What does this change?
This PR updates the featured topics on the Help Centre landing page and includes an FAQ section with further topics.


app/client/components/helpCentre/HelpTopicBox.tsx
```diff
@@ -57,7 +57,7 @@ const linkListItemStyle = css`
   position: relative;
 `;
 
-const linkAnchorStyle = css`
+export const linkAnchorStyle = css`
```
ripecosta, line 60: Because `HelpTopicBox` and `HelpCentreFaq` components don't depend on each other perhaps it's best to move these css declarations into a separate file that both components import. We could start a `helpCentreStyles` file where we put all the styles we need to share across components.

app/client/components/helpCentre/helpCentre.tsx
```diff
@@ -71,6 +72,8 @@ const HelpCentre = (_: HelpCentreProps) => (
             <HelpTopicBox key={topic.id} topic={topic} />
           ))}
         </div>
+        <h2 css={contactUsTitleStyle}>FAQ</h2>
```
ripecosta, line 75: `contactUsTitleStyle` is now being used for more than just the Contact Us title. Can we rename it so it has a more meaningful name?

app/client/components/helpCentre/helpCentreFaq.tsx
```diff
@@ -0,0 +1,159 @@
...
+        top: 0;
+        left: 0px;
+        width: 100%;
+        height: 1px;
+        background-color: ${neutral["86"]}
+      }
+    `}
+  `;
+
+  const innerSectionCss = (isOpen: boolean) => `
```
ripecosta, line 66: Is there a reason to declare these as functions instead of dynamic styles like we do elsewhere?

ripecosta (changes requested):

app/client/components/helpCentre/helpCentreFaq.tsx
```diff
@@ -0,0 +1,159 @@
...
+    padding: 0;
+    list-style: none;
+    background-color: rgba(193, 216, 252, 0.3);
+    border-top: 1px solid #DCDCDC;
+  `;
+
+  const innerSectionDivCss = `
+    ${textSans.medium()};
+    margin-bottom: 0;
+    padding: 12px 20px 12px 0;
```
rBangay, line 78: not a biggie, but 12px and 20px are both part of the source library, so this line could become:

`padding: ${space[3]}px ${space[5]}px ${space[3]}px 0`

app/client/components/helpCentre/helpCentreFaq.tsx
```diff
@@ -0,0 +1,159 @@
...
+    margin: 0 12px;
```
rBangay, line 79:
```suggestion
    margin: 0 ${space[3]}px;
```

app/client/components/helpCentre/HelpTopicBox.tsx
```diff
@@ -67,7 +67,7 @@ const linkAnchorStyle = css`
   }
 `;
 
-const linkArrowStyle = css`
+export const linkArrowStyle = css`
```
rBangay, line 70: If we're going to create a new `helpCentreStyles` file, then this would also be a good candidate for adding in to it :)

app/client/components/helpCentre/HelpTopicBox.tsx
marialani: Sure, I can start a separate file for shared Help Centre styles starting with these two variables 👍

app/client/components/helpCentre/helpCentreFaq.tsx
marialani: Is there an issue with declaring these as functions?

app/client/components/helpCentre/helpCentre.tsx
marialani: What title would you suggest?

app/client/components/helpCentre/helpCentreFaq.tsx
marialani: Will make this change 👍

app/client/components/helpCentre/helpCentre.tsx
ripecosta: Something like 'subtitleStyles' seems appropriate. What do you think?

app/client/components/helpCentre/helpCentreFaq.tsx
ripecosta: I was wondering if there's a benefit to this, as it's a change of pattern from what we do in project

marialani: Ah right, there are few examples of in the codebase using this method at the mo

ripecosta: We have inline styled components, yes. But not functions declared and then converted inline into dynamic styles.

marialani: Have a search in the codebase, you should see a few examples

rBangay (approved):

PR #587, opened 2021-02-09
Soft opt ins

rBangay:
## What does this change?
This pr is the manage-frontend part of the wider soft-opt in work. The scope of this pr is the loading of the new consents endpoint from identity api and displaying of relevant soft opt in's dependent on what products the user has. 

Users should be able to see which emails they are opted into and out of and be able to opt in and out of them.

**Until the other parts of the soft-opt in work in support-frontend (work to opt users in, on product acquisition) and support-service-lambdas (work to opt users out when they reach their cancellation effective date) are done, then the new opt ins should not be visible to the user. This is handled crudely in the code using the following boolean in** `client/components/identity/EmailAndMarketing/ConsentSection.tsx`

Screenshot 2021-02-09 at 16 04 39

## Images
example of the new opt ins (they will be hidden as part of this pr though)
Screenshot 2021-02-09 at 09 47 34

### DIFF
--- a/app/client/components/identity/EmailAndMarketing/ConsentSection.tsx
+++ b/app/client/components/identity/EmailAndMarketing/ConsentSection.tsx
@@ -11,6 +11,10 @@ interface ConsentSectionProps {
   consents: ConsentOption[];
 }
 
+const releaseSoftOptIns = false;
+const softOptInEmailConsents = (consents: ConsentOption[]): ConsentOption[] =>
+  consents.filter(consent => !!consent.isProduct && releaseSoftOptIns);
+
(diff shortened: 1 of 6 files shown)


app/client/components/identity/EmailAndMarketing/ConsentSection.tsx
```diff
@@ -11,6 +11,10 @@ interface ConsentSectionProps {
   consents: ConsentOption[];
 }
 
+const releaseSoftOptIns = false;
```
ripecosta, line 14: 10/10 feature flag right here 🤣

ripecosta (approved):
LGTM! 👍

marialani (approved):

PR #591, opened 2021-02-18
Help centre header

marialani:
## What does this change?
This PR includes a separate header for the help centre.

## Images
MMA header

Help centre header
image


app/client/components/nav/dropdownNav.tsx
```diff
@@ -110,6 +110,10 @@ export const DropdownNav = () => {
     };
   });
 
+  const isHelpCentre = window.location.pathname.startsWith("/help-centre")
```
paulbrown1982, line 113: You don't need the `? true : false` here as startsWith already returns a boolean.

marialani: Thanks Paul, I missed that and will update it now 👍

paulbrown1982 (approved):
👍  Looks great!

PR #593, opened 2021-02-19
Send country code in create reminder request

TomPretty:
## What does this change?
Send country code in create reminder request. The endpoint was initially getting this data from a header set by fastly. However, becaue this request is proxied via the manage backend, the location is the location of the AWS instance that serves that request, not the user. Now we send the country too which the reminder endpoint will use over the fastly header


tomrf1 (approved):
👍

PR #594, opened 2021-02-22
add instance name to ASG

johnduffell:
this makes it easier to understand the AWS console, also for reserved instances/trusted advisor
image


ripecosta (approved):
👍

PR #596, opened 2021-02-24
Remove snyk from the build steps

ripecosta:
## What does this change?
This PR removes snyk from the build step and moves it into a GitHub action

`snyk monitor` will run on any push to main. `snyk test` will run on all other branches. This will make it so our builds aren't blocked in case a vulnerability is found, which means we won't have to ignore them in the snyk policy. The snyk action is not required to pass for a PR to be merged. It's merely informative.

In a separate PR I will reset snyk's policy settings to remove the ignores we've been adding, now that it won't block the builds.

Note: Running `snyk wizard` will prompt to add a `snyk test` step to our `yarn test` script. This will no longer be necessary.

### DIFF
--- a/.github/workflows/snyk.monitor.yml
+++ b/.github/workflows/snyk.monitor.yml
@@ -8,9 +8,9 @@ jobs:
   monitor:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
       - name: Run Snyk to check for vulnerabilities
-        uses: snyk/actions/node@master
+        uses: snyk/actions/node@0.3.0
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
--- a/.github/workflows/snyk.monitor.yml
+++ b/.github/workflows/snyk.monitor.yml
@@ -14,4 +14,5 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
+          args: --org=the-guardian-cuu --project-name=guardian/manage-frontend
           command: monitor
--- a/.github/workflows/snyk.monitor.yml
+++ b/.github/workflows/snyk.monitor.yml
@@ -1,0 +1,17 @@
+# Snyk Vulnerability Check on main branches will report results to snyk.io
+name: Snyk Vulnerability Check
+on:
+  push:
+    branches:
+      - main
+jobs:
+  security:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@master
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/snyk.monitor.yml
```diff
@@ -0,0 +1,18 @@
...
+name: Snyk
+on:
+  push:
+    branches:
+      - main
+jobs:
+  monitor:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
kelvin-chappell, line 11: Might be better to rely on a hardcoded version rather than latest commit

.github/workflows/snyk.monitor.yml
```diff
@@ -0,0 +1,18 @@
...
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@master
```
kelvin-chappell, line 13: Might be better to rely on a hardcoded version rather than latest commit

.github/workflows/snyk.test.yml
```diff
@@ -0,0 +1,17 @@
...
+name: Snyk
+on:
+  push:
+    branches-ignore:
+      - main
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
kelvin-chappell, line 11: Might be better to rely on a hardcoded version rather than latest commit

.github/workflows/snyk.test.yml
```diff
@@ -0,0 +1,17 @@
...
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@master
```
kelvin-chappell, line 13: Might be better to rely on a hardcoded version rather than latest commit

kelvin-chappell (commented):
Good idea!

.github/workflows/snyk.monitor.yml
```diff
@@ -0,0 +1,18 @@
...
+  push:
+    branches:
+      - main
+jobs:
+  monitor:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@v0.3
```
kelvin-chappell, line 13:
```suggestion
        uses: snyk/actions/node@v0.3.0
```

.github/workflows/snyk.test.yml
```diff
@@ -0,0 +1,17 @@
...
+  push:
+    branches-ignore:
+      - main
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@v0.3
```
kelvin-chappell, line 13:
```suggestion
        uses: snyk/actions/node@v0.3.0
```

kelvin-chappell (approved):
👍

PR #602, opened 2021-02-26
Send ophan event on reminder view

TomPretty:
## What does this change?
Send an ophan event when the set reminder component is viewed. This is to allow us to more easily calculate a 'conversion rate' for one of our data studio dashboards.

## Images
Screenshot 2021-02-26 at 08 44 43


tomrf1 (approved):
👍

PR #604, opened 2021-02-26
Add support reminder consent to the email prefs page

TomPretty:
## What does this change?
Add a support reminder consent to the email reminders page. Removing this consent will cancel *all* pending reminders (both recurring and one-offs). Re-consenting will reactivate *only* their recurring reminder. A user that has never set a recurring reminder will not see this consent. 

**DO NOT MERGE** until after: https://github.com/guardian/members-data-api/pull/508 (this updates endpoint in mdapi)

## Images
Screenshot 2021-02-26 at 13 40 25

### DIFF
--- a/app/client/components/cancel/cancellationContributionReminder.tsx
+++ b/app/client/components/cancel/cancellationContributionReminder.tsx
@@ -34,7 +34,7 @@ interface ReminderChoice {
   thankYouMessage: string;
 }
 
-const REMINDER_ENDPOINT = "/api/reminder";
+const REMINDER_ENDPOINT = "/api/reminders";
 const REMINDER_PLATFORM = "MMA";
 const REMINDER_STAGE = "WINBACK";
 const REMINDER_COMPONENT = "CANCELLATION";
--- a/app/client/components/identity/idapi/supportReminders.ts
+++ b/app/client/components/identity/idapi/supportReminders.ts
@@ -1,18 +1,61 @@
+import * as Sentry from "@sentry/browser";
 import { ConsentOption, ConsentOptionType } from "../models";
 
-const supportReminder: ConsentOption = {
+interface ReminderStatusApiResponse {
+  recurringStatus: "NotSet" | "Active" | "Cancelled";
+}
+
+const REMINDERS_STATUS_ENDPOINT = "/api/reminders/status";
+const CANCEL_REMINDERS_ENDPOINT = "/api/reminders/cancel";
+const REACTIVATE_REMINDERS_ENDPOINT = "/api/reminders/reactivate";
+
+const getConsent = (isActive: boolean): ConsentOption => ({
   id: "support_reminder",
   description:
     "We will invite you to make a contribution in support of Guardian journalism, using the cadence you picked when you last signed up.",
   name: "Contribution reminder emails",
   type: ConsentOptionType.SUPPORT_REMINDER,
-  subscribed: true
-};
+  subscribed: isActive
+});
 
 export const read = async (): Promise<ConsentOption[]> => {
-  return [supportReminder];
+  const response = await fetch(REMINDERS_STATUS_ENDPOINT);
+  const reminderStatus = (await response.json()) as ReminderStatusApiResponse;
+  if (reminderStatus.recurringStatus === "NotSet") {
+    return [];
+  }
+
+  return [getConsent(reminderStatus.recurringStatus === "Active")];
 };
 
 export const update = async (id: string, subscribed: boolean = true) => {
-  console.log(id, subscribed);
+  const email = window.guardian.identityDetails.email;
+  if (!email) {
+    Sentry.captureMessage(`No email address to update consent: ${id}`);
+    return;
+  }
+
+  if (!subscribed) {
+    await cancelReminder(email);
+  } else {
+    await reactivateReminder(email);
+  }
 };
+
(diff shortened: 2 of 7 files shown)


app/server/routes/api.ts
```diff
@@ -227,6 +231,12 @@ router.get("/known-issues", async (_, response) => {
...
 router.post("/contact-us", contactUsFormHandler);
 
-router.post("/reminder", reminderHandler);
+router.post("/reminders", createReminderHandler);
+router.get(
+  "/reminders/status",
+  membersDataApiHandler("user-attributes/me/reminders", "MDA_REMINDERS_STATUS")
+);
+router.post("/reminders/cancel", cancelReminderHandler);
+router.post("/reminders/reactivate", reactivateReminderHandler);
```
TomPretty, line 240: might be worth factoring this out into its own router? and then

```js
router.use("/reminders", remindersRouter);
```

app/client/components/cancel/cancellationContributionReminder.tsx
```diff
@@ -34,7 +34,7 @@ interface ReminderChoice {
   thankYouMessage: string;
 }
 
-const REMINDER_ENDPOINT = "/api/reminder";
+const REMINDER_ENDPOINT = "/api/reminders";
```
TomPretty, line 37: Just for consistency with the other endpoints

app/server/routes/api.ts
ripecosta: In this case there is no need. We already use this pattern in the file. All the api endpoints are declared here but their logic is elsewhere. That way we don't have to go several levels deep to know which endpoints we are using.

app/server/routes/api.ts
```diff
@@ -227,6 +231,12 @@ router.get("/known-issues", async (_, response) => {
 
 router.post("/contact-us", contactUsFormHandler);
 
-router.post("/reminder", reminderHandler);
+router.post("/reminders", createReminderHandler);
+router.get(
+  "/reminders/status",
+  membersDataApiHandler("user-attributes/me/reminders", "MDA_REMINDERS_STATUS")
```
ripecosta, line 237: Why does this endpoint get routed through `membersDataApiHandler` but not the other two?
Using `membersDataApiHandler` will give you loads for free, like credentials and logging .

TomPretty: Yeah this is slightly confusing. This is the only request that actually goes to members data api - the others go direct to our reminders endpoint. @tomrf1  I can't remember now why we decided this one had to go through mdapi?

tomrf1: I might be missing something here, but this is the only new request to MDAPI so it's the only one that should be using `membersDataApiHandler`.

TomPretty: Yeah that's right. I was meaning more architecturally, why we decided this endpoint should be implemented in MDAPI?

tomrf1: Ah right.
This request reveals some information about a user (have they signed up?), so it needs to be auth'd, and MDAPI is the right place for that.
The cancel/reactivate endpoints do not tell the client anything, so no auth required

TomPretty: Gotcha! Makes sense 👍

app/client/components/identity/idapi/supportReminders.ts
```diff
@@ -0,0 +1,65 @@
+import * as Sentry from "@sentry/browser";
+import { ConsentOption, ConsentOptionType } from "../models";
+
+interface ReminderStatusApiResponse {
+  recurringStatus: "NotSet" | "Active" | "Cancelled";
+  recurringReminderCode: string;
```
tomrf1, line 6: `recurringReminderCode?: string;`

app/server/routes/api.ts
ripecosta: The cancel/reactivate endpoints are revealing anything about the user but they still make changes to a user's settings. For this reason I believe they should also require authentication.
Otherwise these endpoints will allow people to change each other's reminder settings. Are we ok with that risk?

There might be perfectly valid reasons for doing it this way that I'm just not aware of though. In which case I would like to see them documented in this PR so we can better understand our choices in the future.

tomrf1: These endpoints now take a `reminderCode` rather than the email address. This code is given to braze by our reminders system. It's not authentication, but it's good enough for this purpose

TomPretty: Yeah just to expand on this - `reminderCode` is a `uuid` that we generate at the point of signup. The only way to get access to this code is through the MDAPI call, or the email that we send to a user (it'll be in the query string of the unsub link). Like Tom said, it's not really an authentication system - but it does prevent altering other users reminders. Do you think that is secure enough?

tomrf1 (approved):
this is great, just one comment about the type of `recurringReminderCode`

app/client/components/identity/idapi/supportReminders.ts
TomPretty: oops I never pushed that commit!

tomrf1 (approved):

app/server/routes/api.ts
ripecosta: Sorry for the delay in replying. Yeah, I think that's a great improvement 👍

PR #609, opened 2021-03-04
[RB] - update cloudformation instance update to include prod

rBangay:
## What does this change?
This pr is to change the EC2 instance types from t3.small to t4g.small and also to update the instance OS from Ubuntu Xenial to Ubuntu Bionic (ARM).

A new recipe has been created in amigo `manage-frontend-arm` which is a clone of the original `manage-frontend` with the update to the OS and subsequent ARM version update to the node binary. Once this pr has been merged and monitored the original recipe can be deleted.

## How to test
Cloudwatch logs should be monitored for any errors. Also worth checking the CPU usage of the instances to make sure it is not too high. (has already been checked in CODE, but with hardly any traffic)

### DIFF
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -56,7 +56,7 @@ Mappings:
     PROD:
       MaxInstances: 6
       MinInstances: 3
-      InstanceType: t3.small
+      InstanceType: t4g.small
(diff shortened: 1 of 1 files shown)


ripecosta (approved):
👍

PR #610, opened 2021-03-04
Remove .vscode folder(s) from git

TomPretty:
## What does this change?
Remove the .vscode folder from git. I think you can make an argument for having a handcrafted settings/launch configs/tasks that are shared by the entire team, but given the variety of editors people use, I think it makes sense to remove them from version control and let individual devs hack their own setup. Potentially something that could be added to the docs? "Here's a good `.vscode` to get started with"....


.gitignore
```diff
@@ -11,7 +11,7 @@ coverage/
 
 *.code-workspace
 
-app/.vscode/
+.vscode/
```
TomPretty, line 14: this hits boths the top level `.vscode` and `app/.vscode`

twrichards:
amen!

ripecosta (approved):
👍 
`launch.json` isn't being used by anyone right now and `settings.json` isn't doing what we were hoping it would do so I agree with removing it.
We can bring some kind of config back later. One that does what we need it to do! 😄

PR #612, opened 2021-03-05
Reminder cancellation page

tomrf1:
This is a standalone page which does not require sign in.
It is for use in reminder email links, for when users wish to unsubscribe to reminders.

Screen Shot 2021-03-05 at 10 55 37

On failure:
Screen Shot 2021-03-05 at 12 06 17


TomPretty (dismissed):
LGTM 👍

app/client/components/MMAPage.tsx
```diff
@@ -98,6 +101,8 @@ const MMARouter = () => (
         <AccountOverview path="/" />
         <Billing path="/billing" />
 
+        <CancelReminders path="/cancel-reminders" />
```
ripecosta, line 104: Could you please put this at the bottom of the Router after at a `// Does not require sign in` comment?
It will help anything that looks at it in the future better understand that this page doesn't require sign in.

app/shared/requiresSignin.ts
```diff
@@ -1,9 +1,19 @@
...
 
 // To avoid security vulnerabilities do not add public paths that do not end in a slash
-const publicPaths = ["/api/contact-us/", "/help-centre/", "/api/known-issues/"];
+const publicPaths = [
+  "/api/contact-us/",
+  "/help-centre/",
+  "/api/known-issues/",
+  "/cancel-reminders/",
+  "/api/reminders/cancel/"
+];
```
ripecosta, line 11: While you're here you would mind ordering this alphabetically? Thanks!

```suggestion
const publicPaths = [
  "/api/contact-us/",
  "/api/known-issues/",
  "/api/reminders/cancel/",
  "/cancel-reminders/",
  "/help-centre/"
];
```

app/shared/requiresSignin.ts
```diff
@@ -1,9 +1,19 @@
...
 
 export const requiresSignin = (path: string) => {
-  const normalizedPath = pathLib.normalize(path + "/");
+  const parsed = parse(path);
+  const normalizedPath = pathLib.normalize(
+    parsed.pathname + "/" + parsed.query
+  );
```
ripecosta, line 17: I don't understand the benefit of this change. Could you clarify?

ripecosta (changes requested):
Tiny changes requested. Will approve as soon as addressed.

app/shared/requiresSignin.ts
tomrf1: our new endpoint requires a querystring, this change makes that possible

app/shared/requiresSignin.ts
tomrf1: 👍

app/client/components/MMAPage.tsx
tomrf1: 👍 done

app/shared/requiresSignin.ts
ripecosta: We are using `startsWith` in the next line. Wouldn't that work without changes?

tomrf1: I think the old code just stuck a `/` at the end of the querystring

TomPretty: yeah I took a look at the old code just now - it was adding `/` after the query string which was then not being matched, so it would redirect to the sign in page

ripecosta: `'/cancel-reminders/<querystringparam>/'.startsWith('/cancel-reminders/')` would still return true as desired though. No?

ripecosta: I just got it!
You were using query string params, which mean the comparison would actually be `'/cancel-reminders?<querystringparam>/'.startsWith('/cancel-reminders/')`, not what I posted on the previous comment.

ripecosta (approved):
👍

PR #617, opened 2021-03-09
Create help centre nav component

marialani:
## What does this change?
This includes a navigation component for the help centre which will be used to navigate between topics. Not currently in use.

## Images
| Mobile | Desktop |
| --- | --- |
| image | image |

### DIFF
--- a/app/client/components/helpCentre/helpCentreNav.tsx
+++ b/app/client/components/helpCentre/helpCentreNav.tsx
@@ -1,0 +1,144 @@
+import { css } from "@emotion/core";
+import { space } from "@guardian/src-foundations";
+import { neutral } from "@guardian/src-foundations/palette";
+import { textSans } from "@guardian/src-foundations/typography";
+import { Link } from "@reach/router";
+import React, { useState } from "react";
+import { maxWidth, minWidth } from "../../styles/breakpoints";
+import { HelpCentreNavConfig, helpCentreNavConfig } from "./helpCentreConfig";
+import {
+  containterCss,
+  innerSectionCss,
+  innerSectionDivCss,
+  linkAnchorStyle,
+  linkArrowStyle,
+  sectionTitleCss
+} from "./helpCentreStyles";
+
+interface HelpCentreNavProps {
+  selectedTopicObject?: HelpCentreNavConfig;
+}
+
+const ulCss = css`
+  list-style: none;
+  margin: 0 0 ${space[6]}px 0;
+  padding: 0;
+  ${maxWidth.desktop} {
+    display: none;
+  }
+`;
+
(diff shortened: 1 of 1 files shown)


app/client/components/helpCentre/helpCentreNav.tsx
```diff
@@ -0,0 +1,144 @@
...
+  padding: ${space[4]}px ${space[3]}px;
+  margin: 0;
+`;
+
+const HelpCentreNav = (props: HelpCentreNavProps) => {
+  const [open, setOpen] = useState<boolean>(false);
+
+  const handleSectionClick = () => () => {
+    setOpen(!open);
+  };
```
ripecosta, line 61: This will also work. (together with the suggestion on line 98)
```suggestion
  const handleSectionClick = () => setOpen(!open);
 ```

File: app/client/components/helpCentre/helpCentreNav.tsx [hunk 2]
```diff
@@ -0,0 +1,144 @@
...
+        <h2
+          css={css`
+            ${sectionTitleCss(open, false)};
+            ${textSans.large({ fontWeight: "bold" })};
+            ${minWidth.tablet} {
+              padding: ${space[3]}px ${space[3] * 2 + 15}px ${space[3]}px
+                ${space[5]}px;
+            }
+          `}
+          onClick={handleSectionClick()}
```
ripecosta, line 98: This will also work. (together with the suggestion on lines 59-61) 
```suggestion
          onClick={handleSectionClick}
```

app/client/components/helpCentre/helpCentreNav.tsx
```diff
@@ -0,0 +1,144 @@
...
+                  css={css`
+                    ${innerSectionDivCss};
+                    border-bottom: ${topicIndex < helpCentreNavConfig.length - 1
+                      ? "1px solid #dcdcdc"
+                      : ""};
+                    cursor: pointer;
+                    ${minWidth.tablet} {
+                      padding-left: ${space[3]}px;
+                    }
+                  `}
```
ripecosta, line 121: Can these css chunks be put in a separate function similarly to what you did with `liCss`? It will make the component more readable.

ripecosta: There's also other examples in this file of css chunks that could go into separate functions or variables like you did with `liCss` and `ulCss` but I won't leave individual comments for each. But doing so would improve this component.

ripecosta (changes requested):
Solid component! Good to raise smaller PRs like this too.

I believe improving how we declare the css properties can further improve it. Could you also add a screenshot of the nav to the PR description?

PR #619, opened 2021-03-09
Add Help Centre see all topics page

marialani:
## What does this change?
This adds a see all topics page to the help centre. An endpoint for the page has been included but commented out for now. 

## Images
| Desktop | Mobile |
| --- | --- |
| image | image |
| manage thegulocal com_help-centre_topic_more-topics | manage thegulocal com_help-centre_topic_billing |


app/client/components/helpCentre/helpCentreSeeAllTopics.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    }}
+                  >
+                    {question.title}
+                  </a>
+                  <span css={linkArrowStyle} />
+                </li>
+              ))}
+            </ul>
+          )}
+          {!helpCentreConfigObject && <HelpCentreMoreTopics />}
```
ripecosta, line 93: Because `HelpCentreMoreTopics` is its own component I think it would be best to have the rendering of the single topic be it's own component as well. Since those two are parallel to each other.
That way this component's responsibility would be to place the page furniture and determine whether it should render the `HelpCentreSingleTopic` or the `HelpCentreMoreTopics` component.
What do you think?

ripecosta (changes requested):
Left some minor suggestions. 
Could you also add screenshots for the More Topics page to the PR description? Thanks.

app/client/components/helpCentre/helpCentreSeeAllTopics.tsx
```diff
@@ -0,0 +1,113 @@
...
+`;
+
+const HelpCentreSeeAll = (props: HelpCentreProps) => {
+  const selectedTopicObject = helpCentreNavConfig.find(
+    topic => topic.id === props.topicCode
+  );
+
+  const helpCentreConfigObject = helpCentreConfig.find(
+    topic => topic.id === selectedTopicObject?.id
+  );
```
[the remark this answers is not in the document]
marialani: Agreed `selectedNavTopic` or `selectedNavTopicObject` is a better name for line 46 but on line 50 it's a bit trickier to name. If line 50 is changed to `selectedTopic` on line 93 there'd be `!selectedTopic` which wouldn't be right. 

Though this will all change when things are transferred over to s3 as it's down to the way the `helpCentreConfig.ts` file is currently organised. I'm not sure what name would best communicate what line 50-52 is attempting to do...

app/client/components/helpCentre/helpCentreSeeAllTopics.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <SectionPageContainer
+        hasNav={true}
+        selectedTopicObject={selectedTopicObject}
+      >
+        <div
+          css={css`
+            margin-bottom: ${space[24]}px;
+          `}
+        >
+          <h2 css={h2Css}>{selectedTopicObject?.title}</h2>
```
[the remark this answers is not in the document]
marialani: It needs to be `selectedTopicObject` here because `helpCentreConfigObject` doesn't include "more topics"

app/client/components/helpCentre/helpCentreSeeAllTopics.tsx
marialani: I don't think I understand what you mean here exactly and what do you mean by `HelpCentreSingleTopic`?

If I understand that part correctly as far as I know the "more topics" page will contain a dropdown for all of the 'other' topics. I don't think that each subtopic in "more topics" would have it's own page/component

marialani:
> Left some minor suggestions.
> Could you also add screenshots for the More Topics page to the PR description? Thanks.

Sure I can add another row in the table with images of the More Topics pages. They do adhere to the figma file in visual display, the dropdown just replaces the individual topic links

app/client/components/helpCentre/helpCentreSingleTopic.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { HelpCentreTopic } from "./helpCentreConfig";
+import {
+  linkAnchorStyle,
+  linkArrowStyle,
+  linkListItemStyle,
+  linksListStyle
+} from "./helpCentreStyles";
+
+interface HelpCentreSingleTopicProps {
+  selectedHelpCentreConfigTopic: HelpCentreTopic;
```
ripecosta, line 13: This components doesn't have knowledge that there are several topics elsewhere. It merely render what is passed. Neither does it need to have context that the passed topic comes from a config. For those reasons I suggest renaming it to `topic`.

app/client/components/helpCentre/helpCentreTopic.tsx
```diff
@@ -0,0 +1,91 @@
...
+  padding-top: ${space[4]}px;
+  border-top: 1px solid ${neutral["86"]};
+`;
+
+const HelpCentreTopic = (props: HelpCentreTopicProps) => {
+  const selectedNavTopic = helpCentreNavConfig.find(
+    topic => topic.id === props.topicCode
+  );
+
+  const selectedHelpCentreConfigTopic = helpCentreConfig.find(
```
ripecosta, line 39: Just a reminder from our conversation yesterday. `selectedTopic` would suffice.

ripecosta (changes requested):
I requested tiny changes. Let me know how you feel about them.

app/client/components/helpCentre/helpCentreTopic.tsx
marialani: I suppose with changing this to the if blocks it's clear enough what `selectedTopic` would be now and that this is only temporary code until the config is moved over to s3

ripecosta (approved):
👍

PR #621, opened 2021-03-12
Adjust the spacing of the known issues section in the help centre

marialani:
## What does this change?
This adds top padding to the known issues section and removes inline styling

## Images
| Before | After |
| image | image |


ripecosta (approved):
👍 👍 👍

PR #585, opened 2021-02-08
Move Contact Us into Help Centre

ripecosta:
## What does this change?
The Contact Us form is part of the Help Centre. However, because it was created first it got its own separate URL. This PR moves it from `https//manage.theguardian.com/contact-us` to `https://manage.theguardian.com/help-centre/contact-us`.

PR #625, opened 2021-04-15
Fix scrollToTop incorrectly triggering in the Contact Us form

ripecosta:
## What does this change?
Fixes a bug introduced when we moved the Contact Us path from `/contact-us/` to `/help-centre/contact-us/` in #585 and forgot to update the exception in `scrollToTop`.
This bug made the page jump to the top when a user made a selection in the form.


marialani (approved):
👍

PR #628, opened 2021-04-21
Make article and topic pages available

ripecosta:
## What does this change?
This PR makes the Help Centre's article and topic pages accessible so they can be tested in PROD. Given that they are not linked to from anywhere crawlers should not pick up on them (we don't want them to for now) and users shouldn't stumble upon them.


kelvin-chappell (approved):
🚀

PR #633, opened 2021-05-24
Explicitly pass credentials on fetch requests

ripecosta:
## What does this change?
The `api/reminders/status` introduced in PR #604 did not include the fix introduced in PR #572 so users with older Safari versions accessing the Email & Marketing section of MMA were seeing an error and were unable to change their settings. This PR adds the fix to the relevant endpoint and restores access.

### DIFF
--- a/app/client/components/identity/idapi/supportReminders.ts
+++ b/app/client/components/identity/idapi/supportReminders.ts
@@ -22,7 +22,10 @@ const getConsent = (isActive: boolean): ConsentOption => ({
 });
 
 export const read = async (): Promise<ConsentOption[]> => {
-  const response = await fetch(REMINDERS_STATUS_ENDPOINT);
+  const response = await fetch(REMINDERS_STATUS_ENDPOINT, {
+    credentials: "include",
+    mode: "same-origin"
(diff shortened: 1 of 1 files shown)


TomPretty (approved):
LGTM 👍

twrichards:
in a follow-up PR it might be worth moving the params object into constant, and re-use for ALL fetch requests to avoid recurring PRs like this one and https://github.com/guardian/manage-frontend/pull/572

ripecosta:
> in a follow-up PR it might be worth moving the params object into constant, and re-use for ALL fetch requests to avoid recurring PRs like this one and #572
> 
> You could experiment with writing a little wrapper around fetch (something along the lines of https://stackoverflow.com/questions/44820568/set-default-header-for-every-fetch-request) which could encapsulate these options and standard headers etc. notably `MDA_TEST_USER_HEADER` as per [#572 (comment)](https://github.com/guardian/manage-frontend/pull/572#discussion_r563749749))

I was thinking about something like that, yeah. We can't get to it right now but I've put a card in our Health board so it's not forgotten.

PR #636, opened 2021-05-25
Add lang attribute to docs

kelvin-chappell:
Apparently, without a `lang` attribute screen readers will interpret text as being in the default language.
https://www.w3.org/WAI/WCAG21/Understanding/language-of-page

I'm not sure whether that's true but this is my first PR in the repo!


paulbrown1982 (approved):
👍

ripecosta (approved):
That's pretty cool. Thanks, Kelvin!

PR #638, opened 2021-05-27
Set page title of Help Centre articles to correspond with URL

kelvin-chappell:
Currently all Help Centre pages have the same page title.
This change gives articles their correct individual page title and keeps all other Help Centre content on the default page title.

The format is `Help Centre | The Guardian | <article title>` for articles and `Help Centre | The Guardian` for all other Help Centre content.


app/server/routes/helpCentreFrontend.ts
```diff
@@ -1,4 +1,5 @@
 import { Request, Response, Router } from "express";
+import { defaultPageTitle } from "../../client/components/helpCentre/helpCentrePageTitle";
```
ripecosta, line 2: Hi Kelvin,

In this project we avoid having the server-side code import anything from the client-side code and vice-versa. When both client and server sides need a particular variable or function we put it in the shared folder.

app/client/components/helpCentre/helpCentrePageTitle.tsx
```diff
@@ -0,0 +1,23 @@
+import { Location } from "@reach/router";
+import React from "react";
+
+interface PageTitleProps {
+  title?: string | undefined;
+}
+
+export const PageTitle = (props: PageTitleProps) => (
```
ripecosta, line 8: The standard for this project is that the main component you are exporting is reflected in the file name. So in this case the file name would be `pageTitle.tsx`

app/server/routes/helpCentreFrontend.ts
kelvin-chappell: That makes sense.  Presumably this isn't even accessible to the serverside

rBangay (approved):

PR #639, opened 2021-05-28
Update back to help centre link to use client-side routing

marialani:
## What does this change?
The 'Back to Help Centre' link in the header is a regular link. This PR changes it to a client-side router link so the page doesn't reload.


kelvin-chappell (approved):
👍

PR #641, opened 2021-05-28
Redesign known issues section of the help centre

marialani:
## What does this change?
Updates the design of the Help Centre Alert feature so that this component complies with design and similar features elsewhere on the site.

## Images
| Before  | After |
| image | image  |

### DIFF
--- a/app/client/components/helpCentre/knownIssues.tsx
+++ b/app/client/components/helpCentre/knownIssues.tsx
@@ -92,8 +92,7 @@ export const KnownIssues = () => {
 
   const iconCss = css`
     position: absolute;
-    top: 50%;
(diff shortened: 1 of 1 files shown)


app/client/components/helpCentre/knownIssues.tsx
```diff
@@ -104,7 +103,9 @@ export const KnownIssues = () => {
   `;
 
   const aCss = css`
+    ${textSans.medium()};
     text-decoration: underline;
+    color: black;
```
paulbrown1982, line 108: Is there a palette value for this?

marialani: Ah yes, will update to the source palette

PR #644, opened 2021-06-03
Add robots.txt to control site indexing

kelvin-chappell:
This should regulate well-behaved bots indexing our content to ensure that, in production, they only try to access the new Help Centre content.

In pre-prod stages no content is allowed to be indexed.

Prod:
```
UserAgent: *
Allow: /help-centre
Allow: /help-centre/
Disallow: /

User-agent: AdsBot-Google
User-agent: AdsBot-Google-Mobile
Disallow: /
```

Pre-prod:
```
UserAgent: *
Disallow: /
```

See:
* https://www.theguardian.com/robots.txt
* https://developers.google.com/search/docs/advanced/robots/create-robots-txt
* https://developers.google.com/search/docs/advanced/crawling/overview-google-crawlers
* https://en.wikipedia.org/wiki/Robots_exclusion_standard

### DIFF
--- a/app/server/routes/core.ts
+++ b/app/server/routes/core.ts
@@ -1,5 +1,6 @@
 import { Request, Response, Router } from "express";
 import { authKeysAreFetchableMemoisedHealthcheck } from "../apiGatewayDiscovery";
+import { Environments } from "../config";
 import { withIdentity } from "../middleware/identityMiddleware";
 
 const router = Router();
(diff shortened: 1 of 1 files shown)


app/server/routes/core.ts
```diff
@@ -21,4 +22,23 @@ router.get("/_prout", (_, res: Response) => {
...
+    "UserAgent: *\n" +
+    "Allow: /help-centre\n" +
+    "Allow: /help-centre/\n" +
+    "Disallow: /\n\n";
+  const disallowGoogleAdsBots =
+    "User-agent: AdsBot-Google\n" +
+    "User-agent: AdsBot-Google-Mobile\n" +
+    "Disallow: /\n\n";
+  const prodAccessList = allowHelpCentre + disallowGoogleAdsBots;
+  const preProdAccessList = disallowAll;
```
ripecosta, line 37: I believe `nonProdAccessList` would be clearer name for this. Not all code goes through other stages before going to PROD.

ripecosta (approved):
Looks good! 👍

app/server/routes/core.ts
```diff
@@ -21,4 +22,23 @@ router.get("/_prout", (_, res: Response) => {
...
+  const allowHelpCentre =
+    "UserAgent: *\n" +
+    "Allow: /help-centre\n" +
+    "Allow: /help-centre/\n" +
+    "Disallow: /\n\n";
+  const disallowGoogleAdsBots =
+    "User-agent: AdsBot-Google\n" +
+    "User-agent: AdsBot-Google-Mobile\n" +
+    "Disallow: /\n\n";
+  const prodAccessList = allowHelpCentre + disallowGoogleAdsBots;
```
ripecosta, line 36: Why do we want to block Google's ad bots?

kelvin-chappell: Just to avoid unnecessary load on the site.

rBangay (approved):

app/server/routes/core.ts
kelvin-chappell: I was thinking pre-prod as opposed to prod or post-prod, so any stage before prod regardless of whether there actually is such a stage.

johnduffell: it took a few years(!) but we realised that NODE_ENV was set to production in both CODE and PROD, so actually CODE was being indexed by google
fixed in https://github.com/guardian/manage-frontend/pull/1475

PR #645, opened 2021-06-04
Fix logic to choose a robots.txt

kelvin-chappell:
There was a small bug in my logic!

[Environments](https://github.com/guardian/manage-frontend/blob/main/app/server/config.ts#L23-L26) is just an enum so `Environments.PRODUCTION` is always false.


ripecosta (commented):
Port merge approval 👍

PR #647, opened 2021-06-09
Remove moment

rBangay:
## What does this change?

As per the momentjs documentation page:

`"We would like to discourage Moment from being used in new projects going forward"`
https://momentjs.com/docs/

Moment is now considered a legacy project and we cannot rely on updates being made to it. 

There are also practical reasons for moving away from moment such as moments lack of support for tree shaking.

For the vast majority of MMA we were only making use of a few parsing and formatting methods of moment. These have been replaced with the `date-fns` https://date-fns.org/ library which is tree shakeable so we only load the parts that we use. These methods have also been consolidated into `/shared/dates.ts` which should hopefully make any future migration simpler.

As well as the date methods that we were using from moment we were also using an external dependency `react-daterange-picker` for the calendar date selection in the holiday stops journey. This dependency was also dependent on moment, which means the holiday stop journey has been reworked to remove this dependency .

## Images

gzipped bundle size before             |  gzipped bundle size after
  |  

### bye bye moment
timesupcombined


app/client/components/accountoverview/accountOverviewCancelledCard.tsx
```diff
@@ -3,7 +3,7 @@ import { space } from "@guardian/src-foundations";
 import { brand, brandAlt, neutral } from "@guardian/src-foundations/palette";
 import { textSans } from "@guardian/src-foundations/typography";
 import React from "react";
-import { formatDateStr } from "../../../shared/dates";
+import { parseDate } from "../../../shared/dates";
```
paulbrown1982, line 6: It would have saved a lot of updated lines if we had retained `formatDateStr` and had it wrap `parseDate`? It doesn't require any additional info. Is this simply continuing a change already made on other files in the platform?

rBangay: Hey, that's an interesting point. The parseDate function does a bit more than the original formatDateStr function which just takes a date string input and spits out a date string output ... the parseDate function takes a date string input (or nothing if you want now) and spits out an object with a date object, a date string (which can be in any custom output format) and various helper functions. 

I guess where we are only interested in imputing a date string and outputting a date string and there isn't any subsequent date manipulation then I could recreate the formatDateStr function in dates.ts and as you said have it wrap the parseDate function, something like this:

```
export const formatDateStr = (inputDateStr: string, outputDateFormat:string) => {
  return parseDate(inputDateStr).dateStr(outputDateFormat);
}
```

An assumption is then made that the input date string format is always `"yyyy-MM-dd"; // example: 1969-07-16`

If that isn't a safe assumption then I would have to change the shape of the formatDateStr function and change everywhere that uses it.

app/client/components/delivery/records/deliveryRecords.tsx
```diff
@@ -103,11 +107,12 @@ export const checkForExistingDeliveryProblem = (
...
   records.findIndex(deliveryRecord => {
-    const recordDateEpoch = moment(deliveryRecord.deliveryDate).unix();
-    const fourteenDaysAgoEpoch = moment()
-      .startOf("day")
-      .subtract(14, "days")
-      .unix();
+    const recordDateEpoch = parseDate(
+      deliveryRecord.deliveryDate
+    ).date.valueOf();
+    const now = new Date();
```
ripecosta, line 113: It looks like we only use `now` in one place. Might we get rid of it as a variable to reduce cognitive burden?

app/package.json
```diff
@@ -144,24 +143,23 @@
     "@sentry/node": "^5.22.3",
     "@stripe/react-stripe-js": "^1.1.2",
     "@stripe/stripe-js": "^1.7.0",
+    "@types/react-date-range": "^1.1.4",
```
ripecosta: Was this left over from a test?

app/client/components/billing/invoicesTable.tsx
```diff
@@ -35,7 +34,7 @@ interface InvoicesTableProps {
 export const InvoicesTable = (props: InvoicesTableProps) => {
   const [
     trackingPaginationInteractionCount,
-    setTrackingPaginationInteractionCount
+    setTrackingPaginationInteractionCount,
```
ripecosta, line 37: I believe our linter removes trailing commas. Could you check if the linter ran correctly on this file before merging?

app/client/components/delivery/address/deliveryAddressEditConfirmation.tsx
```diff
@@ -5,11 +5,11 @@ import {
   brand,
   neutral,
   success,
-  text
+  text,
```
ripecosta, line 8: I believe this file have not have been parsed by the linter either.

app/client/components/delivery/address/deliveryAddressForm.tsx
```diff
@@ -29,7 +29,7 @@ import { Button } from "@guardian/src-button";
 import { brand, neutral } from "@guardian/src-foundations/palette";
 import { headline, textSans } from "@guardian/src-foundations/typography";
 import { Link, navigate } from "@reach/router";
-import moment from "moment";
+import {parseDate} from "../../../../shared/dates";
```
ripecosta, line 32: Linter?

app/client/components/delivery/records/deliveryAddressStep.tsx
```diff
@@ -5,21 +5,21 @@ import { space } from "@guardian/src-foundations";
...
 import Color from "color";
-import moment from "moment";
 import React, {
   ChangeEvent,
   Dispatch,
   FormEvent,
   SetStateAction,
   useContext,
-  useState
+  useState,
```
ripecosta, line 14: Linter?

app/client/components/holiday/holidaySelectionInfo.tsx
```diff
@@ -0,0 +1,127 @@
+import {error} from "@guardian/src-foundations/palette";
```
ripecosta, line 1: Linter?

app/client/components/delivery/address/deliveryAddressFormContext.tsx
```diff
@@ -1,6 +1,5 @@
-import { Moment } from "moment";
 import { createContext } from "react";
-import { friendlyLongDateFormat } from "../../../../shared/dates";
+import {DATE_FNS_LONG_OUTPUT_FORMAT, dateString} from "../../../../shared/dates";
```
ripecosta, line 2: Linter?

ripecosta (changes requested):
Thanks, Richard. Happy to see moment go.

There seems to be an issue with linting in some of the files. Once that's checked I'm happy to approve this PR

app/package.json
ripecosta: Addressed. Thanks!

ripecosta (approved):
Ready to go! 👍

ripecosta (approved):
👍  All good!

app/client/components/datePicker.tsx
```diff
@@ -161,40 +155,31 @@ export const DatePicker = (props: DatePickerProps) => (
...
-                    )
+                    range: props.amendableDateRange
                   }
                 ]
               : [])
-          ].sort((a, b) => a.range.start.unix() - b.range.start.unix())}
-          defaultState="available"
-          firstOfWeek={1}
-          daysOfWeekToIconify={props.issueDaysOfWeek}
-          dateToAsterisk={props.dateToAsterisk}
```
twrichards, line 197: the functionality associate with this prop needs to be restored

PR #654, opened 2021-06-15
KnownIssues container fix

marialani:
## What does this change?
The outer container for the knownIssues component was rendered regardless of whether there were no issues i.e. when `issuesData` was empty. This added unnecessary markup and css to the DOM when there were no issues. This PR updates the component to conditionally return markup dependent on whether there are issues.


ripecosta (approved):
👍  Good work on this.

PR #653, opened 2021-06-15
Upgrade node riffraff artefact -> artifact

ripecosta:
## What does this change?
This PR upgrades us to `@guardian/node-riffraff-artifact`. The library we were using is not being actively maintained. This one should bring some improvements, particularly better logging.

## How to test
Two ways to test:

1. Run `yarn node-riffraff-artifact --dryRun` locally to confirm artifact config is working correctly. See the [library page](https://github.com/guardian/node-riffraff-artifact#how-to-use) for more details on expected behaviour
1. Attempt to deploy on CODE to confirm that the artifact config is working correctly with the CloudFormation and Riff-Raff config.

## How can we measure success?
Our continuous deployment pipelines still works. Which it does.

image

PR #656, opened 2021-06-15
Fix deployment pipeline/update compression used for build files

ripecosta:
## What does this change?
Updates the build compression to ZIP and fixes a deployment bug introduced in #653.

The library we upgraded to in #653 made us change the compression of the built files to tar (tgz isn't supported). However we did not update the bash scripts in the CloudFormation to reflect this change. This caused the deployment to succeed but instead of deploying the latest build (inside `manage-frontend.tar`) it was deploying the last version deployed before #653 was merged (inside `manage-frontend.tgz`). This was detected because @prout-bot (🙇 ) warned us that the PR was overdue on PROD. No other PRs were merged since so that was the only merge affected. This bug is now fixed.

While researching the cause of this bug we realised that while the average compressed file size was 16.5MB when using tgz compression, tar compression increases that to 108.5. We tested zip compression, which lower it to 26MB. Because tgz is not longer an updates, zip compression was used instead.

Big thank you to @twrichards for pairing with me on this.

Note: There were a number of linting changed done to this file. I will mark them to make it easier to review.

### DIFF
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -95,7 +95,7 @@ Resources:
       ImageId: !Ref AMI
       SecurityGroups:
         - !Ref InstanceSecurityGroup
-        - !Ref WazuhSecurityGroup 
+        - !Ref WazuhSecurityGroup
       InstanceType: !FindInMap [StageVariables, !Ref Stage, InstanceType]
       IamInstanceProfile: !Ref InstanceProfile
       AssociatePublicIpAddress: false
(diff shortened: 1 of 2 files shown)


cfn.yaml
```diff
@@ -229,29 +229,29 @@ Resources:
...
-          - arn:aws:s3::*:membership-dist/*
+          - Effect: Allow
+            Action: s3:GetObject
+            Resource:
+              - arn:aws:s3::*:membership-dist/*
       Roles:
-      - !Ref AppRole
-        
+        - !Ref AppRole
+
```
ripecosta, line 254: Linting changes

cfn.yaml
```diff
@@ -353,10 +353,10 @@ Resources:
...
         Ref: VpcId
       SecurityGroupEgress:
-      - IpProtocol: tcp
-        FromPort: 1514
-        ToPort: 1515
-        CidrIp: 0.0.0.0/0
+        - IpProtocol: tcp
+          FromPort: 1514
+          ToPort: 1515
+          CidrIp: 0.0.0.0/0
```
ripecosta, line 359: Linting changes

cfn.yaml
```diff
@@ -229,29 +229,29 @@ Resources:
...
       PolicyDocument:
         Statement:
-        - Effect: Allow
-          Resource: "*"
-          Action:
-          - ec2:DescribeTags
-          - ec2:DescribeInstances
-          - autoscaling:DescribeAutoScalingGroups
-          - autoscaling:DescribeAutoScalingInstances
+          - Effect: Allow
```
ripecosta, line 232: Changes from this line onwards are merely linting changes.

twrichards (approved):

rBangay (approved):

PR #657, opened 2021-06-17
Add help centre article feedback widget

marialani:
## What does this change?
This adds the  'Did you find the information you need?' feedback widget at the bottom of Article pages. The buttons track the user's click

## Images
| Desktop  | Mobile | 
| image  | image |

### DIFF
--- a/app/client/components/helpCentre/helpCentreArticle.tsx
+++ b/app/client/components/helpCentre/helpCentreArticle.tsx
@@ -208,8 +208,8 @@ export const ArticleFeedbackWidget = (props: ArticleFeedbackWidgetProps) => {
               onClick={() => {
                 setFeedBackButtonClicked(true);
                 trackEvent({
-                  eventCategory: "help-centre-article-page",
-                  eventAction: "article-feedback-widget-click",
+                  eventCategory: "help-centre",
+                  eventAction: "article-feedback",
                   eventLabel: props.articleCode,
                   eventValue: 1
                 });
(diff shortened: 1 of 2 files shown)


app/client/components/helpCentre/helpCentreArticle.tsx
```diff
@@ -136,4 +146,103 @@ const ArticleBody = (props: ArticleBodyProps) => {
...
+                  width: initial;
+                }
+              `}
+              onClick={() => {
+                setFeedBackButtonClicked(true);
+                trackEvent({
+                  eventCategory: "help-centre-article-page",
+                  eventAction: "article-feedback-widget-click",
+                  eventLabel: props.articleCode,
+                  eventValue: 1
```
ripecosta, line 214: The `eventCategory` for the Help Centre entries needs to be `help-centre` as we do elsewhere in the Help Centre so its categorised correctly in GA and Ophan.

If you wanted to it would be safe to simplify `eventAction` to `article-feedback` as 'widget' and 'click' don't provide any addition info, given this is the only piece of article feedback being tracked and the only way to interact with it is clicking.

app/client/components/helpCentre/helpCentreArticle.tsx
```diff
@@ -136,4 +146,103 @@ const ArticleBody = (props: ArticleBodyProps) => {
...
+                  width: initial;
+                }
+              `}
+              onClick={() => {
+                setFeedBackButtonClicked(true);
+                trackEvent({
+                  eventCategory: "help-centre-article-page",
+                  eventAction: "article-feedback-widget-click",
+                  eventLabel: props.articleCode,
+                  eventValue: 0
```
ripecosta, line 235: Same as above.

The `eventCategory` for the Help Centre entries needs to be `help-centre` as we do elsewhere in the Help Centre so its categorised correctly in GA and Ophan.

If you wanted to it would be safe to simplify `eventAction` to `article-feedback` as 'widget' and 'click' don't provide any addition info, given this is the only piece of article feedback being tracked and the only way to interact with it is clicking.

ripecosta (changes requested):
Good work. Needs a change to the `eventCategory` but once that's done I'm happy  to approve it.

ripecosta (approved):
👍

rBangay (approved):

PR #658, opened 2021-06-18
Remove moment

rBangay:
## What does this change?
attempt no. 2

As per the momentjs documentation page:

`"We would like to discourage Moment from being used in new projects going forward"`
https://momentjs.com/docs/

Moment is now considered a legacy project and we cannot rely on updates being made to it. 

There are also practical reasons for moving away from moment such as moments lack of support for tree shaking.

For the vast majority of MMA we were only making use of a few parsing and formatting methods of moment. These have been replaced with the `date-fns` https://date-fns.org/ library which is tree shakeable so we only load the parts that we use. These methods have also been consolidated into `/shared/dates.ts` which should hopefully make any future migration simpler.

As well as the date methods that we were using from moment we were also using an external dependency `react-daterange-picker` for the calendar date selection in the holiday stops journey. This dependency was also dependent on moment, which means the holiday stop journey has been reworked to remove this dependency .

## Images

gzipped bundle size before             |  gzipped bundle size after
  |  

### bye bye moment
timesupcombined

### DIFF
--- a/app/artifact.json
+++ b/app/artifact.json
@@ -1,0 +1,16 @@
+{
+  "projectName": "support:manage-frontend",
+  "vcsURL": "https://github.com/guardian/manage-frontend",
+  "actions": [
+    {
+      "action": "manage-frontend-cloudformation",
+      "path": "../cfn.yaml",
+      "compress": false
+    },
+    {
+      "action": "manage-frontend",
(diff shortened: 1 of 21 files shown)


ripecosta (approved):
Even better than last time!

PR #659, opened 2021-06-21
[RB] - set explicit date values in tests when comparing

rBangay:
## What does this change?
When comparing 2 dates like the following:

`dateIsSameOrAfter(new Date(), new Date())`

There seems to be a small chance the the 2 dates will not be the same, eg off by a fraction of a second.

Where 2 dates that are supposed to be equal are being compared in the tests I have set them explicitly to be the date of judgement day in Terminator!

This only affected the test run and has no effect on the code.

thumbsup

### DIFF
--- a/app/client/__tests__/dates.ts
+++ b/app/client/__tests__/dates.ts
@@ -59,8 +59,7 @@ describe("dateIsBefore", () => {
   const inputDate = judgementDay;
   const comparisonDate = judgementDay;
   it("returns true if input date is before comparison date", () => {
-    comparisonDate.setMinutes(comparisonDate.getMinutes() + 5);
-    expect(dateIsBefore(inputDate, comparisonDate)).toBeTruthy();
+    expect(dateIsBefore(inputDate, new Date())).toBeTruthy();
   });
   it("returns false if input date is after comparison date", () => {
     comparisonDate.setMinutes(comparisonDate.getMinutes() - 10);
--- a/app/client/__tests__/dates.ts
+++ b/app/client/__tests__/dates.ts
@@ -22,6 +22,9 @@ import {
 // formatting date-fns strings documentation here:
 // https://date-fns.org/v2.17.0/docs/format
 
+const judgementDay = new Date(1997, 7, 29);
+const backToTheFutureDay = new Date(1985, 9, 26);
+
 describe("parseDate", () => {
   it("parses date strings correctly", () => {
     const pd = parseDate("2020-11-30");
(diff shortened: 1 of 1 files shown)


kelvin-chappell (commented):
Perhaps
`const judgementDay = new Date(1997, 7, 29)`
before the tests to keep DRY.

kelvin-chappell (dismissed):
👍

app/client/__tests__/dates.ts
```diff
@@ -59,8 +59,7 @@ describe("dateIsBefore", () => {
   const inputDate = judgementDay;
   const comparisonDate = judgementDay;
   it("returns true if input date is before comparison date", () => {
-    comparisonDate.setMinutes(comparisonDate.getMinutes() + 5);
-    expect(dateIsBefore(inputDate, comparisonDate)).toBeTruthy();
+    expect(dateIsBefore(inputDate, new Date())).toBeTruthy();
```
kelvin-chappell:
```
const comparisonDate = new Date();
it("returns true if input date is before comparison date", () => {
  expect(dateIsBefore(inputDate, comparisonDate)).toBeTruthy();
```

app/client/__tests__/dates.ts
```diff
@@ -53,11 +56,10 @@ describe("dateString", () => {
...
 
 describe("dateIsBefore", () => {
-  const inputDate = new Date();
-  const comparisonDate = new Date();
+  const inputDate = judgementDay;
+  const comparisonDate = judgementDay;
   it("returns true if input date is before comparison date", () => {
-    comparisonDate.setMinutes(comparisonDate.getMinutes() + 5);
-    expect(dateIsBefore(inputDate, comparisonDate)).toBeTruthy();
+    expect(dateIsBefore(inputDate, new Date())).toBeTruthy();
```
ripecosta, line 62: I think you may have forgotten to follow the same pattern as below on this one where you use `comparisonDate` instead of `new Date()`.

ripecosta (approved):
This looks good to go but think you might have missed a change (mentioned in comments). If it was deliberate feel free to merge. 👍

app/client/__tests__/dates.ts
rBangay: > I think you may have forgotten to follow the same pattern as below on this one where you use `comparisonDate` instead of `new Date()`.

This one should be a valid test, we are not expecting the input and comparison date to be the same and so it will not be affected by the same issue as declaring 2 new dates side by side.   

The first test in this describe block checks that the input date (1997) is before new Date (today). The comparison date variable is used for the subsequent 2 falsey tests where we want the comparison date to less than and then equal to the input date so that we can be sure the test fails correctly.

PR #660, opened 2021-06-23
[RB] - create help center outer page furniture wrapper

rBangay:
## What does this change?
Creates a separate help center page furniture component that wraps the suspense router in the help center.

This means that the routes/components within the router are only concerned with their unique content and cuts down on duplication.

Another benefit is that the new wrapper component sits outside of the router meaning the user will have less time to wait to see the page layout before the async content is loaded.

## Images
Screenshot 2021-06-23 at 10 16 28

`HelpCenterContentWrapper` is the new component here, where previously the page furniture was duplicated in the child components eg `<HelpCentre  /> <HelpCentreArticle /> <HelpCentreTopic /> <ContactUs />`


app/client/components/sectionContent.tsx
```diff
@@ -65,18 +65,28 @@ const divCss = (hasNav: boolean | undefined) => css`
...
 `;
 
+export const SelectedTopicObjectContext = React.createContext<
+  Dispatch<SetStateAction<HelpCentreNavConfig | undefined>>
+>(undefined!);
+
 export const SectionContent = (props: SectionContentProps) => {
+  const [selectedTopicObject, setSelectedTopicObject] = useState<
+    HelpCentreNavConfig | undefined
+  >(props.selectedTopicObject);
```
ripecosta, line 75: `selectedTopicObject` is no longer being passed in as a prop so I think we can removed it on line 12 & initialise this state as `undefined`.

app/client/components/sectionContent.tsx
```diff
@@ -65,18 +65,28 @@ const divCss = (hasNav: boolean | undefined) => css`
   }
 `;
 
+export const SelectedTopicObjectContext = React.createContext<
+  Dispatch<SetStateAction<HelpCentreNavConfig | undefined>>
```
ripecosta, line 69: The use the the name `HelpCentreNavConfig` is looking ackward to me given how we're using it. It's also not making sense to me that we pass both the id and the title of the nav in order to select (we only used the `id` in `HelpCentreNav`). Would you be able to improve these two points as part of this PR?

ripecosta (commented):
Will you take through this PR when you have the chance? There's a couple things regarding the use of Context that I'm not sure I'm clear on.

app/client/components/helpCentre/helpCentreArticle.tsx
```diff
@@ -53,16 +52,20 @@ const HelpCentreArticle = (props: HelpCentreArticleProps) => {
...
   }, [props.articleCode]);
 
-  const selectedNavTopic = helpCentreNavConfig.find(
-    topic => topic.id === article?.topics[0].path
-  );
+  const setSelectedTopicObject = React.useContext(SelectedTopicObjectContext);
+  useEffect(() => {
+    const selectedNavTopic = helpCentreNavConfig.find(
+      topic => topic.id === article?.topics[0].path
+    );
```
ripecosta, line 59: I don't think you need this bit anymore because you pass `selectedNavTopic.id` on line 61. You're ok passing `article?.topics[0].path` directly (topic ID and path are synonyms on this system). Not having this check would remove the need to import `helpCentreNavConfig` on this file.

app/client/components/helpCentre/helpCentreTopic.tsx
```diff
@@ -49,17 +48,19 @@ const HelpCentreTopic = (props: HelpCentreTopicProps) => {
...
   }, [props.topicCode]);
 
-  const selectedNavTopic = helpCentreNavConfig.find(
-    topic => topic.id === props.topicCode
-  );
+  const setSelectedTopicObject = React.useContext(SelectedTopicObjectContext);
+  useEffect(() => {
+    const selectedNavTopic = helpCentreNavConfig.find(
+      topic => topic.id === props.topicCode
+    );
```
ripecosta, line 55: As in Article I don't think you need this check anymore. You can pass `props.topicCode` directly on line 57 and that way not import `helpCentreNavConfig` on this file at all.

ripecosta (approved):
image

PR #662, opened 2021-06-28
[RB] - Help centre topic nav fix

rBangay:
## What does this change?
Sync the `website` and `subscriptions` id's in the `helpCentreConfig.ts` to match the json object structure in s3. 

Update the highlight of the selected nav item when navigating between options so that the selected option is always highlighted

## Images
Screenshot 2021-06-28 at 10 44 25


kelvin-chappell (approved):
👍

PR #663, opened 2021-07-01
[RB] Add callCenterEmailAndNumbers component to the bottom of the help centre article pages

rBangay:
## What does this change?
Add callCenterEmailAndNumbers component to the bottom of the help centre article pages

## Images
Before             |  After
  |

### DIFF
--- a/app/client/components/helpCentre/helpCentreArticle.tsx
+++ b/app/client/components/helpCentre/helpCentreArticle.tsx
@@ -7,6 +7,7 @@ import { captureException, captureMessage } from "@sentry/browser";
 import React, { useEffect, useState } from "react";
 import { minWidth } from "../../styles/breakpoints";
 import { trackEvent } from "../analytics";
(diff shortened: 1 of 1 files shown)


app/client/components/helpCentre/helpCentreArticle.tsx
```diff
@@ -7,6 +7,7 @@ import { captureException, captureMessage } from "@sentry/browser";
 import React, { useEffect, useState } from "react";
 import { minWidth } from "../../styles/breakpoints";
 import { trackEvent } from "../analytics";
+import { CallCentreEmailAndNumbers } from "../callCenterEmailAndNumbers";
```
jambo1, line 10: Is this the same module that the live chat button is being added to, i.e. will live chat also be present on every article page?

jambo1: Discussed, and should be fine

jambo1 (approved):

PR #668, opened 2021-07-23
[RB] - update to liveagent embed url

rBangay:
## What does this change?
As a consequence of the following work:

https://trello.com/c/D5K0PO1h/2446-live-agent-create-dedicated-site

The embed url needs updating slightly 🔗 

`https://guardiansurveys.secure.force.com/liveagent` and `https://dev1-guardiansurveys.cs88.force.com/liveagent` are the new urls.

The other changes are linting corrections that override a previous forced push


ripecosta (approved):
👍

PR #669, opened 2021-07-27
Link to new Help Centre articles and topics

kelvin-chappell:
This rewrites all the links to legacy help pages from the Help Centre landing page so that they point at new pages in the Help Centre.
Eventually, this page will be published in the same way as the article and topic pages but at the moment it has to be built manually.

I've tested the links in Code.

### DIFF
--- a/app/client/components/helpCentre/HelpTopicBox.tsx
+++ b/app/client/components/helpCentre/HelpTopicBox.tsx
@@ -111,9 +111,8 @@ export const HelpTopicBox = (props: HelpTopicBoxProps) => (
         margin: auto 11px 20px 11px;
       `}
     >
-      <a
-        href={props.topic.seeAllLink}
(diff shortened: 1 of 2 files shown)


ripecosta (commented):
Hi Kelvin,

The place where we place these links on the page will need to be updated to use internal router linking to avoid the browser reloading the page.

It's quite simple. [Here](https://github.com/guardian/manage-frontend/blob/main/app/client/components/sectionHeader.tsx#L111)'s an example.

kelvin-chappell:
@ripecosta Thanks.  Does [that](https://github.com/guardian/manage-frontend/pull/669/commits/423933defaf26f6ec06faf6eaa5f43284d2bcb36) look right?

ripecosta (approved):
Read for launch! 🚀

PR #672, opened 2021-08-02
Add Google Search Console verification file

kelvin-chappell:
This is to give me access to the [Google Search Console](https://search.google.com/search-console) for manage.theguardian.com.

I'm using the [HTML file upload](https://support.google.com/webmasters/answer/9008080) method.  This appears to be the only way to get access to the search console for a subdomain where the root is authenticated.

Tested access to the file on Code.


ripecosta (approved):
👍

PR #673, opened 2021-08-03
Give all bots access to sitemap

kelvin-chappell:
A small oversight in the robots.txt file meant that the sitemap couldn't be read.

Google Search Console told me that Google couldn't read the sitemap and [this tool](https://www.google.com/webmasters/tools/robots-testing-tool) told me why.


ripecosta (approved):
👍

PR #674, opened 2021-08-03
Allow bots to index pages built on clientside

kelvin-chappell:
Currently, Google bots are only indexing the serverside-built content of pages because they are blocked from accessing `/static/helpcentre.js`.  This is to give access to static resources so that all pages can be meaningfully indexed.


ripecosta (approved):
👍

PR #675, opened 2021-08-03
Enable bots to access API to load pages

kelvin-chappell:
Google still can't quite build clientside-composed pages as it needs access to api paths.

If this doesn't work, I'll give full access to the site.  That was how it was until last week anyway so shouldn't cause any unexpected problems.


ripecosta (approved):

ripecosta (approved):
👍

PR #682, opened 2021-08-19
[RB] - consolidate the feature switch functionality for live chat

rBangay:
## What does this change?
Consolidate the feature switch functionality for live chat into a url param / session storage solution.

Previously the url param and a seperate variable were being used for the instantiation of live chat and the display of the "Start live chat" button.


ripecosta (approved):
Thanks for consolidating and for isolating the logic for it. Looks great 👍

PR #686, opened 2021-08-27
[RB] - livechat overflow-y scroll for small breakpoints

rBangay:
## What does this change?
The live chat window should have overflow-y set to scroll on smaller breakpoints where the content is taller than the window height.

## Images
Screenshot 2021-08-27 at 15 11 44


ripecosta (approved):
👍

ripecosta (approved):
👍

PR #687, opened 2021-08-31
[RB] - consolidate font sizes of start chat and chat ended system messages

rBangay:
## What does this change?
Consolidate font sizes of start chat and chat ended system messages

## Images
Screenshot 2021-08-31 at 10 59 08


ripecosta (approved):
👍

PR #690, opened 2021-09-06
consecutive live chats bug fix

rBangay:
## What does this change?
Remove iframe source param passed through to the initESW function the second/subsequent time a live chat message is established. This resolves a cross origin postMessage error when running in the salesforce dev environment. eg:

```Failed to execute ‘postMessage’ on ‘DOMWindow’: The target origin provided (‘https://service.force.com’) does not match the recipient window’s origin (‘https://gnmtouchpoint--dev1.my.salesforce.com’).```


ripecosta (approved):
👍

PR #692, opened 2021-09-06
[RB] - copy update and minor adjustment to warning icon position

rBangay:
## What does this change?
Adds custom css styling to the queue position state of live chat.

## How to test
- run manage-frontend locally
- visit https://manage.thegulocal.com/help-centre/?liveChat=1
- click on Start live chat button
- simulate the loading error of the salesforce script (`esw.min.js`). One way of doing this is to reject instead of resolve the promise in the `initLiveChat` function in liveChat.tsx

## Images
Before             |  After
  |


kelvin-chappell (approved):
👍

PR #697, opened 2021-09-13
[RB] - correct styling/text colour of the end chat button

rBangay:
## What does this change?
Fixes a bug where only 1 button exists at the end of a conversation where previously there were 2 expected. Now uses even and odd slectors to ensure that the button background and text colours are correct.

## How to test
- log into the dev salesforce envirmonment
- from the App Launcher menu select LiveChatTestApp
- in the omni channel window set your self as Busy
- run manage locally and visit the help center with the live chat url param: `https://manage.thegulocal.com/help-centre/?liveChat=1` 
- click on the `Start live chat` button
- fill in the pre-chat form
- click on the `Start chatting` button
- go back to salesforce and end the chat from there
- go back to manage and you should now see the close chat button with the correct button text colour

## Images
Before                           |  After
Screenshot 2021-09-13 at 12 08 31  |  Screenshot 2021-09-13 at 12 09 26

### DIFF
--- a/app/client/components/liveChat/liveChatCssOverrides.ts
+++ b/app/client/components/liveChat/liveChatCssOverrides.ts
@@ -84,23 +84,23 @@ export const liveChatCss = css`
   .embeddedServiceSidebarButton {
     border-radius: 0;
   }
-  .dialogButtonContainer button:not(:last-of-type) {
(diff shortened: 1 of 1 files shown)


ripecosta (approved):
👍

PR #698, opened 2021-09-14
[RB] - style anchor link colour and text decoration for agent messages

rBangay:
## What does this change?
The agent speech bubbles in the live chat window are dark blue, the copy colour is white, which the anchor links should also be, along with underline text decoration.

## How to test
- log into the dev salesforce envirmonment
- from the App Launcher menu select LiveChatTestApp
- in the omni channel window set your self as Busy
- run manage locally and visit the help center with the live chat url param: `https://manage.thegulocal.com/help-centre/?liveChat=1` 
- click on the `Start live chat` button
- fill in the pre-chat form
- click on the `Start chatting` button
- go back to salesforce and send a message to the user containing a url
- go back to manage and you should now see a link in the chat message window with the correct text colour and decoration

## Images
Before                           |  After
Screenshot 2021-09-14 at 09 25 46  |  Screenshot 2021-09-14 at 09 26 27

### DIFF
--- a/app/client/components/liveChat/liveChatCssOverrides.ts
+++ b/app/client/components/liveChat/liveChatCssOverrides.ts
@@ -146,4 +146,8 @@ export const liveChatCss = css`
   .messageArea .chatSessionStartTime {
     font-size: 13px;
   }
+  .chatContent .chat-content.agent a {
+    color: ${neutral["100"]};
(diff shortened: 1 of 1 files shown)


ripecosta (approved):
👍

PR #703, opened 2021-09-16
[RB] - fix svg transform rotate bug in safari

rBangay:
## What does this change?
Safari does not support transformation of the base svg tag 😖. It does however support transform of a group (g) tag. This pr wraps the inner contents of the thumbs up icon in a g tag and moves the rotation transform from the parent svg tag into the g tag.

## How to test
- run manage-frontend locally
- visit an article page:
  - https://manage.thegulocal.com/help-centre/article/i-need-to-pause-my-delivery

### Before
Screenshot 2021-09-16 at 09 37 41

### After
Screenshot 2021-09-16 at 09 38 32


ripecosta (approved):
👍

PR #704, opened 2021-09-16
[RB] - load the dev sandbox live chat version if you are anywhere but on prod

rBangay:
## What does this change?
conditionally load the dev salesforce live chat sandbox or the prod version dependant on the manage client environment.

Add the `onSettingsCallCompleted` event listener which when loading the prod version of salesforce live chat seems to fire a lot quicker than the `onAvailability` event.


ripecosta (approved):
👍

PR #706, opened 2021-09-21
Eslint migration

rBangay:
## What does this change?
tslint is deprecated this pr migrates manage to use eslint instead.

The eslint default rules are slightly different than the tslint ones that we were using so we have fixed most of the eslint errors. The remaining ones were turned off:

```
rules: {
    "no-prototype-builtins": ["off"],
    "react/no-unescaped-entities": ["off"],
    "react/display-name": ["off"],
    "@typescript-eslint/ban-types": ["off"],
    "@typescript-eslint/explicit-module-boundary-types": ["off"],
    "@typescript-eslint/no-inferrable-types": ["off"],
    "@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "_" }],
    "@typescript-eslint/ban-ts-comment": [
      "error",
      {
        "ts-expect-error": "allow-with-description",
        "ts-ignore": "allow-with-description",
        "ts-nocheck": "allow-with-description",
        "ts-check": "allow-with-description",
        minimumDescriptionLength: 10
      }
    ]
  }
  ```

Of interest is the following rule: `react/display-name` which should be looked at again in the future.

The `@typescript-eslint/no-explicit-any` rule will now throw a warning, which will be addressed in a future pr.


ripecosta (approved):
👍  Great work. Thanks for this!

PR #707, opened 2021-10-01
Fetch wrapper function

jonflynng:
## What does this change?
`credentials` and `mode` parameters were introduced in several fetch calls to avoid the Safari bug issued in this PR: https://github.com/guardian/manage-frontend/pull/572. This is a wrapper function that injects these parameters so they do not have to be written for every fetch call.

## How to test
Tested visually through dev tools network tab using Safari 11 and 12 on Browserstack.

_Any suggestions on a better name for the function?_


rBangay:
LGTM ... congrats on your first pr :+1

rBangay (approved):

PR #708, opened 2021-10-06
refactoring 'any' types

jonflynng:
## What does this change?
Migrates untyped code to have stricter types.

## How to test
This PR only changes types so there are no changes to the final bundled javascript.

### DIFF
--- a/app/client/components/identity/idapi/user.ts
+++ b/app/client/components/identity/idapi/user.ts
@@ -32,24 +32,24 @@ type UserPrivateFields = Partial<
 };
 
 interface UserAPIResponse {
-  user: {
-    id: string;
-    consents: [
-      {
-        id: string;
-        consented: boolean;
-      }
-    ];
-    publicFields: UserPublicFields;
-    privateFields: UserPrivateFields;
-    primaryEmailAddress: User["primaryEmailAddress"];
-    statusFields: {
-      userEmailValidated: boolean;
-    };
-  };
+  user: IdapiUserDetails;
 }
 
-type UserFromAPIResponse = UserAPIResponse["user"];
+interface IdapiUserDetails {
+  id: string;
+  consents: [
+    {
+      id: string;
+      consented: boolean;
+    }
+  ];
+  publicFields: UserPublicFields;
+  privateFields: UserPrivateFields;
+  primaryEmailAddress: User["primaryEmailAddress"];
+  statusFields: {
+    userEmailValidated: boolean;
+  };
+}
 
 interface UserAPIRequest {
   publicFields: UserPublicFields;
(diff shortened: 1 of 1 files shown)


app/client/components/identity/idapi/user.ts
```diff
@@ -49,6 +49,8 @@ interface UserAPIResponse {
   };
 }
 
+type UserFromAPIResponse = UserAPIResponse["user"];
```
ripecosta, line 52: What do you think about extracting `UserAPIResponse.user` into its own type and using it inside `UserAPIResponse` instead, as a way to improve clarity of this type hierarchy?

app/client/components/identity/idapi/fetch.ts
```diff
@@ -30,17 +32,17 @@ const getAPIOptionsForMethod = (method: string) => (
...
   url: string,
   options?: RequestInit
-): Promise<any> => {
+): Promise<T> => {
   const response = await fetch(baseUrl + url, options);
   if (!response.ok) {
     return await handleResponseFailure(response);
   } else if (response.status === 204) {
-    return null;
+    return null as any;
```
ripecosta, line 43: Should this not be `return null as T`?

app/client/components/identity/useConsentOptions.ts
```diff
@@ -41,7 +41,7 @@ const reducer = (state: State, action: Action) => {
       trackEvent({
         eventCategory: "emailPrefError",
         eventAction: "error",
-        eventLabel: payload.toString()
+        eventLabel: (payload as string).toString()
```
ripecosta, line 44: If `payload` is already a String, why do we need to use `.toString()`?

app/client/components/svgs/tickIcon.tsx
```diff
@@ -1,10 +1,9 @@
 import React from "react";
 
-export const TickIcon = (props: any) => (
+export const TickIcon = () => (
```
ripecosta, line 3: How come we removed this argument? Do we not have code that passes it anywhere anymore?

app/client/components/wizardRouterAdapter.tsx
```diff
@@ -15,7 +15,7 @@ interface RouteableProps extends RouteComponentProps {
 }
 
 interface RoutablePropsWithChildren extends RouteableProps {
-  children?: any; // TODO ReactElement<RouteableProps> | ReactElement<MultiRouteableProps>[];
+  children?: any; // JSX.Element | React.Component<RouteableProps> | React.FC<RouteableProps> | React.Component<MultiRouteableProps> | React.FC<MultiRouteableProps>[];
```
ripecosta, line 18: Can you add some detail as to why you're changed this comment? Did you meant to remove the `TODO` keyword?

PR #718, opened 2021-10-19
Call Ophan manually when subscribing to a newsletter

oliverlloyd:
## What does this change?
This PR replaces the use of the `data-link-name` attribute to capture click events related to newsletter subscriptions with a manual call to the `window.guardian.ophan.record` function.

## Why?
Because the click event was being fired twice, once as 'tick' and once as 'untick', which meant it was difficult to understand the data.

## What does `tick` mean?
I reversed this logic so that `tick` means a subscription event and `untick` means unsubscribe. Sounds obvious but the link-name had this the other way around which could have been an oversight or it could have been for a reason.

## How do we ensure these events are only passed for newsletter subscriptions?
The `MarketingPreference` component is reused for marketing preferences and also for newsletter subscriptions (which are a sort of marketing preference).

To ensure that we don't send click events to ophan labelled 'NEWSLETTER_SUBSCRIPTION` I am checking that `identityName` is present. Is this ideal? No, but from testing it seems to act as the right guard.

## Why are we adding `NewsletterPreference`?
I've broken the `MarketingPreference` out into two components. The action of subscribing to a newsletter is, arguably, different to setting marketing preferences and deserves its own component so that this newsletter specific functionality can be isolated. This also allows the `MarketingPreference` to be simplified and the props for each to be more targeted.

### Could this refactor go further?
I started to pull the thread of breaking newsletters apart from consent completely but that refactor would touch a lot of files and ultimately pollute this PR. We're only aiming to enhance tracking for newsletter subscriptions for now.

### DIFF
--- a/app/client/__tests__/components/identity/__snapshots__/NewsletterPreference.tsx.snap
+++ b/app/client/__tests__/components/identity/__snapshots__/NewsletterPreference.tsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`MarketingPreference component renders correctly and displays marketing information 1`] = `
+exports[`NewsletterPreference component renders correctly and displays marketing information 1`] = `
 <div
   css={
     Array [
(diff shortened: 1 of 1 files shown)


oliverlloyd:
[Data: Ensure we’re able to attribute confirmed sign-ups from the embeds (including footer and MMA)](https://trello.com/c/oLlMk4jO/389-data-ensure-were-able-to-attribute-confirmed-sign-ups-from-the-embeds-including-footer-and-mma)

PR #724, opened 2021-11-01
Upping the threshold for 5xx alarm to reduce noise

jambo1:
## What does this change?

Making 5xx alarm less sensitive to reduce noise and only alert for significant issues.


kelvin-chappell (approved):
👍

PR #734, opened 2021-11-11
[RB] - update babel dependencies

rBangay:
## What does this change?

Update babel dependencies to latest version.

Remove deprecated babel/polyfill and replace with corejs and regenerator-runtime/runtime


kelvin-chappell (approved):
Looks good to me 👍

PR #736, opened 2021-11-16
Removing Live Chat opening hours copy

joelisnothere:
Opening hours for Live Chat are increasing. We believe we can remove opening hours entirely from the Live Chat component in the Help Centr
Screenshot 2021-11-16 104040
Screenshot 2021-11-16 113231
e.


jambo1 (approved):
Looks good!

PR #742, opened 2021-12-01
Payments switching two step flow

jonflynng:
## What does this change?

Allows Direct Debit users (currently reviewing UI for paypal users switching into card) to change payment method to card, and also completes the switching in two steps, compared to the current process which takes 3 steps, 4 including the recaptcha screen.

 I've noted that Stripe Elements is buggy with testing libraries and have added an issue on their github.

Currently, this would be merged to a feature branch, and then when the entire feature is completed it would be merged back to main. the feature branch will have to be kept updated with main branch. I'm open to suggestions on a better approach to this. 

## How to test

Click on 'manage payment method' on a subscription in account overview. Use card number '4242 4242 4242 4242' to test successful payments. Stripe has many test numbers to test different error states and also 3D Secure/2 factor authentication: https://stripe.com/docs/testing. For Direct Debit, use account number 55779911 and sort code 200000 to test.


jonflynng:
[Create two step payment update flow](https://trello.com/c/9hnLM2n2/2694-create-two-step-payment-update-flow)

app/client/__tests__/components/payment/cardInputForm.tsx
```diff
@@ -0,0 +1,109 @@
+import React from "react";
+// import { render, fireEvent, waitFor } from "@testing-library/react";
```
rBangay, line 2: probably don't need this line here ... it should be pretty clear we need to add waitFor back in when/if we get to uncomment out the stripe tests

app/client/__tests__/components/payment/currentPaymentDetails.tsx
```diff
@@ -0,0 +1,147 @@
...
+      const { getByText } = render(
+        <CurrentPaymentDetails {...directDebitProductDetail} />
+      );
+
+      getByText("ending 911");
+      getByText("20-00-00");
+    });
+  });
+
+  describe("For Paypal", () => {});
```
rBangay, line 146: Do we need this here?

app/client/components/OverlayLoader.tsx
```diff
@@ -0,0 +1,63 @@
...
+	padding: 0;
+	-webkit-tap-highlight-color: transparent;
+	position: fixed;
+	left: 0;
+	top: 0;
+	width: 100%;
+	height: 100%;
+	text-align: center;
+    ${textSans.medium()};
+`
```
rBangay, line 27: It's weird that the pre commit hook didn't pretty this ... maybe `pretty-quick` doesn't know what to do with the template literal string :s

app/client/__tests__/components/payment/currentPaymentDetails.tsx
jonflynng: I filled this in in the third PR

rBangay (approved):

app/client/components/OverlayLoader.tsx
rBangay: as an update to this. We needed to update the version of pretty-quick after bumping our node version

rBangay (approved):

PR #743, opened 2021-12-01
Update Snyk workflow to monitor, rather than test, dependencies

kelvin-chappell:
Failing tests because of unrelated dependencies isn't helpful.
Instead we just want to keep Snyk up to date with the latest state of the codebase.


jambo1 (approved):
This is definitely a good idea as the red snyk warning is always there and confusing. 
We'll need to make sure we have our own process to stay on top of dependencies as we don't have snyk weeks any more and won't have these reminders

PR #744, opened 2021-12-02
Explicitly set date locale for consistent behaviour in dev and CI

jamesmockett:
## What does this change?

Currently our date tests pass on CI, but fail when ran locally.

This is due to calling the `toLocaleDateString()` method which uses the system's locale. On CI the locale appears to be `en-US` so this method returns dates with a `mm/dd/yyyy` format. eg. 25 December 2021 is returned as `12/25/2021`. Your local environment is likely to have the locale set to `en-GB` so dates are returned with a `dd/mm/yyyy` format. eg. `25/12/2021`.

This replaces calls to `toLocaleDateString()` with a `localiseDate` method that uses `Intl.DateTimeFormat` with the locale explicitly set to `en-GB` to format dates consistently without relying on the system's locale.

Additionally, instances of the `Date()` constructor being instantiated with year, month and day parameters have been replaced with ISO date strings to avoid ambiguity due to months being zero indexed. eg. 26 October 1985 changes from `new Date(1985, 9, 26)` to `new Date("1985-10-26")`.  (Notice how October is 9 rather than 10 in the first example.)

## How to test

- Run `yarn test` locally and ensure that the full test suite passes.
- Check this branch in Team City and ensure that the test suit still passes.

### DIFF
--- a/app/client/__tests__/components/dateInput.tsx
+++ b/app/client/__tests__/components/dateInput.tsx
@@ -8,19 +8,19 @@ Enzyme.configure({ adapter: new Adapter() });
 describe("DateInput", () => {
   it.each([
     {
-      givenDate: "10 Jan 2022",
+      givenDate: "2022-01-10",
       expectedDay: 10,
       expectedMonth: 1,
       expectedYear: 2022
     },
     {
-      givenDate: "16 Mar 2022",
+      givenDate: "2022-03-16",
       expectedDay: 16,
       expectedMonth: 3,
       expectedYear: 2022
     },
     {
-      givenDate: "4 Dec 2023",
+      givenDate: "2023-12-04",
       expectedDay: 4,
       expectedMonth: 12,
       expectedYear: 2023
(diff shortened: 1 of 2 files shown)


jamesmockett:
[dates.ts tests passing/failing based on the local date settings](https://trello.com/c/BecoxBNI/2720-datests-tests-passing-failing-based-on-the-local-date-settings)

rBangay (approved):

jamesmockett:
@rBangay Node includes all locales by default as of `v13.0.0`:
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#2019-10-22-version-1300-current-bethgriggs

Previously we were running `v12.22.7` which only includes the `en-US` locale, but switched to `v17.1.0` a couple of weeks ago which includes all locales. That explains why the tests only recently begun failing as calls to `toLocaleDateString()` would have begun using the system locale (which should be `en-GB` by default) rather than `en-US`. TeamCity must be using `en-US` so the tests continued to pass.

PR #746, opened 2021-12-03
 local phone number in the settings form is a number

QuarpT:
## What does this change?

- local phone number in the settings form is a number and needs to be converted to string before sending to IDAPI endpoint

## Why

Identity API is becoming less tolerant of numbers being passed in place of strings

### DIFF
--- a/app/client/components/identity/idapi/user.ts
+++ b/app/client/components/identity/idapi/user.ts
@@ -76,7 +76,9 @@ export const isErrorResponse = (error: any): error is UserAPIErrorResponse => {
 const toUserApiRequest = (user: Partial<User>): UserAPIRequest => {
   const { countryCode: countryCode, localNumber: localNumber } = user;
   const telephoneNumber =
-    countryCode && localNumber ? { countryCode, localNumber } : undefined;
+    countryCode && localNumber
+      ? { countryCode, localNumber: `${localNumber}` }
+      : undefined;
 
   return {
     publicFields: {
--- a/app/client/components/identity/models.ts
+++ b/app/client/components/identity/models.ts
@@ -38,7 +38,7 @@ export interface User {
   postcode: string;
   country: string;
   countryCode: string;
-  localNumber: string;
+  localNumber: number;
 }
 
 export interface UserError {


app/client/components/identity/idapi/user.ts
```diff
@@ -76,7 +76,9 @@ export const isErrorResponse = (error: any): error is UserAPIErrorResponse => {
 const toUserApiRequest = (user: Partial<User>): UserAPIRequest => {
   const { countryCode: countryCode, localNumber: localNumber } = user;
   const telephoneNumber =
-    countryCode && localNumber ? { countryCode, localNumber } : undefined;
+    countryCode && localNumber
+      ? { countryCode, localNumber: `${localNumber}` }
+      : undefined;
```
pvighi, line 81: maybe a stupid question but why did we need this change here ?

app/client/components/identity/models.ts
```diff
@@ -38,7 +38,7 @@ export interface User {
   postcode: string;
   country: string;
   countryCode: string;
-  localNumber: string;
+  localNumber: number;
```
pvighi, line 41: would this affect the input validation or anything like that ?

QuarpT: It shouldn't, the TypeScript types should only be used at compile time

app/client/components/identity/idapi/user.ts
QuarpT: convert localNumber to string before passing it to IDAPI.
The Identity API expects a string - but right now the JSON parser is very lenient so numbers are accepted

pvighi: ah I might be misunderstanding this code then,  is this localNumber not the same localNumber that was redefined to be a string in models.ts?

pvighi: ah sorry, I now see that it was actually changed the other way around, it used to be a string and now it's a number 🤔

QuarpT: the form input is a number, then we convert it to a string before sending it to idapi

app/client/components/identity/models.ts
pvighi: sorry for some reason I thought you were changing this the other way areound,  why does this need to be a number here ?

QuarpT: this represents the type of the form input field, which is numeric
Before, it was set to string which is incorrect

pvighi (approved):
looks good to me

PR #750, opened 2021-12-08
[RB] - update client web socket url to include missing /ws path

rBangay:
## What does this change?

The webpack dev server websocket connection has been failing since the webpack version bump. This appeared to be affecting  hot reloading and potentially some chunk loading errors (not noticed this error since implementing this fix).

Looks as though the socket-url param was wrong (missing the /ws path)


jamesmockett (approved):
A surprisingly straightforward fix, although I'm sure it took a fair bit of investigation to arrive here 😅

PR #753, opened 2021-12-10
[RB] - update version of prettier

rBangay:
## What does this change?

Bumps the version of prettier in order to stop pretty-quick from moaning about its version :)


kelvin-chappell (approved):
👍

PR #759, opened 2021-12-17
Upgrade Storybook and add Webpack 5 support

jamesmockett:
## What does this change?

Updates Storybook and dependencies, adding support for Webpack 5.

This also fixes the `deliveryRecordsProblemForm.tsx` and `step1.tsx` test which the package updates broke. The only changes to the tests are to replace `find(Button)` with `find("button")`, although it's not entirely clear why the former no longer works. (_Note:_ the diffs for the tests looks bigger due to Prettier reformatting them.)


rBangay (approved):
Nice one 👍

PR #760, opened 2021-12-20
Selective dependency resolutions for `micromatch`

jamesmockett:
## What does this change?

Adds selective dependency resolution for `micromatch` package to avoid dependency on `v0.4.3` of `set-value` which has a critical security vulnerability.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -168,5 +168,9 @@
     "winston": "^3.2.1",
     "yup": "^0.28.2"
   },
+  "resolutions": {
+    "chokidar": "^3.5.2",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -168,5 +168,9 @@
     "winston": "^3.2.1",
     "yup": "^0.28.2"
   },
+  "resolutions": {
```
jambo1, line 171: Didn't know this was an option, thanks!

jambo1 (approved):

jamesmockett:
Closing as we've temporarily removed Storybook which removes this dependency entirely.

PR #761, opened 2021-12-21
Payments switching feature cards

jonflynng:
## What does this change?

 Much anticipated payments switching functionality, allowing users with any payment method to switch to using credit or debit cards. The previous three step (four step including captcha) switching flow has now been reduced to just two steps. The Wiki will be updated in the following weeks with findings from the investigation and the workings of the new functionality.

Screenshot 2021-12-21 at 11 58 07

## How to test

Integration tests have been added for isolated component testing. For testing the whole flow, test card numbers can be found here: https://stripe.com/docs/testing. The default stripe testing number is 4242 4242 4242 4242. For Direct Debit, choose account number 55779911 and sort code 200000 for testing.


app/client/components/payment/update/CurrentPaymentDetail.tsx
```diff
@@ -0,0 +1,273 @@
...
+          }
+        `}
+      >
+        <div
+          css={css`
+            padding-bottom: ${space[3]}px;
+
+            ${minWidth.tablet} {
+              margin: ${space[6]}px 0 0 0;
+              padding: ${space[6]}px 0 0 0;
```
rBangay, line 130: these 2 lines get overwritten on lines 134 and 135

app/client/components/OverlayLoader.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Spinner } from "./spinner";
+
+interface OverlayLoaderProps {
+  message: string;
+}
+
+const overlay = css`
+  z-index: 10000;
+  display: block;
+  border: 0 none transparent;
```
jamesmockett, line 13: It doesn't look like this is doing anything?

app/client/components/OverlayLoader.tsx
```diff
@@ -0,0 +1,63 @@
...
+  margin: 0;
+  padding: 0;
+  -webkit-tap-highlight-color: transparent;
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  text-align: center;
+  ${textSans.medium()};
```
jamesmockett, line 26: Font size and line height are being set on the child `div` so not sure this is needed?

app/client/components/payment/update/CurrentPaymentDetail.tsx
```diff
@@ -0,0 +1,273 @@
...
+    margin: 0;
+    padding: 0;
+  `;
+
+  const keyCss = css`
+    ${textSans.medium({ fontWeight: "bold" })};
+    padding: 0 ${space[2]}px 0 0;
+    display: inline-block;
+    vertical-align: top;
+    width: 14ch;
```
jamesmockett, line 46: Nice use of `ch` units 👍

app/client/components/payment/update/ErrorSummary.tsx
```diff
@@ -0,0 +1,82 @@
...
+  padding: ${space[1]}px;
+  display: flex;
+`;
+
+const iconStyles = (color: string): SerializedStyles => css`
+  display: flex;
+  flex: 0 1 auto;
+  margin-top: 1px;
+  svg {
+    fill: ${color};
```
jamesmockett, line 40: You could set `color` once on the wrapper `div` and let is cascade down to the icon and message with `fill: currentColor` here. Not necessarily worth changing now, but maybe a future tweak to simplify things.

app/client/components/payment/update/CurrentPaymentDetail.tsx
jonflynng: thanks, there may be a couple of these in the PR

app/client/components/OverlayLoader.tsx
jonflynng: This sets the font family as well, otherwise it looks like Times New Roman or whatever the default is

app/client/components/OverlayLoader.tsx
jonflynng: not too sure tbh, I copied these styles directly from Support. https://github.com/guardian/support-frontend/tree/main/support-frontend/assets/components/progressMessage

app/client/components/spinner.tsx
```diff
@@ -38,6 +40,7 @@ export const Spinner = (props: LoadingProps) => (
         height: scaledPx(40, props.scale),
         animation: "spin 2s linear infinite",
         margin: scaledPx(10, props.scale),
+        marginTop: props.alignCenter ? "center" : "",
```
jamesmockett, line 43: Should this be an alignment property rather than `margin-top` as it's being set to `center`?

app/client/components/OverlayLoader.tsx
jamesmockett: Ah, okay, makes sense 👍

app/client/components/spinner.tsx
jonflynng: yeah, that's odd, it already has text-align center on it so removed this

rBangay (approved):

PR #769, opened 2021-12-23
Expired Recaptcha bug fix on card form

jonflynng:
## What does this change?

I noticed I was not removing the recaptcha token state upon expiry. This means that if the Recaptcha has expired after two minutes, and the user does not complete it, the generic error screen is shown and they have to refresh the page:

Screenshot 2021-12-23 at 10 03 26

I've checked Sentry and fortunately this has only happened once over the last day it has been deployed. 

Screenshot 2021-12-23 at 09 59 57

Link to the Sentry exception: https://sentry.io/organizations/the-guardian/issues/2875657110/?environment=theguardian.com&project=1208603&query=is%3Aunresolved

Have tested this manually and it correctly shows the 'Recaptcha has not been completed' error message.

### DIFF
--- a/app/client/components/payment/update/card/Recaptcha.tsx
+++ b/app/client/components/payment/update/card/Recaptcha.tsx
@@ -13,17 +13,17 @@ declare let window: Window & {
 const hrefStyle = {
   textDecoration: "underline",
   color: "inherit",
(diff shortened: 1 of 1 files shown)


jamesmockett (approved):
Looks good to me 👍

PR #773, opened 2022-01-05
Payments form style updates

jonflynng:
## What does this change?

UI updates to the payments form found here: https://www.figma.com/file/judOZzoY74FjzBdvu98Bvb/MMA-since-2020?node-id=6569%3A281466. 

Have also added vertical margin between the error messages which wasn't included when first pushing this live:

Screenshot 2022-01-05 at 15 46 38

### DIFF
--- a/app/client/components/payment/update/dd/directDebitInputForm.tsx
+++ b/app/client/components/payment/update/dd/directDebitInputForm.tsx
@@ -4,7 +4,7 @@ import { maxWidth } from "../../../../styles/breakpoints";
 import { space } from "@guardian/src-foundations";
 import { sans } from "../../../../styles/fonts";
 import { Button } from "@guardian/src-button";
-import { Checkbox } from "../../../checkbox";
+import { Checkbox } from "@guardian/src-checkbox";
 import { cleanSortCode } from "../../directDebitDisplay";
 import { FieldWrapper } from "../fieldWrapper";
 import { NewPaymentMethodDetail } from "../newPaymentMethodDetail";
@@ -155,7 +155,7 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
         css={css`
           display: flex;
           justify-content: flex-start;
-          margin-bottom: ${space[3]}px;
+          margin-bottom: 10px;
 
           ${minWidth.tablet} {
             margin-top: ${space[4]}px;
@@ -195,11 +195,22 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
           />
         </FieldWrapper>{" "}
       </div>
+
       <Checkbox
-        onChange={(newValue) => setSoleAccountHolderConfirmed(newValue)}
+        onChange={(e) => setSoleAccountHolderConfirmed(e.target.checked)}
         checked={soleAccountHolderConfirmed}
         label="I confirm that I am the account holder and I am solely able to authorise debit from the account"
         required
+        value="I confirm that I am the account holder and I am solely able to authorise debit from the account"
+        cssOverrides={css`
+          div {
+            padding: 10px 0;
+          }
+
+          :focus {
+            box-shadow: none !important;
+          }
+        `}
       />
 
       <DirectDebitLegal />
--- a/app/client/components/payment/update/dd/directDebitInputForm.tsx
+++ b/app/client/components/payment/update/dd/directDebitInputForm.tsx
@@ -210,11 +210,6 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
           label="I confirm that I am the account holder and I am solely able to authorise debit from the account"
           required
           value="I confirm that I am the account holder and I am solely able to authorise debit from the account"
-          cssOverrides={css`
-            :focus {
-              box-shadow: none !important;
-            }
-          `}
         />
       </div>
 
--- a/app/client/components/payment/update/dd/directDebitInputForm.tsx
+++ b/app/client/components/payment/update/dd/directDebitInputForm.tsx
@@ -210,6 +210,11 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
           label="I confirm that I am the account holder and I am solely able to authorise debit from the account"
           required
           value="I confirm that I am the account holder and I am solely able to authorise debit from the account"
+          cssOverrides={css`
+            & + span {
+              top: calc(50% - 8px);
+            }
+          `}
         />
       </div>
 
--- a/app/client/components/payment/update/dd/directDebitInputForm.tsx
+++ b/app/client/components/payment/update/dd/directDebitInputForm.tsx
@@ -155,7 +155,6 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
         css={css`
           display: flex;
           justify-content: flex-start;
-          margin-bottom: 10px;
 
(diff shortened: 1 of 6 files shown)


jonflynng:
[Last UI updates for payments switching](https://trello.com/c/LQKU7nE3/2781-last-ui-updates-for-payments-switching)

app/client/components/payment/update/updatePaymentFlow.tsx
```diff
@@ -128,32 +129,71 @@ const PaymentMethodRadioButton = (props: PaymentMethodRadioButtonProps) => {
     }
   `;
 
-  return (
+  const label = () => (
```
jonflynng, line 132: The following changes is due to radio buttons not being fully clickable. e.g.: hovering the mouse over the card logos would prevent the radio from being clicked.

app/client/components/payment/update/dd/directDebitInputForm.tsx
```diff
@@ -195,11 +195,22 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
...
       <Checkbox
-        onChange={(newValue) => setSoleAccountHolderConfirmed(newValue)}
+        onChange={(e) => setSoleAccountHolderConfirmed(e.target.checked)}
         checked={soleAccountHolderConfirmed}
         label="I confirm that I am the account holder and I am solely able to authorise debit from the account"
         required
+        value="I confirm that I am the account holder and I am solely able to authorise debit from the account"
+        cssOverrides={css`
+          div {
+            padding: 10px 0;
```
jamesmockett, line 207: Could this padding be applied elsewhere? It feels better not to touch the internal structure of Source components as it's liable to change with new versions. (I've pulled down the branch and can't see this being applied to anything within the checkbox, unless I'm looking in the wrong place?)

app/client/components/payment/update/dd/directDebitInputForm.tsx
```diff
@@ -195,11 +195,22 @@ export const DirectDebitInputForm = (props: DirectDebitUpdateFormProps) => {
...
+          }
+
+          :focus {
+            box-shadow: none !important;
```
jamesmockett, line 211: It looks like this is being done as the default focus state doesn't match the existing controls on the form. It's not ideal to remove the focus state entirely though. Could we make it match the others instead or removing it?

app/client/components/payment/update/fieldWrapper.tsx
```diff
@@ -93,9 +96,9 @@ export class FieldWrapper extends React.Component<
         <div
           css={{
             border: `${
-              this.state.error?.message
-                ? "4px solid " + error[400]
-                : "2px solid " + neutral[60]
+              (this.state.error?.message && "4px solid " + error[400]) ||
+              (this.state.focus && "2px solid " + brand[500]) ||
+              "2px solid " + neutral[60]
```
jamesmockett, line 101: Might be clearer to move this logic elsewhere rather than doing it all inline? Not sure how we handle this kind of thing elsewhere though, so up to you.

app/client/components/payment/update/updatePaymentFlow.tsx
```diff
@@ -128,32 +129,71 @@ const PaymentMethodRadioButton = (props: PaymentMethodRadioButtonProps) => {
...
+          padding: ${space[4]}px;
 
           div {
             font-weight: bold;
           }
         }
 
+        :hover {
+          -webkit-box-shadow: inset 0px 0px 0px 4px ${brand[500]};
+          -moz-box-shadow: inset 0px 0px 0px 4px ${brand[500]};
```
jamesmockett, line 185: I don't think you need to worry about prefixing `box-shadow` anymore as it's only needed for _really_ old browsers. (Doesn't Emotion include auto-prefixing anyway?)

jonflynng: good point. this was something i wanted to ask Richard about. I'll google it.

app/client/components/payment/update/dd/directDebitInputForm.tsx
jonflynng: Screenshot 2022-01-05 at 13 32 07

So the source component has a min-height of 44px, this means on desktop, when the checkbox label is clamped to one line, it acts like it has a padding of 10px vertically. 

Screenshot 2022-01-05 at 13 32 21

On mobile the component is taller because of the line breaks so there's no padding. That being said, padding is a bit odd so I'm going to wrap the source component in a div and add margin instead just for mobile. Also, note in the screenshot that if you go really small, the tick is oddly positioned. Does this happen on yours?

jonflynng: yeah, somethings not right here actually. Basically the source component has a min-height of 44px so it's being a bit tricky to work with. The vertical padding/margin needs to be 24px.

Screenshot 2022-01-05 at 13 32 07

Also, when the checkbox label is clamped to 3 lines on very small screens, the tick is out of place. happening both with and without the 10px padding I applied, does it happen on yours?

Screenshot 2022-01-05 at 13 32 21

jonflynng: So I need to sort the margin out, and then also make the font smaller on mobile to avoid the clamping issue?

jamesmockett: Yeah, I'm seeing the same thing with the tick when the label wraps on to multiple lines. I'm not sure there's much you can do about that though as it looks like a bug in the Source component itself. Was wondering if it's fixed in a newer version, but still seeing it in v3 and v4.

jamesmockett: It's because the checkbox and label are vertically centred, but the tick is absolutely positioned with `top: 15px`. Overriding that with `top: calc(50% - 8px);` does the job.

jonflynng: I fixed the vertical margin if you want to pull the branch again, overriding the source component styles don't seem to be working though, could you try please?

app/client/components/payment/update/fieldWrapper.tsx
jonflynng: have added the correct focus state to the form fields

app/client/components/payment/update/dd/directDebitInputForm.tsx
jonflynng: have added the correct focus state to the form fields

app/client/components/payment/update/updatePaymentFlow.tsx
jonflynng: yep, seems like emotion includes auto-prefixing by default

app/client/components/payment/update/dd/directDebitInputForm.tsx
jamesmockett: Nice work 👍 Much better for accessibility.

app/client/components/payment/update/dd/directDebitInputForm.tsx
jamesmockett: Managed to figure this one out:

```jsx
cssOverrides={css`
  & + span {
    top: calc(50% - 8px);
  }
`}
```

`cssOverrides` targets the `input` element directly and the `span` containing the tick is at the same level in the DOM.

jamesmockett (approved):
Looks great, nice work 👍

PR #775, opened 2022-01-06
remove `withIdentity` from the healthcheck 

twrichards:
…since HelpCentre needs to be available when identity is inaccessible 

PLUS don't 500 on endpoints which don't require signin

Co-authored-by: @jamesmockett 
Co-authored-by: @jonflynng 

## How to test
When deployed in CODE with Identity down, help centre should be accessible, the rest of manage should 500.

Users can access manage during the upcoming identity downtime.


twrichards:
> @jonflynng @jamesmockett might be worth addressing https://github.com/guardian/manage-frontend/pull/775/files#R153 if users are going to see `Internal Server Error` during the identity downtime of an hour

You might want to server side render this to reuse components/style - see previous usage of this in manage-frontend that was removed in https://github.com/guardian/manage-frontend/pull/565

kelvin-chappell (approved):
💯  Thanks for jumping in @twrichards

PR #776, opened 2022-01-10
Add minimal maintenance page

jamesmockett:
## What does this change?

This adds a minimal 'down for maintenance' page and route to the app. This is primarily in support of the upcoming Identity outage, but can also be used for any other planned maintenance or system outages.

## Images

Screenshot 2022-01-11 at 09 24 18

### DIFF
--- a/app/server/middleware/identityMiddleware.ts
+++ b/app/server/middleware/identityMiddleware.ts
@@ -3,7 +3,7 @@ import fetch from "node-fetch";
 import url, { UrlWithParsedQuery } from "url";
 import {
   getScopeFromRequestPathOrEmptyString,
-  X_GU_ID_FORWARDED_SCOPE
+  X_GU_ID_FORWARDED_SCOPE,
 } from "../../shared/identity";
 import { requiresSignin } from "../../shared/requiresSignin";
 import { handleAwsRelatedError } from "../awsIntegration";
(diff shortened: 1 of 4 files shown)


twrichards:
nice!
on the server-side, guessing you're going to add a redirect to this /maintenance, rather than returning 500?

jamesmockett:
> nice! on the server-side, guessing you're going to add a redirect to this /maintenance, rather than returning 500?

I couldn't get server-side rendering working as it was complaining about the use of hooks, so redirecting to a client-side page was one thought. Even that seems a bit fiddly to test locally as it throws up NGINX errors. The other thought was to temporarily change our Fastly config to redirect all traffic to the maintenance page. (This is what `support-frontend` are doing.)

However, whilst just testing this I noticed that Identity now have their own maintenance page so it seems like we don't even need to show this for the outage tomorrow.

jamesmockett:
[Show a helpful error message when Identity is unavailable](https://trello.com/c/JMFI9prK/2795-show-a-helpful-error-message-when-identity-is-unavailable)

app/server/middleware/identityMiddleware.ts
```diff
@@ -132,106 +133,101 @@ const redirectOrCustomStatusCode = (
...
-              [X_GU_ID_FORWARDED_SCOPE]:
-                req.header(X_GU_ID_FORWARDED_SCOPE) ||
-                getScopeFromRequestPathOrEmptyString(req.path),
-              Cookie: getCookiesOrEmptyString(req)
+) => express.RequestHandler =
+  (statusCodeOverride?: number) =>
+  (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    const errorHandler = (message: string, detail?: any) => {
+      handleAwsRelatedError(message, detail);
+      res.redirect("/maintenance");
```
jamesmockett, line 149: This line is the only code change in this file, replacing `sendStatus(500)` with a redirect to the maintenance page. The remainder of the diff is due to Prettier reformatting.

kelvin-chappell (approved):
Looks good to me! 👍

PR #777, opened 2022-01-12
Add Storybook

jamesmockett:
## What does this change?

This reinstates the temporarily removed Storybook dependencies, updates the config (adding global styles to ensure components render correctly and to allow viewport switching), and adds example stories for the `DateInput`, `Footer` and `HelpCentrePhoneNumbers` components.

Storybook has dependencies with known security vulnerabilities that were previously flagged by Dependabot. As Storybook is a dev dependency and not user facing these are low risk. To prevent these warnings from becoming noise (as we're unable to resolve them until Storybook is updated) and inadvertently masking vulnerabilities in user facing code we may wish to dismiss them once they have been investigated.

## Images

Screenshot 2022-01-13 at 17 54 10


jamesmockett:
[TASK: Set up Storybook](https://trello.com/c/qQkyYdAY/2801-task-set-up-storybook)

app/client/components/dateInput.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+    date: {
+      control: {
+        type: "date",
+      },
+    },
+  },
+} as ComponentMeta<typeof DateInput>;
+
+const Template: ComponentStory<typeof DateInput> = (args: DateInputProps) => {
+  args.date = new Date(args.date);
```
jamesmockett, line 24: The `date` prop expects a `Date` object, but Storybook's date control returns a string.

jonflynng (approved):
good work to get started, I don't understand much of the code so will have to take a look at their docs

PR #780, opened 2022-01-21
Breakout supporter exclusive section on Emails & Marketing Section

vlbee:
## What does this change?

In preparation of updating our consent section (content and UI/UX) this PR breaks out the product email consents into a new "Supporter Exclusive" section, which will only display if the user has purchased a product.  

All other UI/UX remains the same. End goal visible [here on Figma ](https://www.figma.com/file/WOztYf2GIbvowlvJeB3vmH/Identity---MMA?node-id=207%3A5892)

## Images

**BEFORE:** 
image

**NOW with Products**
image

**NOW without Products**
image


jamesmockett:
This looks great. My only real question was whether the `ConsentSection` component has been split out into two components in anticipation of future changes that are coming in a later PR? If not, the supporter email section could be moved within the existing component and wrapped in a conditional to avoid duplication. (ie. `consentPreference` and `consentPreferences` appear in both components.)

app/client/components/identity/EmailAndMarketing/SupporterEmailsSection.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const consentPreferences = (
+  consents: ConsentOption[],
+  clickHandler: ClickHandler
+) => consents.map((consent) => consentPreference(consent, clickHandler));
+
+export const SupporterEmailsSection: FC<ConsentSectionProps> = (props) => {
+  const { consents, clickHandler } = props;
+
+  return shouldDisplay(softOptInEmailConsents(consents)) ? (
```
jamesmockett, line 45: As `shouldDisplay` is returning a boolean you could use `&&` to conditionally render this instead of a ternary.

vlbee:
> This looks great. My only real question was whether the `ConsentSection` component has been split out into two components in anticipation of future changes that are coming in a later PR? If not, the supporter email section could be moved within the existing component and wrapped in a conditional to avoid duplication. (ie. `consentPreference` and `consentPreferences` appear in both components.)

@jamesmockett[ Here is the Figma ](https://www.figma.com/file/WOztYf2GIbvowlvJeB3vmH/Identity---MMA?node-id=207%3A5892)with the final designs.

I think it makes sense to keep them as separate components to keep the logic of which newsletters to display in each one separated. Marketing emails hard code the ids and the Support emails search to the product ids.

I can tidy it up so the css seperation lines form part of each component though.

jamesmockett (approved):
Thanks for recombining back into a single component 👍 This is looking good.

PR #770, opened 2021-12-23
Upgrade React, Emotion and Source

jamesmockett:
## What does this change?

🚨 WIP

Major version upgrades to React `17.x.x`, Emotion `11.x.x` and Source `3.x.x`.

PR #784, opened 2022-01-28
Introduce toggles to the Emails & Marketing page

vlbee:
**DO NOT MERGE**
Depends on IDAPI work to update new consents - do not merge until api work updating the Consent Model is released (https://github.com/guardian/identity/pull/2079)

Storybook has been added for easy review. 

To review locally, release build 16275 to Identity CODE and set up this project's nginx which will fallback to CODE IDAPI. You can then run MMA locally, log in via code and view changes. 

## What does this change?

Introduce toggles to the Emails & Marketing page - Supporter Exclusive section and below. 
[Figma designs (first iteration)](https://www.figma.com/file/WOztYf2GIbvowlvJeB3vmH/Identity---MMA?node-id=207%3A5892)

IDAPI Api Response model has been updated to reflect that a Consent description is not always returned. 

## How to test

New unit tests added for MarketingToggle and OptOutSection
- [x] Tested in CODE with Idapi Consent Model changes

## Images

**NEW:** 

Note: the title and description has a bit more margin between them than the UI calls for. This is because the title is in the label built into the switch which has a minimum height. It's not so noticeable when the label text wraps to two lines but I don't believe there is a way for within CSS for conditional styling if the content overflows. Any thoughts on a simple fix welcome. 
image

image

**OLD:**
image

## Accessibility


app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
```diff
@@ -12,30 +12,34 @@ interface OptOutSectionProps {
...
+ */
+const optOutFinderAndInverter =
+  (consents: ConsentOption[], clickHandler: ClickHandler) => (id: string) => {
+    const consent = consents.find((c) => c.id === id);
+    return (
+      consent && (
+        <MarketingToggle
+          id={consent.id}
+          description={consent.description}
+          selected={!consent.subscribed} // NOTE: Opt Out consent value is inverted
```
vlbee, line 29: This is the simplest solution I found. Keeps other components generic and places the logic next to the hard coded opt outs. Tests have been added to protect the logic.

app/client/components/identity/ToggleStyles.tsx
```diff
@@ -0,0 +1,137 @@
+import { css } from "@emotion/core";
+import { textSans } from "@guardian/src-foundations/typography"; // TODO v4 -> 2 change
+import { neutral, success } from "@guardian/src-foundations/palette"; // TODO v4 -> 2 change
+
+/**
+ *  TODO: this toggle stylesheet is copied from Source (Design System)
+ *  It should be updated to the Source version once MMA migration to Emotion 11 and Source v3+ is complete
+ */
```
vlbee, line 8: Should be able to use the Source toggle switch once PR #770 is merged.  Styles and component copied over with minor change - Addition of `labelId` for testing purposes.

app/package.json
```diff
@@ -115,7 +114,6 @@
     "pretty-quick": "^3.1.2",
     "raw-loader": "^4.0.2",
     "react-test-renderer": "^16.13.1",
-    "react-testing-library": "^3.1.7",
```
vlbee, line 118: outdated testing library - replaced by @testing-library/react

app/package.json
```diff
@@ -96,7 +96,6 @@
     "copy-node-modules": "^1.0.4",
     "copy-webpack-plugin": "^9.1.0",
     "core-js": "^3.19.1",
-    "dom-testing-library": "^2.3.2",
```
vlbee, line 99: outdated testing library - replaced by "@testing-library/jest-dom"

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
```diff
@@ -12,30 +12,49 @@ interface OptOutSectionProps {
...
+const optOutFinderAndInverter =
+  (consents: ConsentOption[], clickHandler: ClickHandler) =>
+  (id: string, altDescription?: string) => {
+    const consent = consents.find((c) => c.id === id);
+    return (
+      consent && (
+        <MarketingToggle
+          id={consent.id}
+          title={consent.name}
+          description={consent.description || altDescription} // Not all consents from IDAPI have a description
```
vlbee, line 30: The description field on the [Consent model is an Option](https://github.com/guardian/identity/blob/51c6c1b4db68bb4586e07383166a00202d9f5d8f/identity-model/src/main/scala/com/gu/identity/model/Consent.scala#L43), and the [api change](https://github.com/guardian/identity/pull/2079) is now returning consents without. This doesn't seem to break the TS interface and it's served to the client as an "undefined" value and handled line 71 in the `MarketingToggle` component.

But we could amend it in the ConsentOption interface [here](https://github.com/guardian/manage-frontend/blob/88b00f82ea7ee327497f93e979014412158ce85d/app/client/components/identity/models.ts#L60)

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
```diff
@@ -12,30 +12,49 @@ interface OptOutSectionProps {
...
+          onClick={clickHandler}
+        />
+      )
+    );
+  };
 
-const standardTextSize = {
-  fontSize: "17px"
-};
+const YourDataDescription: FC = () => (
```
vlbee, line 38: I'm getting a React warning in the console here: ` validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.` Any recommendations?

app/client/components/identity/ToggleSwitch.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {labelPosition === "left" && <Label />}
+      <button
+        css={[buttonStyles, getPlatformStyles(platform)]}
+        role="switch"
+        aria-checked={isChecked()}
+        aria-labelledby={labelId || "notify"}
+        onClick={onClick}
+      ></button>
+      {labelPosition === "right" && <Label />}
+    </div>
```
vlbee, line 135: Would like to introduce the alternate label position back into the Source component but hardcoded here for now.

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
coldlink: I assume that where this component is being used is under a `<p>` tag, in turn causing the `<ul>` to appear as a child of `<p>`.

Ideally you'd want to change where this component is inserted to be a `<div>`, or if that's not possible then insert this component in a different way.

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
coldlink: imo it's worth having the type/interface match the API response, and that avoids the problem of having to check that it's defined

although by changing that type, there may be other cases in the codebase where it's hasn't been handled in a way where it expects undefined values, so may need some refactoring involved

i say try changing the type and then running a type check, if theres lots of errors in the codebase and it doesn't look easy to refactor then leave the type as it was, and use the way you've handled it here.

it if doesn't look too difficult to refactor after the typecheck then it's worth doing that

app/client/__tests__/components/identity/MarketingToggle.tsx
```diff
@@ -0,0 +1,53 @@
+import { cleanup, fireEvent, render } from "@testing-library/react";
+import React from "react";
+import { create } from "react-test-renderer";
+import { MarketingToggle } from "../../../components/identity/MarketingToggle";
```
liamness, line 4: I think it would be best to stick with `@testing-library/react` instead of also using `react-test-renderer`, particularly if using the latter  with snapshot testing. Partially because it is a little confusing to mix approaches like this, but also because I have found that testing-library encourages you to write tests that are more legible in their purpose, and more resilient to future changes / refactoring.

You can use testing-library to implement snapshot testing as well, but I think it is better to write your assertions using queries. Makes it clearer which behaviours you actually care about, whereas snapshots can include things like auto-generated classnames.

app/client/__tests__/components/identity/OptOutSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    expect(screen.getByLabelText("Allow post")).toHaveAttribute(
+      "aria-checked",
+      "true"
+    );
+    expect(screen.getByLabelText("Allow phone")).toHaveAttribute(
+      "aria-checked",
+      "false"
+    );
```
liamness, line 30: https://github.com/testing-library/jest-dom#tobechecked

`@testing-library/jest-dom` should probably be set up in all these tests, currently it seems to have been manually included in a few.

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
```diff
@@ -12,30 +12,49 @@ interface OptOutSectionProps {
...
-        selected={consent.subscribed}
-        onClick={clickHandler}
-      />
-    )
-  );
-};
+/**
+ * NOTE:
+ * Only use this method for an OPT OUT consent, eg. "post_optout"
+ * The description of Opt On consents have changed so for UX/UI purposes they are now opt INs
```
liamness, line 18: I assume "Opt On" is meant to say "Opt Out" here?

app/client/components/identity/MarketingToggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+              font-weight: bold;
+            }
+          `}
+          label={title}
+          labelPosition="left"
+          labelId={id}
+          checked={!!selected}
+          onClick={(_) => {
+            return;
+          }}
```
liamness, line 68: Not sure why this is needed?

app/client/__tests__/components/identity/OptOutSection.tsx
vlbee: Yes - i wasn't sure how much to mess around with other tests. Will look at raising a seperate pr to change testing. I spent some time to try using .toBeChecked attribute but it doesn't seem to work based on the aria role=switch even though according to the docs it should. I will get it set up again and maybe you could help me debug what I was doing wrong or maybe there is a bug

app/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "trailingComma": "none"
```
vlbee, line 2: @jamesmockett There is a Prettier pre-commit hook but no prettier config file so added one as the trailing comma was giving me issues despite changing my settings in VSCode.

app/client/components/identity/MarketingToggle.tsx
vlbee: Its needed because our onClick is handled in the wrapping div but the generic checkbox component requires a function - https://github.com/guardian/manage-frontend/blob/88b00f82ea7ee327497f93e979014412158ce85d/app/client/components/checkbox.tsx#L7

I'd rather not refactor in this PR.

app/client/components/identity/EmailAndMarketing/OptOutSection.tsx
vlbee: Refactored in the API response and trickled it down the model 
```
interface ConsentAPIResponse {
  id: string;
  description?: string;
  name: string;
  isOptOut: boolean;
  isChannel: boolean;
  isProduct: boolean;
}

export interface ConsentOption {
  id: string;
  description?: string;
  frequency?: string;
  name: string;
  theme?: string;
  type: ConsentOptionType;
  subscribed: boolean;
  identityName?: string;
  isProduct?: boolean;
  isChannel?: boolean;
}
```
Undefined descriptions and other optional fields are now all handled like this where they are displayed in components: 
```
{description && getDescription(description)}
```

app/.prettierrc
jamesmockett: This has been annoying me too as it's relying on the default config that comes with Prettier. I'm not sure why it's not using the standard Guardian Prettier and ESLint configs, but I'm looking at adding those and cleaning up the entire codebase so files don't get reformatted every time you touch something.

app/client/__tests__/components/identity/MarketingToggle.tsx
vlbee: Snapshot tests have been removed in favour of queries for this and the MarketingCheckbox

app/client/__tests__/components/identity/OptOutSection.tsx
vlbee: I had a  look and setting this up in a jest config file and it broke some global imports so would prefer to look at in a separate PR so this one doesn't get unwieldy.

app/client/__tests__/components/identity/MarketingCheckbox.tsx
```diff
@@ -0,0 +1,53 @@
+import { cleanup, render, fireEvent, screen } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import React from "react";
+import { MarketingCheckbox } from "../../../components/identity/MarketingCheckbox";
+
+afterEach(cleanup);
+
+describe.only("MarketingCheckbox component", () => {
```
liamness, line 8: `.only` is still used here. Maybe should add some linting for that? (in a follow up PR).

vlbee: oops. fixed. Yes, will look at adding now in a new pr

vlbee:
As this PR can't be merged until the Gateway ui work is ready, I will look at adding in some linting checks and updated the identity component tests in the meantime.

vlbee:
Updated Copy 
image

vlbee:
UX feedback TODO: 
1/ move bullet list over to left
2/ spacing between label and description
3/ import new toggle styles - spacing around the toggle button is off

PR #785, opened 2022-02-04
Populate Storybook with first batch of generic components

jamesmockett:
## What does this change?

This continues the task of building out Storybook by adding stories for some of the app's generic components. There are no functional changes to components in this PR other than adding an `export` to the props interface so that it can be imported by the corresponding story. (The diff for some components is larger due to Prettier reformatting.)

## Images

Screenshot 2022-02-04 at 10 09 21


jamesmockett:
[TASK: Add generic components to Storybook (Part 1)](https://trello.com/c/5IolYDAm/2826-task-add-generic-components-to-storybook-part-1)

rBangay (approved):
Nice work 👍

PR #786, opened 2022-02-04
Add cloudformation template for GHA user

kelvin-chappell:
Now has permissions to discover API gateways.

### DIFF
--- a/gha-user.cfn.yaml
+++ b/gha-user.cfn.yaml
@@ -1,0 +1,90 @@
+# Stack: manage-frontend-github-actions-user
+# ==========================================
+
+Resources:
+
+  GetObjectsDE66E0D1:
+    Type: AWS::IAM::Policy
+    Properties:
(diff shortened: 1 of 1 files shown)


jonflynng (approved):

twrichards:
My only worry with this is whether people could fork the repo, then write something malicious which invokes the various API Gateway lambdas. Since we wrote something yesterday in the codebase to skip the identity piece, could we adjust the codebase again to prevent the attempt to discover the API Gateways, since their responses are mocked anyway in the cypress tests AFAIK?

twrichards (changes requested):
My only worry with this is whether people could fork the repo, then write something malicious which invokes the various API Gateway lambdas. Since we wrote something yesterday in the codebase to skip the identity piece, could we adjust the codebase again to prevent the attempt to discover the API Gateways, since their responses are mocked anyway in the cypress tests AFAIK?

PR #790, opened 2022-02-10
Set up cypress

jonflynng:
**Why Cypress and how does it benefit us - will add these points and further detail to the wiki**

- We do a lot of unit testing in software, running tests against individual functions in our code. Cypress brings the whole app together, runs tests on the final compiled app, tests it in the environment the user will actually use (i.e.: a web browser), and tests it with real user journeys.
- Application state (cookies, network responses, react state, etc…) is not tested with Cypress. We only test what has been printed on the screen.
- unit testing for components less needed, much more difficult to maintain. no need to mock complex router state.
- Great way to document how user journeys work, i.e.: what buttons are clicked, which API endpoints are called and in what order
- Can check how many times API endpoints are called at the end of each test run/user journey. We can avoid backend and react render errors with this.

**NOTE:**

I would use API response objects from our fixtures folder, however, importing anything outside the cypress creates the following error: https://github.com/cypress-io/cypress/issues/18435.   Hopefully will be solved soon. As of now, I've created a fixtures folder within the cypress folder.

I've excluded the Cypress folder within the tsconfig in /app. It was running the typescript compiler over tests within cypress and causing errors. Cypress uses it's own tsconfig located within the cypress folder.

**Current state of mocking**

As of now, all of our backends are being mocked, including identity. In the future, we could possibly have one test suite running with mocked backends, and the other a complete end to end test with no mocking, I assume this will require setting up NGINX on github actions to properly proxy requests. I'll add this to the wiki.


app/server/routes/frontendCommon.ts
```diff
@@ -17,8 +14,14 @@ if (conf.ENVIRONMENT === Environments.PRODUCTION && !conf.CLIENT_DSN) {
 
 export const getRecaptchaPublicKey = async () => {
   try {
-    const recaptchaConfig = await recaptchaConfigPromise;
-    const recaptchaPublicKey = recaptchaConfig?.publicKey;
+    let recaptchaPublicKey: string | undefined;
+
+    if (conf.ENVIRONMENT === Environments.DEVELOPMENT) {
+      recaptchaPublicKey = "6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI";
```
jonflynng, line 20: I've made it so test recaptcha is always used in development environment, assume this is ok?

app/server/html.ts
```diff
@@ -27,13 +27,17 @@ const html: (_: {
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1">
       <title>${title}</title>
+
       ${insertGlobals(globals)}
+      <script src="https://js.stripe.com/v3/"></script>
```
jamesmockett, line 32: Does this mean Stripe is loaded on every page rather than just the card payment screen?

twrichards: good spot. i feel sure there's privacy requirement not to do this. This is why stripe is loading asynchronously using `loadStripe` function, only when it's needed https://github.com/guardian/manage-frontend/blob/e2efc7219fead3918a176efd1089cdf7b61899c9/app/client/stripe.ts#L33

app/server/html.ts
```diff
@@ -16,6 +16,14 @@ const insertGlobals = (globals: Globals) => {
   </script>`;
 };
 
+const insertStripe = () => {
```
jonflynng, line 19: @jamesmockett @twrichards     preloading stripe when running cypress

app/server/html.ts
twrichards: should be fine to load it on every page in cypress, but I'm curious why 

also as general rule one should try to keep changes to production code (particularly branching logic) for the sole purpose of testing - to a minimum

app/cypress/support/commands.ts
```diff
@@ -0,0 +1,139 @@
...
+    value: string;
+  }
+];
+export const randomMailosaurEmail = () => {
+  return uuidv4() + "@" + Cypress.env("MAILOSAUR_SERVER_ID") + ".mailosaur.net";
+};
+
+export const randomPassword = () => uuidv4();
+
+export const createTestUser = ({
```
jonflynng, line 96: Keeping this in for future when we run tests without mocked backends

app/server/html.ts
jonflynng: thanks for the suggestion richard, works nicely

app/webpack.common.js
```diff
@@ -17,66 +28,69 @@ const definePlugin = new webpack.DefinePlugin({
     : `'DEV_${new Date().getTime()}'`,
   GIT_COMMIT_HASH: process.env.BUILD_VCS_NUMBER
     ? `'${process.env.BUILD_VCS_NUMBER}'`
-    : `'${new GitRevisionPlugin().commithash()}'`
+    : `'${new GitRevisionPlugin().commithash()}'`,
+  "process.env.CYPRESS": process.env.CYPRESS
+    ? JSON.stringify(process.env.CYPRESS)
+    : undefined,
```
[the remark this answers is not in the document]
jonflynng: just updated this

app/server/routes/frontendCommon.ts
rBangay: Assuming that is the public recaptcha development key:

https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do

👍

rBangay (approved):

jamesmockett (approved):
Nice work 👍 It's great to have this in place.

jonflynng:
[TASK: Set up Cypress](https://trello.com/c/D0KLa4VO/2816-task-set-up-cypress)

PR #791, opened 2022-02-10
Clean up test data

jamesmockett:
## What does this change?

This removes `testData.ts` and updates unit tests to use test data from the `/fixtures` folder which is shared with Storybook. In addition, the test data has been cleaned up to remove real email and postal addresses.


jonflynng (approved):
good stuff, I'll merge the cypress test data into this as soon as the dynamic imports are fixed with it.

jonflynng (approved):

PR #792, opened 2022-02-11
[RB] - Add currency and currencyISO to invoice table

rBangay:
## What does this change?

Add currency and currencyISO properties to the defualt returned object in the getPlan function. This targets the gift subscriptions and allows the correct data to be displayed in the invoice table on the billing page

## Images

Before             |  After
  |

### DIFF
--- a/app/client/components/billing/billing.tsx
+++ b/app/client/components/billing/billing.tsx
@@ -118,8 +118,6 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                   const paidPlan = getMainPlan(
                     productDetail.subscription
                   ) as PaidSubscriptionPlan;
-                  console.log("paidPlan = ");
-                  console.log(paidPlan);
                   const productInvoiceData = invoiceData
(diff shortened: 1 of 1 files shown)


app/client/components/billing/billing.tsx
```diff
@@ -119,18 +118,20 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
                   const paidPlan = getMainPlan(
                     productDetail.subscription
                   ) as PaidSubscriptionPlan;
+                  console.log("paidPlan = ");
+                  console.log(paidPlan);
```
kelvin-chappell, line 122: Do you want to leave these here?

kelvin-chappell (approved):
Nice solution! 👍

app/client/components/billing/billing.tsx
rBangay: ooo no :s

kelvin-chappell (approved):
👍

PR #794, opened 2022-02-16
Exclude Dependabot PRs from Chromatic and Cypress actions

jamesmockett:
## What does this change?

The Chromatic and Cypress checks fail on PRs raised by Dependabot. This is due to [the actions not having access to GitHub secrets](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#accessing-secrets) so they can't be run. As Dependabot PRs are raised relatively infrequently, tend not to cause UI regressions, and the checks are still run when merging to `main` we felt that it was safe to exclude Dependabot PRs from these actions for now rather than duplicating the secrets for Dependabot. (We may revisit this at a later date.)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,5 +1,8 @@
 name: Chromatic
-on: push
+on:
+  push:
+    branches-ignore:
+      - "dependabot/**"
 
 jobs:
   chromatic:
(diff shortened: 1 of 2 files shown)


jamesmockett:
[Make Chromatic and Cypress actions work with Dependabot PRs](https://trello.com/c/14YWJu7o/2919-make-chromatic-and-cypress-actions-work-with-dependabot-prs)

kelvin-chappell (approved):
Seems sensible 👍

PR #796, opened 2022-02-22
Upgrade to React 16.14.0 and new JSX transform

jamesmockett:
## Accessibility


jamesmockett:
[SPIKE: Upgrade to React v16.14.0 and use new JSX transform](https://trello.com/c/aJ8BqKwh/2934-spike-upgrade-to-react-v16140-and-use-new-jsx-transform)

app/.storybook/.babelrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
jamesmockett, line 1: I've converted this to `.js` so I could copy from the Webpack config without converting to JSON. In a separate PR we could extract this to a common config as it's now shared across Webpack, Jest and Storybook.

app/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     // "lib": [],                             /* Specify library files to be included in the compilation. */
     // "allowJs": true,                       /* Allow javascript files to be compiled. */
     // "checkJs": true,                       /* Report errors in .js files. */
-    "jsx": "react" /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */,
+    "jsx": "preserve" /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */,
```
jamesmockett, line 10: There's no need to explicitly import `react` to use JSX when using the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). Unfortunately we're still using some older packages that _do_ import `react` leading to type check errors.

As the transformation of JSX to JS is handled by Babel we can skip this step here as the TypeScript compiler is only used for type checking.

rBangay: good spot

rBangay (dismissed):

PR #799, opened 2022-03-01
Emotion 11 and Source upgrade

jamesmockett:
## What does this change?

This follows on from #796 (which updated React to 16.14.0 and introduced the new JSX transform) and updates the codebase to use Emotion 11 and newer versions of the Source packages. (This is still a draft so it's undecided if we leave the Source packages at 3.x and update to 4.x in a separate PR to keep things incremental and more manageable, or do it all in one go in this PR.)

Although this touches a large number of files it's mostly renaming `@emotion/core` imports to `@emotion/react`. There are also some changes to how styles are defined in the Identity components to wrap object styles in Emotion's `css` function and fix how these are composed together.

## How to test

As this touches all of the UI in the app it would be prudent to regression test key pages and journeys through the app. (This is already partly automated with visual regression testing via Chromatic.)

There is no change in functionality so we should not expect to see any additional client-side errors logged in Sentry. (Other than the usual transient chunk loading errors following deployment.)


jamesmockett:
[Upgrade to Source V4](https://trello.com/c/HJotfGOf/2746-upgrade-to-source-v4)

jamesmockett:
[[M] Upgrade Emotion to v11](https://trello.com/c/z2TCE3iS/2103-m-upgrade-emotion-to-v11)

jamesmockett:
This was put together before the React Router upgrade (#812) which has touched much of the codebase. Rather than attempting to unpick this branch and fix all of the conflicts I've created a new PR: #818

PR #806, opened 2022-03-18
Do not render theme dropdown if there are no newsletters

georgeblahblah:
## What does this change?

It is possible for newsletters to be marked as inactive, and therefore hidden from MMA. If there are no newsletters visible for a given newsletter category, we should no longer display that theme dropdown as it can be confusing for users.

* Do not render theme dropdown if there are no newsletters for that theme

An alternative approach would be to render a message within the theme dropdown instead of hiding the dropdown

## How to test

* Verify that there all categories have newsletters, so that there are no empty categories

### Before

* Both the Comment and From the Papers sections are empty yet still appear

Screenshot 2022-03-18 at 16 10 19

### After

* The empty sections no longer appear

Screenshot 2022-03-18 at 16 07 40

### DIFF
--- a/app/client/components/identity/EmailAndMarketing/NewsletterSection.tsx
+++ b/app/client/components/identity/EmailAndMarketing/NewsletterSection.tsx
@@ -60,13 +60,18 @@ const newsletterPreferenceGroups = (
 		Theme.work,
 		Theme.FromThePapers,
 	];
-	return themes.map((theme) => (
-		<DropMenu key={theme} color={colors[theme]} title={theme}>
-			{newsletters
-				.filter((n) => n.theme === theme)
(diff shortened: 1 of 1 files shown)


alinaboghiu (approved):

georgeblahblah:
[MMA page categories/themes don't align with the newsletters page](https://trello.com/c/mYpTugUy/619-mma-page-categories-themes-dont-align-with-the-newsletters-page)

georgeblahblah:
I reverted this to a Draft PR because we noticed some further details about the bug which should be fixed as part of this PR. There is more detail in [the trello ticket](https://trello.com/c/mYpTugUy/619-mma-page-categories-themes-dont-align-with-the-newsletters-page)

PR #808, opened 2022-04-06
Add Cypress tests for holiday stops journey

jamesmockett:
## What does this change?

Extends end to end test coverage of the holidays stops journey to give increased confidence as we refactor it as part of the React Router upgrade.

Fixtures for the Cypress tests have been consolidated with those in `client/fixtures` so there is a single set used for all tests.


jamesmockett:
[Improve holidays stops journey test coverage](https://trello.com/c/UebDQkgW/3074-improve-holidays-stops-journey-test-coverage)

app/cypress.json
```diff
@@ -19,5 +19,6 @@
     "*the-ozone-project.com",
     "*openx.net"
   ],
+  "fixturesFolder": false,
```
jonflynng, line 22: What does this do?

jonflynng (approved):
Looks good and passes locally. How did you get the fixtures folder working? As when I moved it outside the cypress folder I got webpack errors (https://github.com/guardian/manage-frontend/pull/790). Just a small improvement, if you could check that certain endpoints were called the correct number of times at the end of each test would be good :) like so: https://github.com/guardian/manage-frontend/blob/main/app/cypress/integration/parallel-2/contributions.spec.ts#L116-L117

app/cypress.json
rBangay: It should stop cypress from creating a fixtures directory inside the cypress directory as our fixtures now live outside of cypress

app/cypress/integration/parallel-3/holidayStops.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+		cy.findByText('Review details').click();
+
+		cy.get('table').contains('9 February - 11 February 2022');
+		cy.get('table').contains('1 issue');
+		cy.get('table').contains('$8.13 off your 1 February 2023 payment');
+
+		cy.findByText('Confirm').click();
+
+		cy.wait('@create_holiday_stop');
+		cy.findByText('Your schedule has been set').should('exist');
```
jonflynng, line 96: is the .should('exist') necessary here?

rBangay:
> Looks good and passes locally. How did you get the fixtures folder working? As when I moved it outside the cypress folder I got webpack errors (#790). Just a small improvement, if you could check that certain endpoints were called the correct number of times at the end of each test would be good :) like so: https://github.com/guardian/manage-frontend/blob/main/app/cypress/integration/parallel-2/contributions.spec.ts#L116-L117

Hey, so moving the fixtures outside of the cypress directory just seemed to work without us having to do anything other than update the import paths. I'm not sure why it wasn't working previously and is now.

I've added the api response n calls expectations to the holiday tests

jonflynng (approved):
nice

PR #809, opened 2022-04-08
Clean up Cypress specs

jamesmockett:
## What does this change?

- Moves authentication and cookie banner set up to separate module to avoid duplication in specs
- Removes unnecessary `setLocalBaseUrl` method (this is already defined in `cypress.json`)
- Moves IAB consent management framework exception handling to global test set up in `index.ts` to avoid duplicating in specs


jonflynng (approved):
thanks, looks good

PR #813, opened 2022-05-18
Upgrade to React 17

jamesmockett:
## What does this change?

Updates the app to React 17. This is a relatively small change as a previous update to React 16.14.0 (#796) has already tackled rolling out the updated JSX transform.

All existing Enzyme tests have been migrated to React Testing Library. Enzyme is no longer actively maintained and there is no official adapter for React 17. Whilst there is an unofficial adapter that we could use in the interim, the author has stated that they have [no plans to support React 18](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl) so we would need to do this eventually in any case.

_This PR is dependent on #812_

## How to test

This PR updates the Stripe React packages and Formik so particularly attention should be paid to Billing, and the Profile and Settings pages. The core journeys through the app are covered by end-to-end tests, but it's still worthwhile running through these manually.


jamesmockett:
[[M] Upgrade React to latest version that's compatible with Source](https://trello.com/c/AExPrmuY/2102-m-upgrade-react-to-latest-version-thats-compatible-with-source)

rBangay (dismissed):
👍

rBangay (approved):

PR #816, opened 2022-05-25
[RB] - add optional chaining operator

rBangay:
## What does this change?
Add optional chaining operator to the router state selected reason check in the cancellation review stage. In order to catch the possibility that the router state itself is null and throwing an error trying to access a property of null


jamesmockett (approved):
👍

PR #818, opened 2022-05-27
Migrate to Source 4

jamesmockett:
## What does this change?

Migrates the application to Source 4 and Emotion 11.

Although this touches a large number of files the bulk of the changes are renaming imports for Source and Emotion, and replacing deprecated colour tokens. There are also some changes to how styles are defined and composed in the Identity components. (Mainly to apply Emotion's `css` function to object styles.)

### Custom radio button styles

Screenshot 2022-06-06 at 17 27 45

The custom radio buttons used on the _payment update_ page have been rebuilt so that they no longer rely upon overriding Source's own radio button styles. This makes them more robust and less liable to break when upgrading to newer versions of Source. The `<RadioGroup>` component is now used to group the radio buttons and provide a hidden label to improve accessibility. There are further improvements we could make to these, but for the purposes of this PR the aim is to replicate what is already in place.

## How to test

As this touches all of the UI in the app it would be prudent to regression test key pages and journeys through the app. (This is already partly automated with visual regression testing via Chromatic.)

There is no change in functionality so we should not expect to see any additional client-side errors logged in Sentry.


jamesmockett:
[Upgrade Emotion and Source](https://trello.com/c/z2TCE3iS/2103-upgrade-emotion-and-source)

jamesmockett:
[Upgrade Emotion and Source](https://trello.com/c/z2TCE3iS/2103-upgrade-emotion-and-source)

rBangay (approved):
👍

vlbee (approved):

jonflynng (approved):
Not looked through everything in-depth but happy if the tests are passing. As we can't run tests on IE11 through Cypress we'll have to always check via browserstack when upgrading or refactoring packages.

 just adding this here for future reference: https://guardian.github.io/source/?path=/story/release-notes-source-v4--page

PR #821, opened 2022-06-01
check that ALL api gateway abstractions have their host & api key in the healthcheck

twrichards:
FIRST IDEA ([of 2](https://github.com/guardian/manage-frontend/pull/822)) to solve the slight race condition which exists now where the healthcheck's loading of the apikey and host for the `holiday-stop-api` succeeds but the real one used for the /api/ endpoints fails (due to AWS rate limiting, likely due Prism - see #559 )


jamesmockett (approved):
Thanks Tom, this looks great 👍 We've decided to go with this option for now, but may revisit the 2nd option and look at adding a retry mechanism when fetching credentials.

PR #826, opened 2022-06-09
Product switching journey 'review' page

jamesmockett:
## What does this change?

Adds components and styles for product switching 'review' page. The switching journey sits behind a feature switch so this page is not accessible to users.

The page includes two new components that could be extracted as generic components for use elsewhere: `<Card>` and `<KeyValueTable>`. I've left these in place for now as the switching journey is still under development. (And we may want to discuss naming and where they should live.)

manage thegulocal com_cancel_contributions_switch (1)


jamesmockett:
[create the visual components for the new cancellation switching journey - page 2](https://trello.com/c/lqxZEeTY/3219-create-the-visual-components-for-the-new-cancellation-switching-journey-page-2)

rBangay (approved):

PR #835, opened 2022-06-24
Disable cmp via cookie in Cypress tests

vlbee:
## What does this change?

The cmp banner can be disabled via a cookie. This is convenient for integration testing with Cypress. 

## Accessibility


rBangay (approved):
👍 much nicer

PR #837, opened 2022-07-01
[RB] - Remove subscription.plan object

rBangay:
## What does this change?
Use the currentPlans and futurePlans objects instead of the plan object as part of the product detail member data api response. 

Wrap usage of these (currentPlans and futurePlans) with a sentry alert if the response from the api doesn't include them for some reason and add early return of fallback error message component where applicable to stop potential run time error.

#### Relevant members data api pr:
https://github.com/guardian/members-data-api/pull/371

### DIFF
--- a/app/client/components/accountoverview/accountOverviewCard.tsx
+++ b/app/client/components/accountoverview/accountOverviewCard.tsx
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/browser';
 import { css } from '@emotion/react';
 import {
 	space,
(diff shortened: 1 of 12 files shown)


app/client/components/accountoverview/accountOverviewCard.tsx
```diff
@@ -34,6 +34,9 @@ interface AccountOverviewCardProps {
 
 export const AccountOverviewCard = (props: AccountOverviewCardProps) => {
 	const mainPlan = getMainPlan(props.productDetail.subscription);
+	if (!mainPlan) {
+		throw 'mainPlan does not exist in accountOverviewCard';
```
jamesmockett, line 38: The `<ErrorBoundary>` component is expecting an error object so you'll need to wrap the string in `new Error()` otherwise it gets logged as `undefined: undefined`.  (Internally it's returning `${error.name}: ${error.message}`.) I've pulled down the branch and double-checked as I wasn't sure if that was the case 😅 

```suggestion
		throw new Error('mainPlan does not exist in accountOverviewCard');
```

jamesmockett (approved):
👍

PR #838, opened 2022-07-01
Add Guardian Patrons into manage

rupertbates:
## What does this change?

We want Guardian Patrons (non-member patrons acquired through patrons.theguardian.com) to be able to view their account details through manage in the same way as customers of all other products.

This PR follows on from https://github.com/guardian/members-data-api/pull/752 which adds Guardian Patrons into the /mma endpoint of mdapi. 

The majority of the code changes are just to hide the 'Manage payment method' button for Guardian Patrons because the payment method update code in members data api is very coupled to Zuora so will need to be rewritten to support Patrons which is outside the scope of this work. If there is an easier way to do this, then do let me know!

## How to test

The only way to create a test patron account is via the Stripe dashboard unfortunately, I've attached a screenshot of what they look like though.

## Images

Screen Shot 2022-07-07 at 16 40 59
Screen Shot 2022-07-08 at 12 13 28
Screen Shot 2022-07-08 at 12 13 50

## Accessibility

- [x] [The change doesn't use only colour to convey meaning](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#use-of-colour)


app/client/components/accountoverview/accountOverviewCard.tsx
```diff
@@ -97,6 +100,8 @@ export const AccountOverviewCard = (props: AccountOverviewCardProps) => {
 		display: inline-block;
 		vertical-align: top;
 		width: calc(100% - 15ch);
+		overflow: hidden;
```
rupertbates, line 103: Subscription ids are much longer for Stripe accounts than they are for Zuora accounts, this hides and ellipsises it (see screenshot)

rBangay (approved):
👍

PR #840, opened 2022-07-06
[RB] - consolidate the space at the bottom of the page so that it is …

rBangay:
## What does this change?
Attempt to make the spacing at the bottom of all pages in manage consistent.

## Images
Before             |  After
  |


app/client/components/delivery/address/deliveryAddressConfirmation.tsx
```diff
@@ -227,7 +227,10 @@ const AddressConfirmation = (props: ProductType) => {
 					<p
 						css={css`
 							${textSans.medium()};
-							margin-top: ${space[12]}px;
+							margin: ${space[12]}px 0
+								${showTopCallCentreNumbers
+									? `${space[5]}px`
+									: '0'};
```
jamesmockett, line 233: You could replace the logic for the bottom margin with a `<Stack>` wrapped around the paragraph and `<CallCentreEmailAndNumbers />` component? As it only applies spacing between adjacent elements it won't add any extra spacing if the call centre component isn't rendered. (Sorry, I've turned into a `<Stack>` cheerleader now 😀)

rBangay: 👍  yep, good idea

jamesmockett (approved):
Nice use of the `<Stack>` component 😍  Removing a bunch of custom spacing rules makes things much easier to follow.

PR #845, opened 2022-07-14
Remove product switch ab

rBangay:
## What does this change?
Remove the cancellation product switch a/b test while keeping the a/b test infrastructure.

Implement a simple feature switch mechanism (`/shared/featureSwitches.ts`) which can be overridden via a urlParam which should allow us to deploy a feature switched off in any environment and allow people to test that feature by loading the accountoverview with the corresponding feature url param.


jamesmockett (approved):
It's really nice being able to override feature switches with a query string for testing 👍

PR #849, opened 2022-07-20
Add `<Heading>` component and update cancellation heading styles

jamesmockett:
## What does this change?

We use the same heading style across `manage-frontend`, but the CSS is duplicated in multiple places and implementations vary so there are subtle variations across the site. (Particularly with regards to responsive behaviour.)

This PR introduces a new `<Heading>` component in an attempt to standardise headings across the site and avoid duplication. It's fairly barebones for now so will likely be expanded to add additional options as it is used more widely.

The component includes two props: `level` and `cssOverrides`. `level` can be used to set the desired heading level. (If not specified it defaults to a `<h2>` element.) `cssOverrides` follows the Source convention and allows styling overrides to be applied to the heading element directly.

Headings do not include any external spacing. Instead, layout components such as `<Stack>` should be used to space out headings and adjacent elements. As this isn't always straightforward when using the component within existing layouts the `cssOverrides` prop can be used to apply spacing where needed. `cssOverrides` is also used to restrict the maximum width of headings on the cancellation pages by applying the `measure.heading` style.

The contribution cancellation flow has been updated to use the `<Heading>` component so that the styling matches the new product switch pages. (As well as other pages on the site.)

The diff is best viewed with _Hide whitespace_ enabled due to indentation changes in the existing product switch pages which makes it look like there are more changes than there really are.

Screenshot 2022-07-20 at 18 02 32

### Existing style

Screenshot 2022-07-28 at 10 22 16

### Updated style with feature switch off

Screenshot 2022-07-20 at 17 36 46

### Updated style with feature switch on

If coming from the product switch offer page different content is shown to avoid repetition.

Screenshot 2022-07-20 at 17 37 26

### Updated, but not dependent on feature switch

For subsequent steps in the flow only the heading style has changed. Content remains the same.

Screenshot 2022-07-20 at 17 37 45

Screenshot 2022-07-20 at 17 38 11

### Other product types

Screenshot 2022-07-29 at 10 40 26

### DIFF
--- a/app/client/components/cancel/CancellationReasonReview.tsx
+++ b/app/client/components/cancel/CancellationReasonReview.tsx
@@ -45,6 +45,7 @@ import {
 import { Spinner } from '../spinner';
 import { GenericErrorScreen } from '../genericErrorScreen';
 import { Heading } from '../Heading';
+import { measure } from '../../styles/typography';
 
 const getPatchUpdateCaseFunc =
 	(isTestUser: boolean, caseId: string, feedback: string) => async () =>
@@ -434,10 +435,12 @@ const CancellationReasonReview = () => {
 				) : (
 					<>
 						<Heading
-							measure="medium"
-							cssOverrides={css`
-								margin-bottom: ${space[6]}px;
-							`}
+							cssOverrides={[
+								measure.heading,
+								css`
+									margin-bottom: ${space[6]}px;
+								`,
+							]}
 						>
 							{productType.cancellation.hideReasonTitlePrefix
 								? ''
--- a/app/client/components/cancel/CancellationReasonReview.tsx
+++ b/app/client/components/cancel/CancellationReasonReview.tsx
@@ -44,7 +44,7 @@ import {
 } from '../delivery/records/deliveryRecordsApi';
 import { Spinner } from '../spinner';
 import { GenericErrorScreen } from '../genericErrorScreen';
-import { heading, measure } from '../../styles/typography';
+import { Heading } from '../Heading';
 
 const getPatchUpdateCaseFunc =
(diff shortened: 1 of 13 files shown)


app/client/components/cancel/CancellationReasonReview.tsx
```diff
@@ -432,12 +433,17 @@ const CancellationReasonReview = () => {
 					)
 				) : (
 					<>
-						<h3 id="save_title">
```
jamesmockett, line 435: Is this ID used for anything? I can't find any references to it elsewhere in the codebase. If it turns out it _is_ doing something then the `<Heading>` component will probably need an optional `id` prop.

rBangay: I wonder if it was put in there as a hook for a test. I can't find any other reference to this in the codebase either so I would say it's safe to bin it off 👍

app/client/components/productSwitch/CancellationSwitchConfirmed.tsx
```diff
@@ -55,226 +56,224 @@ const CancellationSwitchConfirmed = () => {
 	const newProduct = productSwitchConfirmationInfo.newProduct;
 
 	return (
-		<>
-			<h2 css={headingCss}>Your {newProduct.name} is now active</h2>
-			<Stack space={9}>
-				<p css={standfirstCss}>
+		<Stack space={9} cssOverrides={pageTopCss}>
```
jamesmockett, line 59: This additional CSS is needed to prevent the content sitting flush with the main page header. Ideally the spacing around the content would be defined as part of the overall page layout (like how we add consistent spacing to the bottom of pages), but this would require changes to a number of pages so is outside the scope of this PR.

rBangay (approved):
👍

PR #852, opened 2022-07-29
Use Source media query API

jamesmockett:
## What does this change?

Uses the Source media query API in preference to our own breakpoint definitions and removes `breakpoints.ts` completely.

## How to test

The application is partially covered by visual regression testing, but it would be prudent to manually test some key pages at different viewport widths to ensure no regressions.


jamesmockett:
[Make all breakpoints in Manage Frontend use Source v5](https://trello.com/c/qvlesFUb/3343-make-all-breakpoints-in-manage-frontend-use-source-v5)

app/client/components/holiday/holidaySelectionInfo.tsx
```diff
@@ -71,15 +70,15 @@ export const HolidaySelectionInfo = (props: HolidaySelectionInfoPros) => {
 				</div>
 				<div
 					css={{
-						[queries.maxHeight(600)]: {
+						'@media(max-height: 600px)': {
```
jamesmockett, line 73: This is the only instance of a vertical media query in the app, but as there's no support in Source and it's not based on any of the breakpoint definitions it's been replaced with a regular media query.

rBangay (approved):
👍

PR #850, opened 2022-07-22
Add Supporter Plus product

rupertbates:
## What does this change?
This PR adds the new Supporter Plus product into manage. Depends on [this pr in members-data-api](https://github.com/guardian/members-data-api/pull/778)

## How to test

Create an account with a Supporter Plus subscription and sign in to manage to check that it appears correctly

## Images

Screen Shot 2022-07-22 at 09 58 18

PR #854, opened 2022-08-05
Self-service cancellation for Supporter Plus

jamesmockett:
## What does this change?

Adds cancellation flow for Supporter Plus (based on existing digital subscription flow).

The new product type added in #850 has been updated to disabled the Thrasher shown on the cancellation confirmation page. (`shouldHideThrasher: true`);

_TODO_: Copy for the cancellation flow will be updating in a subsequent PR.

## Images

With Thrasher enabled:

Screenshot 2022-08-05 at 16 57 38

With Thrasher hidden:

Screenshot 2022-08-05 at 16 59 09


jamesmockett:
[Cancellations for new prop must be available via self-service](https://trello.com/c/84byB0z1/3355-cancellations-for-new-prop-must-be-available-via-self-service)

rBangay (approved):
👍

PR #856, opened 2022-08-16
Migrate CI from TeamCity to GitHub Actions

jamesmockett:
🚨 Before merging update the `LAST_TEAMCITY_BUILD` var and remove the `dryRun` flag 🚨

## What does this change?

Migrates all CI tasks from TeamCity to GitHub Actions.

Currently we use GitHub Actions to run Chromatic and Cypress, but everything else (unit testing, bundling and uploading to Riff Raff) happens on TeamCity. By migrating to GitHub Actions we benefit from having all of our build config in a single place and under version control, and build logs accessible from the GitHub UI. (Removing the friction of having to connect to the VPN and log in to TeamCity.)

Builds should be visible in Riff Raff and deployable.


jamesmockett:
[Migrate from TeamCity to GitHub Actions](https://trello.com/c/i9HojtqU/3420-migrate-from-teamcity-to-github-actions)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Compress for upload
+        run: |
+          cd dist
+          zip -r ../manage-frontend.zip .
+        working-directory: app
+
+      - name: Upload to Riff-Raff
+        uses: guardian/actions-riff-raff@v1
+        with:
+          dryRun: true
```
nicl, line 72: (I imagine this will be set to `false`/removed on merge.)

jamesmockett: I've left this in whilst TeamCity is still in place so we don't accidentally upload a build and mess up the build numbers, but there's a note on the PR to remove this before merging. (As well as updating `LAST_TEAMCITY_BUILD`.)

PR #858, opened 2022-08-22
[RB] - update node to lts version (16)

rBangay:
## What does this change?
Downgrade the node version running manage from 17 to 16 (which is the current LTS version).

This is so that other dependencies like the CDK migration work can progress as they have a dependency on LTS versions of node.


jamesmockett (approved):
👍

PR #859, opened 2022-08-23
Improve test suite reliability

jamesmockett:
## What does this change?

Improves the reliability of the test suite by fixing issues in the `dates` and `identityMiddleware` specs. These came to light whilst migrating from TeamCity to GitHub Actions (#856).

We've previously fixed issues with tests involving dates behaving differently in dev and CI in #744, but the `dateRange` test didn't need updating. However, this does fail when run in GitHub Actions due to timezone differences so has been updated to use the existing `localiseDate` function.

The `identityMiddleware` has been generating warnings for a while, but we've never managed to get to the bottom of the issues. The warnings don't cause the test suite to fail when run locally or in TeamCity, but do when run in GitHub Actions. The source of the warnings was tracked down to the `idapiConfig` module which tries to access AWS to fetch config and also sets up a promise handler which resolves _after_ the tests have finished. This module isn't used by the code that's being tested so `identityMiddleware` has been broken out into two separate modules so we can test `augmentRedirectURL` without including this additional code.

### DIFF
--- a/app/client/__tests__/dates.test.ts
+++ b/app/client/__tests__/dates.test.ts
@@ -253,11 +253,9 @@ describe('dateRange', () => {
 		const inputStartDate = '1969-07-16';
 		const inputEndDate = '1969-08-16';
 		const dateRangeObj = dateRange(inputStartDate, inputEndDate);
-		expect(
-			dateRangeObj.start.toString() + dateRangeObj.end.toString(),
-		).toEqual(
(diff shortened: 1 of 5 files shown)


jamesmockett:
[Improve test suite reliability](https://trello.com/c/07HRnEKJ/3431-improve-test-suite-reliability)

rBangay (approved):
👍

PR #861, opened 2022-09-01
[RB] - update dev sandbox livechat url config

rBangay:
## What does this change?
Updates the salesforce live chat config urls without which the non prod environments couldn't load the live chat window.


jamesmockett (approved):
👍

PR #864, opened 2022-09-06
Relocate application code from `app` folder

jamesmockett:
Co-authored-by: Richard Bangay <richard.bangay@gmail.com>

## What does this change?
Removes the app directory and moves all its contents into the root of the project, which makes integration with tooling easier.

## How to test
Deployment to the CODE environment tests both the client side bundling and deployment processes still work.


jamesmockett:
[[M] SPIKE moving application code from `/app` folder to root and Move project config files into root path](https://trello.com/c/2y1cm8az/3452-m-spike-moving-application-code-from-app-folder-to-root-and-move-project-config-files-into-root-path)

KentGrm (approved):

akash1810:
Wonderful! Not a blocker here, but you might want to consider using [workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/). The biggest benefit with workspaces is the sharing of config and dependencies.

PR #865, opened 2022-09-07
Remove unnecessary environment variables from CloudFormation config

rBangay:
## What does this change?

Removes environment variables defining application and API domains as these don't need to be part of the CloudFormation config and can be defined directly in the application code. (Fallbacks for dev environment domains are already defined in the code so this also cleans up some duplication.) `NODE_ENV` has also been removed from CloudFormation as this is always set to the same value, and `BUILD` has been removed from the app config as it's unused.

### DIFF
--- a/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
+++ b/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
@@ -34,17 +34,13 @@ Object {
     },
     "StageVariables": Object {
       "CODE": Object {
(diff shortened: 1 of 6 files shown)


jamesmockett:
[Remove unnecessary environment variables from CloudFormation config](https://trello.com/c/VkM1WNRA/3459-remove-unnecessary-environment-variables-from-cloudformation-config)

akash1810 (approved):

PR #868, opened 2022-09-09
Remove SonarQube integration

jamesmockett:
## What does this change?

Removes SonarQube scanner and config as it is no longer in use. (See #191 for the original work to add this.)


jamesmockett:
[Remove SonarQube scanner and config from Manage](https://trello.com/c/t5oGoFkp/3465-remove-sonarqube-scanner-and-config-from-manage)

rBangay (approved):

PR #871, opened 2022-09-09
fix: Use correct type for CFN Parameters sourced from SSM

akash1810:
## What does this change?

In #869 we moved to loading the VPC parameters from SSM Parameter Store, however we forgot to update the `Type` which tells CloudFormation where to source the value from.

In this change, we set the `Type` to [`AWS::SSM::Parameter::Value`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types) so that it reads from SSM rather than use the raw string value.


rBangay (approved):

akash1810:
[Deployed to CODE successfully.](https://riffraff.gutools.co.uk/deployment/view/42ffb254-5eb0-460d-bafa-52265d973bbd)

PR #874, opened 2022-09-15
Fix logic which selects correct Stripe Public Key

paulbrown1982:
## What does this change?

There have been increasing reports of Recurring Contributors, Digital Subscribers and Supporter Members being charged an AU banking fee for their AUD subscriptions.

Upon investigation, this appears to be due to some logic in manage-frontend which is not selecting the AU Stripe Public Key properly, as the account affected has no delivery country to allow Manage-Frontend to make its decision. 

Members Data API [has been enhanced ](https://github.com/guardian/members-data-api/pull/795) to return the Billing Country name.

## How to test

Create an Australia Recurring Contribution using PayPal.
Log in to MMA
Change method to Card.
Zuora Gateway should say `Stripe PaymentIntents GNM Membership AUS` rather than `Stripe PaymentIntents GNM Membership` which it will do without this test.

Due to protections in MDAPI, cross wiring should not be possible (where gateway does not match tokens).

### DIFF
--- a/client/components/payment/update/PaymentDetailUpdate.tsx
+++ b/client/components/payment/update/PaymentDetailUpdate.tsx
@@ -332,7 +332,9 @@ const PaymentDetailUpdate = (props: WithProductType<ProductType>) => {
 	const getInputForm = (subscription: Subscription, isTestUser: boolean) => {
 		let stripePublicKey: string | undefined;
 
-		if (subscription.card) {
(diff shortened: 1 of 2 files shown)


rBangay (approved):

rupertbates (approved):
👍

rBangay (approved):

PR #875, opened 2022-09-15
Decouple payment update spec from GoCardless

jamesmockett:
## What does this change?

The payment update spec was calling the real `/api/validate/payment/dd?mode=live` endpoint which in turn was calling out to GoCardless. This intercepts the API call and mocks the response in the relevant payment update tests so there is no reliance on real APIs.

It's not entirely clear why this was previously working, so it's possible that it may fail again.

### DIFF
--- a/cypress/integration/parallel-1/updatePaymentDetails.spec.ts
+++ b/cypress/integration/parallel-1/updatePaymentDetails.spec.ts
@@ -88,6 +88,11 @@ describe('Update payment details', () => {
 			body: [digitalDD],
 		}).as('product_detail');
 
+		cy.intercept('POST', '/api/validate/payment/**', {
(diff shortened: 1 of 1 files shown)


jonflynng (approved):
Looks good. How did you find out about this by the way?

jamesmockett:
> Looks good. How did you find out about this by the way?

I only found this as the spec randomly starting failing in all of my branches and on `main`. At first I thought it was due to a copy change as the test is looking for a specific error message about validating bank details and was getting a slightly reworded version back instead.

This particular error message is inside a `catch` block and only shown if there's a problem when calling the validation API. After looking in the network tab when running Cypress I realised that API hadn't been mocked so it was calling the real thing.

PR #876, opened 2022-09-16
Update `caniuse-lite` database

jamesmockett:
## What does this change?

Updates the Browserslist `caniuse-lite` database. This is used for queries such as `>0.25%` in our current config and ensures we're using up-to-date browser usage data. (Updating the database also gets rid of constant warnings about it being outdated when running `yarn`.)

Longer term we should remove our custom Browserslist config and replace with [`@guardian/browserslist-config`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/browserslist-config). As the standard config doesn't include support for IE11 we will need to consider our stance on supporting this browser. (As a single page React app IE11 users will be served a blank page without the necessary transpilation steps and polyfills included.)


rBangay (approved):

jamesmockett:
[Update `caniuse-lite` database](https://trello.com/c/Kh8hGTOF/3489-update-caniuse-lite-database)

PR #877, opened 2022-09-16
Introduce route53

rBangay:
## What does this change?
Introduce the route53 DNS layer in between fastly and the load balancers. 

This helps us manage the migration to cdk, as part of the dual stack deployment (before we tear down the old stack) we can make routing changes at the DNS (route53) level instead of messing around in fastly.

Added a new deploy-code script in package.json and explicitly set the `App` tag via cdk hence why the snapshot is larger than just the route53 changes.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -6,8 +6,10 @@ const app = new App();
 new ManageFrontend(app, 'ManageFrontend-CODE', {
 	stack: 'support',
 	stage: 'CODE',
+	domain: 'manage.code.dev-theguardian.com.origin.membership.guardianapis.com',
 });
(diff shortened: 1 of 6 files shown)


akash1810 (dismissed):

cdk/bin/cdk.ts
```diff
@@ -6,8 +6,12 @@ const app = new App();
 new ManageFrontend(app, 'ManageFrontend-CODE', {
 	stack: 'support',
 	stage: 'CODE',
+	domain: 'manage.code.theguardian.com.origin.membership.guardianapis.com',
```
rBangay, line 9: this doesn't follow the `-dev` convention in order to match the certificate domain (which also doesn't have `-dev` in it)

akash1810 (approved):

PR #879, opened 2022-09-21
Exclude `node_modules` from server bundle

jamesmockett:
## What does this change?

Exclude `node_modules` from server bundle. These modules are already available on the server and generate warnings when bundled:

```
WARNING in ./node_modules/express/lib/view.js 81:13-25
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/express/lib/application.js 22:11-28
 @ ./node_modules/express/lib/express.js 18:12-36
 @ ./node_modules/express/index.js 11:0-41
 @ ./server/server.ts 4:0-45 11:15-22 35:22-36

1 warning has detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.
```

The [`webpack-node-externals`](https://github.com/liady/webpack-node-externals) package is used to scan `node_modules` to generate a list of modules to exclude from the bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
     "webpack-cli": "^4.9.1",
     "webpack-dev-server": "^4.4.0",
     "webpack-merge": "^5.8.0",
+    "webpack-node-externals": "^3.0.0",
     "whatwg-fetch": "^2.0.4"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


rBangay (approved):
👍

jamesmockett:
[Exclude `node_modules` from server bundle](https://trello.com/c/fhL3UQHh/3500-exclude-nodemodules-from-server-bundle)

PR #880, opened 2022-09-22
Switch cancellation API for Supporter Plus

jamesmockett:
## What does this change?

Updates the product cancellation flow to use the product movement API if the product being cancelled is Supporter Plus. All other products continue to use the existing cancellation API.

Supporter Plus is not currently available in production so this change should have no impact. We're explicitly checking if the product being cancelled is Supporter Plus when switching API. The existing cancellation API is the default and therefore used for all other products.


jamesmockett:
[MMA Change Cancellation Endpoint for New Prop](https://trello.com/c/AxQEOHbJ/3497-mma-change-cancellation-endpoint-for-new-prop)

PR #878, opened 2022-09-16
Use `@guardian/eslint-config-typescript`

jamesmockett:
## What does this change?

Use standard Guardian ESLint config ([`@guardian/eslint-config-typescript`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-config-typescript)).

As the codebase was built using a more relaxed set of rules the updated config generates a large number of errors. These rules have been overridden for now and will be progressively fixed in subsequent PRs. (To attempt to do so in one go would result in a huge PR that would be difficult to test and review.)

This PR overrides the value of `ecmaVersion` as this is erroneously set to [`es2020` rather than `2020`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/parser#parseroptionsecmaversion) in `@guardian/eslint-config-typescript`. This affects how scoping works in certain rules [causing `prefer-const` to throw errors when it shouldn't](https://github.com/eslint/eslint/issues/15461). eg:

```ts
let config;
try {
    config = await getConfig();
} catch (e) {
    handleError(e, res, next);
    return;
}
const options = getOptions('GET', req.cookies, config);
```

## Run time

The updated config has a fairly significant impact on run time due to the increased rule set. To mitigate this [`lint-staged`](https://github.com/okonet/lint-staged) has been added to the pre-commit hook so that we only lint files that have changed. (This is also used for Prettier, replacing `pretty-quick`.) Running ESLint against the entire codebase now only happens during CI where the increased run time has less impact.

### New config

Rule                                        | Time (ms) | Relative
:-------------------------------------------|----------:|--------:
import/default                              |  2207.720 |    22.7%
import/no-named-as-default-member           |  1577.147 |    16.3%
import/no-unresolved                        |  1378.754 |    14.2%
import/no-named-as-default                  |  1369.280 |    14.1%
@typescript-eslint/no-unnecessary-qualifier |  1180.074 |    12.2%
@typescript-eslint/no-unused-vars           |   658.563 |     6.8%
react/no-direct-mutation-state              |   248.076 |     2.6%
@typescript-eslint/await-thenable           |   158.251 |     1.6%
react/no-unknown-property                   |   134.176 |     1.4%
react/prop-types                            |    99.988 |     1.0%

Total time: 43.71s.

### Old config

Rule                                    | Time (ms) | Relative
:---------------------------------------|----------:|--------:
import/no-duplicates                    |   867.675 |    41.4%
@typescript-eslint/no-unused-vars       |   332.145 |    15.8%
[...]

PR #882, opened 2022-09-27
Lint rules fixes

rBangay:
## What does this change?

This follows on from #878 which introduced `@guardian/eslint-config-typescript`.

Overrides for the following linting rules have been removed and any errors / warnings in the code fixed:

| Rule |
| --- |
|array-type|
|consistent-indexed-object-style|


jamesmockett (approved):
👍

PR #889, opened 2022-10-03
tear down old cf.yaml resources

rBangay:
## What does this change?

As part of the guCDK migration workflow, now that we have a dual stack deployment (old yaml and new cdk) with the DNS switched over to the new cdk side. We can now tear down the old cfn.yaml resources (apart from the cloudwatch dashboard) and move towards a single stack deployment of just the guCDK resources.


akash1810:
Looks great! Has it deployed to CODE successfully?

akash1810 (approved):

PR #892, opened 2022-10-10
[RB] - add withoutImdsv2 to EC2 class to re-enable cloudwatch logs

rBangay:
## What does this change?

In reference to this pr: https://github.com/guardian/support-frontend/pull/3647

We need to add the `withoutImdsv2 ` property to the GuEc2App class in order to make the instance compatible with the  cloudwatch agent. Without doing this the logs disappear.


rupertbates (approved):
👍

PR #893, opened 2022-10-10
Remove tab role from left hand nav

andrewHEguardian:
## What does this change?

We're currently applying `role="tablist"` to the left-hand navigation although it's not actually a tab container and we're not applying any other tab related roles elsewhere.

The currently selected item is visually highlighted. We should use `aria-current` to also highlight the selected item to assistive technology.

## How to test

The `role="tablist"` attribute no longer applies to the left hand nav.
The current page that is visually highlighted has attribute `aria-current='page'`

### DIFF
--- a/client/components/nav/dropdownNav.tsx
+++ b/client/components/nav/dropdownNav.tsx
@@ -205,7 +205,7 @@ export const DropdownNav = () => {
 				My account
 			</button>
 
-			<ul role="tablist" css={dropdownNavCss(showMenu, isHelpCentre)}>
+			<ul css={dropdownNavCss(showMenu, isHelpCentre)}>
 				{Object.values(NAV_LINKS).map(
 					(navItem: MenuSpecificNavItem) => (
 						<li key={navItem.title}>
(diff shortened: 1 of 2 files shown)


jamesmockett (approved):
Looks good 👍  It feels like we should also have `aria-current` on the dropdown nav for mobile, except this is still shown at wider viewports when the left-hand nav is visible so would probably be confusing. (The current page isn't visually highlighted in the dropdown nav though so it's consistent.) I think there's a wider question about whether we really need two versions of the nav on the same page.

KentGrm:
For simple visual changes, it's always nice to see a screen shot of before / after at desktop and mobile resolution. Makes it easier to dip into the code changes.

andrewHEguardian:
[Remove tab role from left-hand nav and highlight current selection to assistive technology](https://trello.com/c/4jtCxyI3/3522-remove-tab-role-from-left-hand-nav-and-highlight-current-selection-to-assistive-technology)

jamesmockett:
> For simple visual changes, it's always nice to see a screen shot of before / after at desktop and mobile resolution. Makes it easier to dip into the code changes.

Completely agree. In this case the changes are purely updating accessibility roles so there are no visual changes.

PR #895, opened 2022-10-11
Fix `SupportTheGuardianButton` component accessibility issues

jamesmockett:
## What does this change?

Replaces the existing custom button markup and styles with a Source `<LinkButton>` and colour themes. Previously this was constructed as a button element nested inside an anchor which made it confusing and difficult to interact with when using a screenreader or navigating the site by keyboard. Due to having two interactive elements the button would receive focus twice.

The component's open-ended styling API has also been stripped out and replaced with a `theme` and `size` prop to ensure consistency and constrain the possible variants to those that are actually used across the site. (The themes could probably be simplified further as one is only used in a single place.)

This component primarily appears in the footer, but is also used in various other locations across the site.

## How to test

- [ ] The button can be navigated to and activated by keyboard
- [ ] When navigated to by keyboard the button only receives focus once
- [ ] The button takes you to the _Contribute_ or _Subscribe_ page as appropriate
- [ ] Click events are tracked in Ophan and GA

### Before

Screenshot 2022-10-10 at 17 50 48

Screenshot 2022-10-11 at 18 11 23

### After

Screenshot 2022-10-10 at 17 51 10

Screenshot 2022-10-11 at 18 11 33


jamesmockett:
[Fix markup of 'Contribute' and 'Subscribe' buttons in footer.](https://trello.com/c/IzfeWxjL/3519-fix-markup-of-contribute-and-subscribe-buttons-in-footer)

rBangay (approved):

rBangay (approved):

PR #896, opened 2022-10-12
eslint enable and fix consistent-type-imports rule

rBangay:
## What does this change?

The `consistent-type-imports` rule requires us to be explicit when we import a type


jamesmockett (approved):
👍 (I hope ESLint autofixed these all for you!)

PR #898, opened 2022-10-13
Use `@guardian/tsconfig` and simplify TypeScript config

jamesmockett:
## What does this change?

Uses `@guardian/tsconfig` as a base for our TypeScript config and simplifies things by stripping out unnecessary and commented out options.

`outDir` has been removed and `noEmit` added as TS compilation happens via Webpack rather than running `tsc`. We use `tsc` for type checking and this prevents it from generating any output if run without the `--noEmit` flag.

The Cypress TS config was previously based on the app's root config, although some of it is unnecessary so this has been stripped back to a minimal version.

### Disabled options

The `noImplicitReturns` and `noUncheckedIndexedAccess` options from `@guardian/tsconfig` have been disabled as the app includes code that violates these checks. As with our linting errors, we should aim to fix these and re-enable the options.

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -1,64 +1,6 @@
 {
+  "extends": "@guardian/tsconfig/tsconfig.json",
   "compilerOptions": {
-    /* Basic Options */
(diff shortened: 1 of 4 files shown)


jamesmockett:
[Use standard Guardian TypeScript config](https://trello.com/c/hZVbb2UY/3532-use-standard-guardian-typescript-config)

rBangay (approved):

PR #899, opened 2022-10-14
SPIKE: Account Overview updates for new prop

jamesmockett:
## What does this change?

Spikes changes to Account Overview page to support new product prop:

- Products are grouped as _Recurring support_ and _Subscriptions_.
- For _Recurring support_ products the billing interval is shown instead of the product name. eg. _Monthly_ instead of _Recurring contribution_. (This hasn't been done for Supporter Plus as this still comes through as a subscription.)
- Subscription ID can be exposed for contributions, but is displayed as _Supporter ID_.

These changes only affect contributions at the moment and need to be replicated for Supporter Plus. (This also needs to be recategorised in the API response.) There may be other instances of _Subscription ID_ that need to be changed to _Supporter ID_ for recurring support products.

## Images

Screenshot 2022-10-14 at 16 45 57

### DIFF
--- a/client/components/accountoverview/accountOverview.tsx
+++ b/client/components/accountoverview/accountOverview.tsx
@@ -1,6 +1,7 @@
 import { css } from '@emotion/react';
 import { headline, neutral, space, until } from '@guardian/source-foundations';
 import { Stack } from '@guardian/source-react-components';
+import { capitalize } from 'lodash';
 import { Fragment } from 'react';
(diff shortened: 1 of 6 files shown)


jamesmockett:
[SPIKE: Account overview product grouping and naming changes](https://trello.com/c/COpaXFJs/314-spike-account-overview-product-grouping-and-naming-changes)

PR #900, opened 2022-10-14
Manage buttons accessible names

andrewHEguardian:
## What does this change?

If you have multiple products you'll see multiple buttons labelled Manage subscription and Manage payment method on the Account overview page.

Screenreader users often navigate around a page by getting a list of all of the headings, links or buttons. When presented with multiple buttons with the exact same label out of context they have no way of knowing which product the button refers to.

We should add an accessible name to each of these buttons using aria-label so that screenreader users will hear something like Manage Guardian Weekly subscription instead of just Manage subscription. 

As the manage text is repeated it may make sense to put the product name first to allow users to find the product the want to manage more quickly. eg. Guardian Weekly — manage subscription and Guardian Weekly — manage payment method. Hidde de Vries has written an [excellent article on making better accessible names](https://hidde.blog/better-accessible-names/). In this PR the separator between the product name and the action is a colon - which seems to be read out better than a dash when the product name also contains a dash.

## How to test

- The buttons should be visually the same as before, except for minor differences between source and the previous buttons such as focus state
- The buttons should be functionally the same as before
- The extra information should be readable by a screen reader
- The description should be "Product : Action"

## Images

A button in focus state and a button with an alert.
Before             |  After
image  |  image

## Accessibility

### DIFF
--- a/client/components/accountoverview/accountOverviewCard.tsx
+++ b/client/components/accountoverview/accountOverviewCard.tsx
@@ -7,13 +7,15 @@ import {
 	space,
 	textSans,
 	until,
+	visuallyHidden,
 } from '@guardian/source-foundations';
+import { Button } from '@guardian/source-react-components';
+import { useNavigate } from 'react-router-dom';
 import { cancellationFormatDate, parseDate } from '../../../shared/dates';
 import type { ProductDetail } from '../../../shared/productResponse';
 import { getMainPlan, isGift } from '../../../shared/productResponse';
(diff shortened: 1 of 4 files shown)


andrewHEguardian:
[Add accessible names for manage buttons on 'Account overview' page](https://trello.com/c/2YUNTtPm/3527-add-accessible-names-for-manage-buttons-on-account-overview-page)

rBangay (approved):
👍  Well done Andrew 🎉

PR #911, opened 2022-11-01
Update cypress test that failed sometimes

andrewHEguardian:
## What does this change?

The Cypress test 'Cancel contribution' assertion 'attempts cancellation but then updates payment details instead. (reason: A payment issue)' would fail on some runs of the entire suite. It was searching for text that occurs multiple times on the page if the whole page loads. This is seemingly a timing issue since it usually passes if the test is run alone.


jamesmockett:
It looks like it was matching the 'Manage payment method' text against the title of the skeleton loader page before the actual _Billing_ page had even loaded. 'Manage payment method' appears more than once on the _Billing_ page so the test fails if it does get that far.

jamesmockett (approved):
I never thought our tests would be _too_ fast so something to watch out for in the future 👍

PR #913, opened 2022-11-02
Show product info table on Billing page for contributions

jamesmockett:
## What does this change?

Reveals the product info table for contributions on the _Billing_ page. Previously this was shown for all other products, but explicitly hidden for contributions. Now that we're able to show _Supporter ID_ in place of _Subscription ID_ for recurring support products there's no longer any need to hide this information.

### Before

Screenshot 2022-11-02 at 12 28 03

### After

Screenshot 2022-11-02 at 12 29 25

### DIFF
--- a/client/components/billing/billing.tsx
+++ b/client/components/billing/billing.tsx
@@ -19,7 +19,6 @@ import type {
 import {
 	getMainPlan,
 	isGift,
-	isPaidSubscriptionPlan,
 	isProduct,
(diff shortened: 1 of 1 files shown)


andrewHEguardian (approved):
👍 interesting that it wasn't checking it was a contribution by the product type

jamesmockett:
Good point. Was a fairly complicated way of checking when it already has the product type available 🤷

PR #914, opened 2022-11-02
Allow cancellation when visiting cancellation page directly

jamesmockett:
## What does this change?

Fixes two separate bugs that prevent cancellation when visiting the cancellation page directly. eg. `/cancel/contributions`

The first of these is due to relying on the presence of `productDetail` to derive the grouped product type from the `mmaCategory` property in order to set the page title. `productDetail` is passed through as part of the router state which will be undefined if visiting the page directly. We do a fetch from MDAPI to get product details for subsequent steps, but this happens after we've set the page title so the initial reference throws an error.

To fix this we've added a `groupedProductType` property to `ProductType` which allows us to get a product's grouped product type without relying on `mmaCategory` from the API response. Once we've removed all references to `mmaCategory` in subsequent PRs we will be able to remove this property from MDAPI entirely.

The second bug is due to `<CancellationSwitchEligibilityCheck>` assuming that router state is available. If this is undefined the user is redirected back to _Account Overview_. However, `productDetail` _is_ available from `CancellationContext` regardless of whether it was passed through via router state or fetched from MDAPI when visiting the page directly. Rather than getting `productDetail` from the router state we now get it from the context instead which allows the cancellation flow to work regardless of how the user arrived at the page.


shared/productTypes.tsx
```diff
@@ -693,12 +705,14 @@ export const GROUPED_PRODUCT_TYPES: {
 		...PRODUCT_TYPES.contributions,
 		mapGroupedToSpecific: () => PRODUCT_TYPES.contributions,
 		groupFriendlyName: 'contributions',
+		groupedProductType: 'contributions',
```
jamesmockett, line 708: Because `GroupedProductType` extends `ProductType` we have to include the `groupedProductType` property in the grouped property types and refer back to themselves. We can't omit this property when extending `ProductType` due to the `mapGroupedToSpecific` function which may return a grouped product type instead of a product type. (Although this is a fallback that shouldn't ever happen.)

jamesmockett:
[MMA: Cancel Sub Bug](https://trello.com/c/g3ME2ubU/3601-mma-cancel-sub-bug)

andrewHEguardian (approved):

PR #915, opened 2022-11-04
Add default product title

jamesmockett:
## What does this change?

`productTitle` has an optional `mainPlan` argument, but contributions and Supporter Plus, which use the plan to generate the title, don't have a fallback if `mainPlan` is empty. This adds a default title for these two products. The only instance of this happening that I'm aware of is in the cancelled product cards on _Account Overview_ due to `CancelledProductDetail` only including a subset of subscription properties. (Crucially, `interval` is not one of these.)

## Images

Screenshot 2022-11-04 at 09 50 17


andrewHEguardian (approved):
👍 happy with those as defaults

PR #920, opened 2022-11-09
Fix contribution cancellation spec

jamesmockett:
## What does this change?

Updates spec to match new copy in cancellation flow.


rBangay (approved):
👍 great work, very complicated

PR #923, opened 2022-11-16
Storybook top level pages

andrewHEguardian:
## What does this change?

Add top level pages to Storybook:

- [x] Cancellation > Confirmation
- [x] Update payment method
- [x] Update delivery details
- [x] Update delivery details > Review
- [x] Update delivery details > Confirmation
- [x] Delivery history
- [x] Delivery history > Review
- [x] Delivery history > Confirmation
- [x] Manage suspensions
- [x] Manage suspensions > Create suspension

Also adds the Mock Service Worker (MSW) framework to mock requests, it provides more detail and feedback than fetch-mock - which caused trouble using in some of the tests. Following this [guide](https://github.com/mswjs/msw-storybook-addon) to set up MSW, with the only difference being that the mockServiceWorker.js file is located in '.storybook/static' instead of a 'public' or 'dist' folder as suggested. We only need MSW for Storybook so there is no need to put it there.

Decided not to add pages for Manage suspensions > Review and Confirm - these pages rely on dates being selected in the 'Create suspension' step and being present in the page's state - which would require a lot of work to implement in Storybook.


andrewHEguardian:
[TASK: Add remaining top-level pages to Storybook](https://trello.com/c/UxCIzsMI/2891-task-add-remaining-top-level-pages-to-storybook)

rBangay (dismissed):
Nice work Andrew 👍

rBangay (approved):

PR #925, opened 2022-11-17
Update support button URL and surrounding copy

jamesmockett:
## What does this change?

Updates the `<SupportTheGuardianButton>` component so that it points to the new contribute page by default. The default button text has also been changed from 'Support The Guardian' to 'Support the Guardian' to follow Guardian style. Surrounding copy on pages where this button appears have also been updated accordingly.

## Images

Screenshot 2022-11-17 at 12 43 48

Screenshot 2022-11-23 at 10 25 47

### DIFF
--- a/client/components/resubscribeThrasher.tsx
+++ b/client/components/resubscribeThrasher.tsx
@@ -67,8 +67,7 @@ const getThrasher =
 				>
 					<div>
 						<h2 css={{ fontWeight: 'bold', margin: '0' }}>
-							Have you considered a monthly or annual
-							contribution?
+							Support us another way?
 						</h2>
 						<p
 							css={{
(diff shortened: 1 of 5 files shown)


jamesmockett:
[MMA: Update 'Support' footer button to go to V2 Checkout](https://trello.com/c/g573y5be/3622-mma-update-support-footer-button-to-go-to-v2-checkout)

client/components/resubscribeThrasher.tsx
```diff
@@ -80,14 +79,13 @@ const getThrasher =
 								},
 							}}
 						>
-							Support The Guardian with a recurring contribution
-							of your choice. You can use your existing payment
-							details so setting it up only takes a minute.
+							From just $1, or a little more on a regular basis,
```
rBangay, line 82: ello 👋  ... is this always supposed to be a dollar value?

rBangay (approved):

client/components/resubscribeThrasher.tsx
jamesmockett: We don't have easy access to the payment currency in the thrasher unfortunately. It does make an API call to fetch existing payment options, but this doesn't include the currency.

Iona is fine with this and has advised that we already do this in a number of other places where regionalisation is impossible or not worth the dev effort. (Her style guide also includes this as guidance.)

PR #928, opened 2022-11-21
Remove the 'Premium tier access' from help centre

rBangay:
## What does this change?

Remove the 'Premium tier access' option from the 'More Topics' section on the homepage of the help centre

## Images
Before             |  After
  |

### DIFF
--- a/client/components/helpCentre/helpCentreConfig.ts
+++ b/client/components/helpCentre/helpCentreConfig.ts
@@ -220,26 +220,21 @@ export const helpCentreMoreQuestionsConfig: HelpCentreMoreQuestionsTopic[] = [
 		links: [
 			{
 				id: 'q1',
-				title: 'Premium tier access',
-				link: '/help-centre/article/how-can-i-gain-access-to-the-premium-tier-of-your-app',
-			},
-			{
-				id: 'q2',
 				title: 'Apple/Google subscriptions',
 				link: '/help-centre/article/i-have-a-googleitunes-subscription-that-i-need-help-with',
 			},
 			{
-				id: 'q3',
+				id: 'q2',
 				title: 'Personalising your apps',
 				link: '/help-centre/article/making-your-app-more-personal',
 			},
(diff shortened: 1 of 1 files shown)


client/components/helpCentre/helpCentreConfig.ts
```diff
@@ -220,26 +220,21 @@ export const helpCentreMoreQuestionsConfig: HelpCentreMoreQuestionsTopic[] = [
...
-				link: '/help-centre/article/how-can-i-gain-access-to-the-premium-tier-of-your-app',
-			},
-			{
-				id: 'q2',
 				title: 'Apple/Google subscriptions',
 				link: '/help-centre/article/i-have-a-googleitunes-subscription-that-i-need-help-with',
 			},
 			{
-				id: 'q3',
+				id: 'q2',
```
andrewHEguardian, line 227: What are these IDs for?

jamesmockett: I was wondering about these IDs too as they get sent to Ophan when you click an article link:

```
"labels": ["HELP-CENTRE", "MORE-TOPICS-Q-CLICK", "THE-GUARDIAN-APPS-Q1"]
```

Are we confusing the stats by re-ordering questions? (ie. what was `q2` is now `q1`) I guess the actual page views are being logged separately though.

rBangay: Looks like they are used as keys for elements in a map (helpCentreLandingMoreTopics.tsx)

They are also used in the tracking of the clicking of those articles eg `eventLabel: `${topic.id}-${question.id}``

Which means that the questions/articles will shift up one. I'll check with Jess to see if this is an issue or not.

rBangay: Yep Jess has ok'd this change/shift to the id's 👍

jamesmockett (approved):
👍

PR #931, opened 2022-12-01
Supporter+ benefits dropdown

jamesmockett:
## What does this change?

Adds Supporter+ benefits section and toggle button.

## Images

manage thegulocal com_ (1)

manage thegulocal com_ (2)

## Accessibility

### DIFF
--- a/client/components/accountoverview/accountOverviewCardV2.tsx
+++ b/client/components/accountoverview/accountOverviewCardV2.tsx
@@ -230,8 +230,9 @@ export const AccountOverviewCardV2 = ({
 							`}
 						>
 							You’re supporting the Guardian with a{' '}
-							{nextPaymentDetails.paymentValue} per month support
-							and extra benefits.
+							{nextPaymentDetails.paymentValue} per{' '}
+							{nextPaymentDetails.paymentInterval} support and
+							extra benefits.
 						</p>
 						<ul
 							id="benefits"
--- a/client/components/accountoverview/accountOverviewCardV2.tsx
+++ b/client/components/accountoverview/accountOverviewCardV2.tsx
@@ -11,7 +11,8 @@ import {
 	buttonThemeReaderRevenueBrand,
 	SvgTickRound,
 } from '@guardian/source-react-components';
-import type { ReactNode } from 'react';
+import type { ReactNode} from 'react';
+import { useState } from 'react';
 import { useNavigate } from 'react-router';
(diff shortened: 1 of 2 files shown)


jamesmockett:
[MMA / Account Overview: Add Dropdown for S+ Benefits](https://trello.com/c/p11BjaNo/3638-mma-account-overview-add-dropdown-for-s-benefits)

client/components/accountoverview/accountOverviewCardV2.tsx
```diff
@@ -185,8 +187,105 @@ export const AccountOverviewCardV2 = ({
...
+									<strong>
+										A regular supporter newsletter.
+									</strong>{' '}
+									Get exclusive insight from our newsroom
+								</span>
+							</li>
+							<li>
+								<SvgTickRound size="small" />
+								<span>
+									<strong>Uninterrupted reading.</strong> See
```
andrewHEguardian, line 254: Some of these are using {' '} to create a space and some are just on the same line. Should we keep it consistent?

client/components/payment/nextPaymentDetails.tsx
```diff
@@ -34,13 +35,14 @@ export const getNextPaymentDetails = (
 			? paidFuturePlans[1]
 			: paidFuturePlans[0];
 	if (isPaidSubscriptionPlan(mainPlan)) {
-		const paymentKey = `Next ${augmentInterval(
+		const paymentInterval =
 			subscription.currentPlans.length !== 0 &&
-				isSixForSix(mainPlan.name) &&
-				planAfterMainPlan
+			isSixForSix(mainPlan.name) &&
```
andrewHEguardian, line 40: Could we maybe split this up? I find it hard to understand what the logic is here

jamesmockett: Yeah, I completely agree that the logic in here is hard to follow. I think any refactoring of this is probably outside the scope of this PR though. All I've done here is to move the existing logic into a separate variable so we can get the interval (ie. `month`) in isolation from the existing `paymentKey` (ie.`Next monthly payment`)

A lot of the interval logic is tied into whether it's a six for six plan, but I need to find out whether that is still a thing. I don't believe it's something you can sign up for anymore, but there may be readers still on that plan that we need to accommodate.

andrewHEguardian (approved):
Looks good 👍

PR #933, opened 2022-12-05
Wire up 'Change to support and extras' button to logic to determine whether to show it

andrewHEguardian:
## What does this change?

Adds logic to determine when to show 'Change to support and extras' button. The button is shown on a product card if:

- the current product is a Recurring Contribution
- user does not have DigiSub **and** RC
- user is not in payment failure

## How to test

Storybook stories have been added to test the different scenarios where

- user has DigiSub and RC
- user is in payment failure
- user only has RC

## Accessibility

### DIFF
--- a/client/components/accountoverview/accountOverview.tsx
+++ b/client/components/accountoverview/accountOverview.tsx
@@ -15,9 +15,16 @@ import type {
 	MembersDataApiItem,
 	ProductDetail,
 } from '../../../shared/productResponse';
-import { isProduct, sortByJoinDate } from '../../../shared/productResponse';
+import {
+	isProduct,
+	isSpecificProductType,
+	sortByJoinDate,
+} from '../../../shared/productResponse';
 import type { GroupedProductTypeKeys } from '../../../shared/productTypes';
-import { GROUPED_PRODUCT_TYPES } from '../../../shared/productTypes';
+import {
+	GROUPED_PRODUCT_TYPES,
+	PRODUCT_TYPES,
+} from '../../../shared/productTypes';
 import { fetchWithDefaultParameters } from '../../fetch';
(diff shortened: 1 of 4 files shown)


andrewHEguardian:
[Account Overview: CTA allowing user 'Change to support and extras' [S]](https://trello.com/c/KDB1cnyH/3555-account-overview-cta-allowing-user-change-to-support-and-extras-s)

client/components/accountoverview/accountOverview.tsx
```diff
@@ -62,6 +69,18 @@ const AccountOverviewRenderer = ([mdaResponse, cancelledProductsResponse]: [
 		(_) => _.alertText,
 	);
 
+	const hasDigiSubAndContribution =
+		allActiveProductDetails.some((productDetail) =>
+			isSpecificProductType(productDetail, PRODUCT_TYPES.contributions),
```
jamesmockett, line 74: Could you use an OR here rather than looping through all of the active products twice? eg.

`isSpecificProductType(productDetail, PRODUCT_TYPES.contributions) || isSpecificProductType(productDetail, PRODUCT_TYPES.digipack)`

andrewHEguardian: I thought about this, but wouldn't that be true if you had one or the other but not necessarily both?

jamesmockett: Ah, sorry I misread the existing logic 🤦  Yes, you're absolutely right as you need to check that they have _both_ products 👍

jamesmockett (approved):
Sorry for adding to the product type confusion 😅 This looks good 👍 Really nice to have those additional examples in Storybook too.

PR #937, opened 2022-12-15
move the known issues config/copy into the client codebase so that it…

rBangay:
## What does this change?
Move the known issues config/copy from a json file in an S3 bucket into the client codebase so that it is documented.

An entry has been made in the wiki to explain the process of adding an issue to the help centre (https://github.com/guardian/manage-frontend/wiki/HOW-TOs#adding-a-known-issue-banner-to-the-help-centre)

### DIFF
--- a/client/components/helpCentre/KnownIssues.tsx
+++ b/client/components/helpCentre/KnownIssues.tsx
@@ -12,26 +12,45 @@ import { gridBase, gridItemPlacement } from '../../styles/grid';
 import { allProductsDetailFetcher } from '../../utilities/productUtils';
 import { ErrorIcon } from '../mma/shared/assets/ErrorIcon';
 
-interface Issue {
-	date: string;
+/*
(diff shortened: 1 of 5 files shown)


client/components/helpCentre/KnownIssues.tsx
```diff
@@ -12,26 +12,45 @@ import { gridBase, gridItemPlacement } from '../../styles/grid';
 import { allProductsDetailFetcher } from '../../utilities/productUtils';
 import { ErrorIcon } from '../mma/shared/assets/ErrorIcon';
 
-interface Issue {
-	date: string;
+/*
```
andrewHEguardian, line 15: I'm OK with commenting this out - but I think we should maybe put a ticket into the health board to take a look at this at a later date.

andrewHEguardian (dismissed):
Thanks for documenting this!

andrewHEguardian (approved):

PR #939, opened 2022-12-16
Create new Product Switching route and options page

andrewHEguardian:
## What does this change?

This PR adds a new route and first page for the product switching journey under `/switch` and the 'Your current support' block of the Options page. 

## How to test

This is behind the 'Change to monthly + extras' button which is behind the feature flag 'accountOverviewNewLayout'. Alternatively navigate to '/switch' with a contribution product.

Acceptance criteria:
Wording should display 'Monthly' or 'Annual' support
Should display current spend per month / year
Should display correct currency
Should include a message
'You're currently supporting the Guardian with a [insert monthly OR annual] contribution of [insert spend here].

## Images

image

## Accessibility

### DIFF
--- a/client/components/mma/shared/Card.tsx
+++ b/client/components/mma/shared/Card.tsx
@@ -1,4 +1,3 @@
-import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
 import { from, palette, space } from '@guardian/source-foundations';
 import type { ReactNode } from 'react';
@@ -10,7 +9,7 @@ export const Card = (props: { children: ReactNode }) => {
 Card.Header = (props: {
 	children: ReactNode;
 	backgroundColor?: string;
-	cssOverrides?: SerializedStyles | SerializedStyles[];
+	headerHeight?: number;
 }) => {
 	const headerCss = css`
 		padding: ${space[3]}px ${space[4]}px;
@@ -18,13 +17,11 @@ Card.Header = (props: {
 		background-color: ${props.backgroundColor ?? palette.neutral[97]};
 
 		${from.tablet} {
-			min-height: 128px;
+			min-height: ${props.headerHeight ?? 128}px;
 		}
 	`;
 
-	return (
-		<header css={[headerCss, props.cssOverrides]}>{props.children}</header>
-	);
+	return <header css={headerCss}>{props.children}</header>;
 };
 
 Card.Section = (props: { children: ReactNode; backgroundColor?: string }) => {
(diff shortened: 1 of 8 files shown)


client/components/mma/shared/Card.tsx
```diff
@@ -0,0 +1,42 @@
...
+import type { ReactNode } from 'react';
+
+export const Card = (props: { children: ReactNode }) => {
+	return <div>{props.children}</div>;
+};
+
+Card.Header = (props: {
+	children: ReactNode;
+	backgroundColor?: string;
+	cssOverrides?: SerializedStyles | SerializedStyles[];
```
jamesmockett, line 13: I tend to prefer _not_ allowing arbitrary CSS overrides in components as it can make things more brittle. Any future changes to the component's core styles might conflict with any overrides that are being used elsewhere. It also means a component potentially has infinite variations. That said, it's a pattern that's already used in all of the Source components 😅

It's less of an issue here as we're the only consumer of the component, but once we've built more of the pages it might be nice to see what overrides we've used and add specific variants to the component to handle those cases so we can remove the overrides.

client/components/mma/switch/SwitchOptions.tsx
```diff
@@ -0,0 +1,109 @@
...
+	const mainPlan = getMainPlan(
+		productDetail.subscription,
+	) as PaidSubscriptionPlan;
+	const monthlyOrAnnual = calculateMonthlyOrAnnualFromInterval(
+		mainPlan.interval,
+	);
+
+	return (
+		<Stack space={3} cssOverrides={pageTopCss}>
+			<Heading cssOverrides={pageHeadingCss}>
```
jamesmockett, line 76: Rather than adding a CSS override it might be better to have a new prop that allows switching the header style from serif to sans-serif as we're going to want to do this on every page as well as the Account Overview. (It will probably become the default heading style everywhere eventually, although only a handful of places use this component currently.)

jamesmockett (approved):
This is looking really good, nice work 👍 The comment about changing the `<Heading>` component doesn't necessarily need to be done in this PR as I don't want to hold up merging this.

client/components/mma/shared/Card.tsx
andrewHEguardian: I like this 👍

client/components/mma/switch/SwitchOptions.tsx
andrewHEguardian: I've added this as a new prop because I actually found using the cssOverride to be not so obvious - and had to set the style for the minimum breakpoint in order for it to actually override the default style.

andrewHEguardian:
[MMA: Switching Options Page: Your Current Support Block](https://trello.com/c/g5jaiMQ9/3722-mma-switching-options-page-your-current-support-block)

PR #940, opened 2022-12-16
More TypeScript linting fixes

jamesmockett:
## What does this change?

Fixes the following linting errors and removes overrides to disable these rules from the ESLint config:

`@typescript-eslint/no-unnecessary-type-arguments`
`@typescript-eslint/prefer-includes`
`@typescript-eslint/prefer-ts-expect-error`
`@typescript-eslint/restrict-plus-operands`

### DIFF
--- a/client/__tests__/components/payment/stripe.test.ts
+++ b/client/__tests__/components/payment/stripe.test.ts
@@ -4,7 +4,7 @@ import {
 } from '../../../fixtures/subscription';
 import { getStripeKey } from '../../../utilities/stripe';
 
-// @ts-ignore
+// @ts-expect-error
 window.guardian = {
 	stripeKeyAustralia: {
 		uat: 'uatKeyAustralia',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,7 +39,6 @@ module.exports = {
 		'@typescript-eslint/no-unsafe-call': ['off'],
 		'@typescript-eslint/no-unsafe-member-access': ['off'],
 		'@typescript-eslint/no-unsafe-return': ['off'],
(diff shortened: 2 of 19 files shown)


jamesmockett:
[Fix linting errors and remove overrides — Batch 2](https://trello.com/c/vmGsKe9c/3735-fix-linting-errors-and-remove-overrides-batch-2)

client/__tests__/components/payment/stripe.test.ts
```diff
@@ -4,7 +4,7 @@ import {
 } from '../../../fixtures/subscription';
 import { getStripeKey } from '../../../utilities/stripe';
 
-// @ts-ignore
+// @ts-expect-error
```
andrewHEguardian, line 7: Interesting. What's the benefit of expect error over ignore?

andrewHEguardian: Googled it: 

TypeScript allows you to suppress all errors on a line by placing a comment starting with @ts-ignore or @ts-expect-error immediately before the erroring line. The two directives work the same, except @ts-expect-error causes a type error if placed before a line that's not erroring in the first place.

This means its easy for @ts-ignores to be forgotten about, and remain in code even after the error they were suppressing is fixed. This is dangerous, as if a new error arises on that line it'll be suppressed by the forgotten about @ts-ignore, and so be missed.

andrewHEguardian (approved):
🙌

client/__tests__/components/payment/stripe.test.ts
jamesmockett: Exactly 😄 Seems like a sensible change.

PR #941, opened 2022-12-19
Create Change your support section

andrewHEguardian:
## What does this change?

Adds the 'Change your support' section to the Product Switch Options page.

When attempting to promote switching, we want to highlight to users the benefits of switching and what they can get when switching to Supporter+. Note: there are two different scenarios here and the block should display different content depending on the users payments (below or above threshold)

## How to test

Navigate to `/switch` with a recurring contribution.

Acceptance criteria:
- H2 should read 'Add extras' if above threshold
- H2 should read 'Change your support' if below threshold
- Change your support subtext should differ depending on Monthly or Annual + extras depending on what the user currently has.

Banner should read Monthly or Annual + extras depending on their current product
List of benefits should be in order as per designs, with app listed first
IF user is spending under threshold, the banner should display the threshold price. I.E spending £8, this should show £10
IF the user is spending £12 no price should show
IF above threshold, CTA should read 'Add extras with no extra cost'
IF below threshold, CTA should read 'change to monthly/annual + extras'
IF above threshold, Ts and Cs should show to allow users to know that there is a minimum spend, and any extra they are giving is out of choice.

## Images

Above threshold:
image

Below threshold:
image

## Accessibility

### DIFF
--- a/client/components/mma/shared/SupporterPlusBenefits.tsx
+++ b/client/components/mma/shared/SupporterPlusBenefits.tsx
@@ -1,0 +1,69 @@
+import { css } from '@emotion/react';
+import { from, palette, space, textSans } from '@guardian/source-foundations';
+import { SvgTickRound } from '@guardian/source-react-components';
+
+export const SupporterPlusBenefitsSection = (props: {
+	hidden?: boolean;
+	withMargin?: boolean;
+}) => {
+	const benefitsCss = css`
+		${textSans.medium()};
+		list-style: none;
+		margin: ${props.withMargin ? `${space[5]}px 0 ${space[4]}px -4px` : 0};
+		padding: 0;
+
+		li + li {
+			margin-top: ${space[2]}px;
+		}
+
+		li {
+			display: flex;
+			align-items: flex-start;
+		}
+
+		svg {
+			flex-shrink: 0;
+			margin-right: ${space[2]}px;
+			fill: ${palette.brand[500]};
+		}
+	`;
+
+	const lineBreakCss = css`
+		${from.tablet} {
+			display: none;
+		}
+	`;
+
+	return (
+		<ul id="benefits" css={benefitsCss} hidden={props.hidden}>
+			<li>
+				<SvgTickRound size="small" />
+				<span>
+					<strong>Full access to our news app.</strong>
+					<br css={lineBreakCss} /> Read our reporting on the go
+				</span>
+			</li>
+			<li>
+				<SvgTickRound size="small" />
+				<span>
+					<strong>A regular supporter newsletter.</strong> Get
+					exclusive insight from our newsroom
+				</span>
+			</li>
+			<li>
+				<SvgTickRound size="small" />
+				<span>
+					<strong>Uninterrupted reading.</strong>
+					<br css={lineBreakCss} /> See far fewer asks for support
+				</span>
+			</li>
+			<li>
+				<SvgTickRound size="small" />
+				<span>
+					<strong>Ad-free reading.</strong>
+					<br css={lineBreakCss} /> Avoid ads on all your devices
+				</span>
+			</li>
+		</ul>
+	);
+};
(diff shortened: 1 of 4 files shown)


client/components/mma/shared/SupporterPlusBenefits.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { SvgTickRound } from '@guardian/source-react-components';
+
+export const SupporterPlusBenefitsSection = (props: {
+	hidden?: boolean;
+	withMargin?: boolean;
+}) => {
+	const benefitsCss = css`
+		${textSans.medium()};
+		list-style: none;
+		margin: ${props.withMargin ? `${space[5]}px 0 ${space[4]}px -4px` : 0};
```
jamesmockett, line 12: I think it would be better if the component didn't include any spacing. The spacing at the top and bottom is specific to the layout in the product card so would be better handled there. (Also makes things simpler when this gets reused somewhere else and needs different spacing around it.)

You probably _do_ always want to apply `margin-left: -4px` though as that removes the spacing on the edge of the tick icon so it's flush with the left-hand edge.

client/components/mma/switch/SwitchOptions.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const AboveThreshold: ComponentStory<typeof SwitchOptions> = () => (
+	<SwitchOptions />
+);
+
+export const BelowThreshold: ComponentStory<typeof SwitchOptions> = () => (
+	<SwitchOptions />
+);
+
+// @ts-ignore: this method is not in our current version of TypeScript
```
jamesmockett, line 29: You'll need to change this to `@ts-expect-error` as I just merged the linting changes 😅

client/components/mma/shared/SupporterPlusBenefits.tsx
```diff
@@ -0,0 +1,69 @@
+import { css } from '@emotion/react';
+import { from, palette, space, textSans } from '@guardian/source-foundations';
+import { SvgTickRound } from '@guardian/source-react-components';
+
+export const SupporterPlusBenefitsSection = (props: {
+	hidden?: boolean;
```
jamesmockett, line 6: I'm not sure if this really needs a `hidden` prop as it's only used in a single place. You could wrap it in a `div` in the one place where it's used and hide that.

client/components/mma/switch/SwitchOptions.tsx
```diff
@@ -84,10 +109,69 @@ const SwitchOptions = () => {
 						You're currently supporting the Guardian with a{' '}
 						{monthlyOrAnnual.toLowerCase()} contribution of{' '}
 						{mainPlan.currency}
-						{mainPlan.amount / 100}.
+						{currentAmount}.
 					</div>
 				</Card.Section>
 			</Card>
+			<div css={pageTopCss}>
```
jamesmockett, line 116: There's already  a `<Stack>` wrapping everything on the page so could that be adjusted to give you the spacing you need between the sections? (You might then want to use nested stacks to space elements within. eg. headings and cards.)

client/components/mma/switch/SwitchOptions.stories.tsx
andrewHEguardian: Do you think this code is why the story didn't show up in Chromatic? Or is that a coincidence

jamesmockett: Oh, perhaps it doesn't support `structuredClone` or we need to polyfill it? You could clone the object in a slightly hackier (but well supported) way:

```js
const contributionBelowThreshold = JSON.parse(JSON.stringify(contribution));
```

(You'll see this being done in a couple of the Cypress specs.)

client/components/mma/switch/SwitchOptions.tsx
andrewHEguardian: I think the spacing should be different for mobile layout, but not sure if Stack spaces support that?

jamesmockett: Ah, okay. No, Stack doesn't currently support multiple spacing values for different breakpoints. I've got an idea how to do this, but need to put together a proposal and prototype 😅

I chatted to Helene about this whilst building the previous iteration and she was happy to leave the spacing the same for mobile and desktop, but we can leave it as is for now.

andrewHEguardian:
[MMA Switching Options Page: Change your support section](https://trello.com/c/5lCTfEyS/3724-mma-switching-options-page-change-your-support-section)

PR #942, opened 2022-12-21
Sticky product switch button

jamesmockett:
## What does this change?

Makes the product switch button sticky on smaller screens so that it is always visible at the bottom.

This uses an Intersection Observer to detect when the button is 'stuck' and not part of the normal document flow in order to apply additional styles. The observable area is reduced by `1px` on the bottom edge so that when the button is 'stuck' to the bottom of the viewport it is not considered to be fully visible. Similarly, the observable area is extended beyond the top of the viewport so that the 'stuck' state is not triggered when the button scrolls off the top of the page.

## Images

https://user-images.githubusercontent.com/1166188/208878586-99f0de64-d502-445b-95f7-f8876eeb3965.mov

### DIFF
--- a/client/components/mma/shared/Heading.tsx
+++ b/client/components/mma/shared/Heading.tsx
@@ -22,14 +22,14 @@ export const Heading = (props: HeadingProps) => {
 
 	const headingStyles = css`
 		${props.sansSerif
-			? textSans.xxlarge({ fontWeight: 'bold' })
+			? textSans.large({ fontWeight: 'bold' })
 			: headline.xsmall({ fontWeight: 'bold' })};
 
 		margin-top: 0;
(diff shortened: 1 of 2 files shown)


jamesmockett:
[MMA Switching Options Page: Sticky button (mobile only)](https://trello.com/c/fIh9h5Oo/3731-mma-switching-options-page-sticky-button-mobile-only)

rBangay (approved):
Nice work 👍 

Hopefully at some point they'll be a css `I'm stuck` selector that we can use

andrewHEguardian:
This is really cool!

andrewHEguardian (approved):
I'm not Richard Bangay and I approve this PR

PR #944, opened 2022-12-23
[RB] - update xmas opening hours 🎄🎁☃️

rBangay:
## What does this change?

Update to Christmas opening hours in the known issues banner and opening times component.

## Happy Christmas everyone 👋 
9e05f4ab-fc21-4aff-aad5-aa0c306d4cb8_text

### DIFF
--- a/client/components/helpCentre/HelpCentrePage.tsx
+++ b/client/components/helpCentre/HelpCentrePage.tsx
@@ -66,16 +66,19 @@ const HelpCentreRouter = () => {
 	]
 	*/
 
-	const ransomwareMessage = [
+	const knownIssues: KnownIssueObj[] = [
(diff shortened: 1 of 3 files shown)


client/components/helpCentre/HelpCentrePage.tsx
```diff
@@ -67,6 +67,11 @@ const HelpCentreRouter = () => {
 	*/
 
 	const ransomwareMessage = [
```
jamesmockett, line 69: Super minor thing, but should this be called something else as it has _multiple_ issues in it now? In fact, does it even need to be a separate var as it's being assigned to `knownIssues` immediately below?

jamesmockett (approved):

PR #949, opened 2023-01-05
Create app only URL and add text to Switch options page

andrewHEguardian:
## What does this change?

Background: Due to Apple laws we aren't able to express all of the copy we'd like to within the app, therefore we'd like a box to display if users come from the app. The expectation from our side is that users from the app would come directly to this page via a url with a string of text that would therefore expose the box of text.

This PR adds a specific URL for users coming from the app and persists this in context, and also adds additional copy on the switch options page

## How to test

Navigate to `/switch#from-app` and see the additional copy, without the extra URL specifier it should not be visible.

## Images

image

## Accessibility

### DIFF
--- a/client/components/mma/switch/SwitchOptions.tsx
+++ b/client/components/mma/switch/SwitchOptions.tsx
@@ -77,10 +77,8 @@ const buttonStuckCss = css`
 `;
 
 const fromAppHeadingCss = css`
-	${textSans.large({ fontWeight: 'bold' })}
+	${textSans.large({ fontWeight: 'bold', lineHeight: 'regular' })}
 	color:${palette.brand[500]};
-	line-height: 1.35;
-	border: 0;
 	margin-bottom: 0;
 `;
 
@@ -143,8 +141,8 @@ const SwitchOptions = () => {
 					>
 						To unlock unlimited reading in our news app, please make
 						a small change to your support type. If this doesn't
-						suit you, no change is needed, but note you will continue
-						to have limited app access.
+						suit you, no change is needed, but note you will
+						continue to have limited app access.
 					</p>
 				</div>
 			)}
--- a/client/components/mma/switch/SwitchOptions.tsx
+++ b/client/components/mma/switch/SwitchOptions.tsx
@@ -143,7 +143,7 @@ const SwitchOptions = () => {
 					>
 						To unlock unlimited reading in our news app, please make
 						a small change to your support type. If this doesn't
-						suit you, no change is needed, but not you will continue
+						suit you, no change is needed, but note you will continue
 						to have limited app access.
 					</p>
 				</div>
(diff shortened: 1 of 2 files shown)


andrewHEguardian:
[MMA Switching Options Page: App only text](https://trello.com/c/dbXebNry/3730-mma-switching-options-page-app-only-text)

client/components/mma/switch/SwitchOptions.tsx
```diff
@@ -76,6 +76,14 @@ const buttonStuckCss = css`
 	}
 `;
 
+const fromAppHeadingCss = css`
+	${textSans.large({ fontWeight: 'bold' })}
+	color:${palette.brand[500]};
+	line-height: 1.35;
```
jamesmockett, line 82: The typography API has a `lineHeight` option you could use here and _looks_ like `regular` maps to 1.35 which is the value you're after:
https://guardian.github.io/csnx/?path=/docs/source-foundations_typography--headline#options

client/components/mma/switch/SwitchOptions.tsx
```diff
@@ -76,6 +76,14 @@ const buttonStuckCss = css`
...
+	border: 0;
```
jamesmockett, line 83: Is this needed? There shouldn't be a border on the `h2` to start with or is it picking it up from somewhere else?

client/components/mma/switch/SwitchOptions.tsx
```diff
@@ -122,6 +130,24 @@ const SwitchOptions = () => {
...
+						news app
+					</h2>
+					<p
+						css={css`
+							${textSans.medium()}
+						`}
+					>
+						To unlock unlimited reading in our news app, please make
+						a small change to your support type. If this doesn't
+						suit you, no change is needed, but not you will continue
```
jamesmockett, line 146:
```suggestion
						suit you, no change is needed, but note you will continue
```

client/components/mma/switch/SwitchOptions.tsx
andrewHEguardian: Good catch, had this as a CSS override on <Heading> at first, but you can't remove the border from that

client/components/mma/switch/SwitchOptions.tsx
andrewHEguardian: 🤦 grauniad

client/components/mma/switch/SwitchOptions.tsx
andrewHEguardian: Thanks!

client/components/mma/switch/SwitchOptions.tsx
jamesmockett: 😂

jamesmockett (approved):
Nice work 👍

PR #950, opened 2023-01-06
AB TEST -  hide the default onboarding trial newsletter

vlbee:
## What does this change?

We a running a one month trial of a soft opt in default onboarding newsletter in the new registration onboarding flow.
https://github.com/guardian/gateway/pull/2105
 

Becasue this is a trial, we need to hide the newsletter from appearing in MMA. 
Users can still unsubscribe from the trial newsletter via the email footer, or by hitting unscubscribe ALL in MMA

This will be reverted when trial is over.

### DIFF
--- a/client/components/mma/identity/identity.ts
+++ b/client/components/mma/identity/identity.ts
@@ -113,7 +113,14 @@ export const ConsentOptions: ConsentOptionCollection = {
 		return await RemoveSubscriptionsAPI.execute();
 	},
 	newsletters(options: ConsentOption[]): ConsentOption[] {
-		return options.filter(isNewsletter);
+		return (
+			options
+				.filter(isNewsletter)
+				// @AB_TEST: Default Onboarding Newsletter Test: START
+				// Prevent trial newsletter from displaying.
+				.filter((newsletter: ConsentOption) => newsletter.id !== '6028') // identityId: 'saturday-roundup-trial'
+			// @AB_TEST: Default Onboarding Newsletter Test: END
+		);
 	},
 	consents(options: ConsentOption[]): ConsentOption[] {
 		return options.filter(isConsent);


client/components/mma/identity/identity.ts
```diff
@@ -113,7 +113,14 @@ export const ConsentOptions: ConsentOptionCollection = {
 		return await RemoveSubscriptionsAPI.execute();
 	},
 	newsletters(options: ConsentOption[]): ConsentOption[] {
-		return options.filter(isNewsletter);
+		return (
+			options
+				.filter(isNewsletter)
+				// @AB_TEST: Default Onboarding Newsletter Test: START
+				// Delete this line when trial is over
+				.filter((newsletter: ConsentOption) => newsletter.id !== 'XXXX') // TODO
```
dblatcher, line 121: It would be clearer which newsletter this is if we use the `newsletter.identityName` (unique kebab-cased name) for the filter instead of the `newsletter.id`. eg the test could be something like `newsletter.identityName ===  'default-onboarding-test-newsletter'`

We'll be able to get the actual value for the identityName when we add the new newsletter to the list. I'll reach out about doing that today

dblatcher:
```suggestion
				.filter((newsletter: ConsentOption) => newsletter.identityName !== 'saturday-roundup-trial') 
```

vlbee: I'd prefer to use the numerical id becuase the identityName is an optional field in the code as this is also used for consents.  

```
export interface ConsentOption {
	id: string;
	description?: string;
	frequency?: string;
	name: string;
	theme?: string;
	type: ConsentOptionType;
	subscribed: boolean;
	identityName?: string;
	isProduct?: boolean;
	isChannel?: boolean;
}
```

dblatcher: 👍  thanks for talking me through that!

dblatcher (approved):

PR #947, opened 2023-01-04
Refactor zuora properties

jonflynng:
Refactors the names of API response properties to be more in-line with Zuora's naming conventions (specifically the conventions in their REST API). That is, we are changing the property `amount` to be named `price`, and the property `interval` to be named `billingPeriod`. Refer to some examples of Zuora rest API responses [here](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/test/resources/GuardianWeekly/CappedPriceIncrease/Subscription.json) and [here](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/test/resources/GuardianWeekly/CappedPriceIncrease/Catalogue.json).

PR to change on the backend is [here](https://github.com/guardian/members-data-api/pull/872).

PR #951, opened 2023-01-06
Refactor zuora properties - remove old properties

jonflynng:
As mentioned in the [previous PR](https://github.com/guardian/manage-frontend/pull/947). This PR removes the old properties.

 The best way to review this PR would be to do a diff of this branch and the main branch before my last PR, and assert that the property names have changed accordingly.

I have also added a new Cypress test for the update contribution amount journey:

Screenshot 2023-01-06 at 14 48 14

### DIFF
--- a/client/components/mma/billing/Billing.tsx
+++ b/client/components/mma/billing/Billing.tsx
@@ -151,13 +151,10 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
 											productUrlPart:
 												specificProductType.urlPart,
 										}));
-									const resultsPerPage = (
-										paidPlan.billingPeriod ||
-										paidPlan.interval ||
-										''
-									)?.includes('year')
-										? productInvoiceData.length
-										: 6;
+									const resultsPerPage =
+										paidPlan.billingPeriod?.includes('year')
+											? productInvoiceData.length
+											: 6;
 									return (
 										<Fragment
 											key={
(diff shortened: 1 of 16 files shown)


client/components/mma/billing/Billing.tsx
```diff
@@ -151,13 +151,10 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
...
 										}));
-									const resultsPerPage = (
-										paidPlan.billingPeriod ||
-										paidPlan.interval ||
-										''
-									)?.includes('year')
-										? productInvoiceData.length
-										: 6;
+									const resultsPerPage =
+										paidPlan.billingPeriod?.includes('year')
```
jonflynng, line 155: not sure this looks right, ill investigate

client/components/mma/billing/Billing.tsx
```diff
@@ -151,13 +151,10 @@ const BillingRenderer = ([mdaResponse, invoiceResponse]: [
 											productUrlPart:
 												specificProductType.urlPart,
 										}));
-									const resultsPerPage = (
-										paidPlan.billingPeriod ||
-										paidPlan.interval ||
-										''
-									)?.includes('year')
```
jonflynng, line 158: not sure why this is optional. I've checked the main branch before my refactors and it seems it always has been, but it is required on the type interface.

jonflynng:
[Rename object properties to be in-line with Zuora's naming](https://trello.com/c/ZfJMciLT/3737-rename-object-properties-to-be-in-line-with-zuoras-naming)

client/components/mma/billing/Billing.tsx
andrewHEguardian: No idea why either, seems fine to leave as it was for now

PR #953, opened 2023-01-06
Create Functional Async Loading component

andrewHEguardian:
## What does this change?

This PR adds two new approaches to replace the old class component AsyncLoader. One is a hook `useAsyncLoader` and the other is a functional component `loadingComponent` that uses said hook.

### Motivation: 

AsyncLoader is used throughout MMA as a way to obtain asynchronous data before rendering a component that requires that data. It still uses classes rather than hooks. It tries to handle a number of edge cases while trying to use TypeScript for type safety. It introduces a level of indirection that makes it hard to use and maintain, and doesn't flow well with modern React syntax and best practices.

## Future steps

- Class asyncLoader based components are going to be updated to the new functional loading component (6 month target on that)
- New components should be functional and use the AsyncLoader hook pattern rather than the LoadingComponent


client/components/mma/shared/asyncComponents/DefaultApiResponseHandler.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const JsonResponseHandler: ResponseProcessor = (response: Response) => {
+	return handleAResponses(response, (r: Response) => r.json());
+};
+
+export const TextResponseHandler: ResponseProcessor = (response: Response) => {
+	return handleAResponses(response, (r: Response) => r.text());
+};
+
+function handleAResponses(
```
rupertbates, line 11:
```suggestion
function handleAResponse(
```

andrewHEguardian: Definite typo. But need to take another look at this and decide whether response can be an array of responses

andrewHEguardian:
[Investigate refactoring/replacing/deprecating/removing AsyncLoader](https://trello.com/c/K292fN9v/3732-investigate-refactoring-replacing-deprecating-removing-asyncloader)

client/components/mma/shared/asyncComponents/DefaultApiResponseHandler.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const TextResponseHandler: ResponseProcessor = (
+	response: Response | Response[],
+) => {
+	return handleResponses(response, (r: Response) => r.text());
+};
+
+function handleResponses(
+	response: Response | Response[],
+	transformResponse: any,
```
rupertbates, line 17: This could be type `Response => any` couldn't it?

client/components/mma/shared/asyncComponents/DefaultApiResponseHandler.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const TextResponseHandler: ResponseProcessor = (
+	response: Response | Response[],
+) => {
+	return handleResponses(response, (r: Response) => r.text());
+};
+
+function handleResponses(
+	response: Response | Response[],
```
rupertbates, line 16: Is there any better way to handle this so we don't have to cater to a response or an array of responses? It seems quite confusing, for instance the function name is `handleResponses` but the parameter is `response` . Could the calling code handle mapping an array onto individual calls to this function where needed?

andrewHEguardian: We're catering for a response or an array of responses to be able to swap out existing uses of asyncLoader which can use either. I think this is the right place to do map the array. Any thoughts @KentGrm? Would it make more sense if the parameter was named `responses`?

rupertbates: Ok, it's not a big deal, just prefer to avoid having to check `isArray` throughout the code but it's only in two places I guess.

rupertbates (approved):
Really nice work 👍

client/components/mma/shared/asyncComponents/useAsyncLoader.ts
```diff
@@ -0,0 +1,37 @@
+import * as Sentry from '@sentry/browser';
+import { useEffect, useState } from 'react';
+import { trackEvent } from '../../../../utilities/analytics';
+import type ResponseProcessor from './ResponseProcessor';
+
+export default function useAsyncLoader<T>(
```
jamesmockett, line 6: As this is a hook it should probably live in `/utilities/hooks` with the others rather than in `/components`.

client/components/mma/shared/asyncComponents/LoadingComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+	LoadingView: () => React.ReactElement;
+	LoadedView: (data: any) => React.ReactElement;
+	ErrorView: () => React.ReactElement;
+};
+
+export function LoadingComponent({
+	asyncFetch,
+	responseProcessor,
+	LoadingView,
+	LoadedView,
```
KentGrm, line 16: I still don't like this variable - too similar to LoadingView... but this is a hard thing to name.

rupertbates (approved):

PR #957, opened 2023-01-10
Enable `eslint-comments/require-description` rule and add comments

andrewHEguardian:
## What does this change?

Removes override for `eslint-comments/require-description` and adds comments where they are missing from `eslint-disable` comments


andrewHEguardian:

PR #958, opened 2023-01-11
Remove `import/no-default-export` rule override, excluding storybook, and fix exports and imports

andrewHEguardian:
## What does this change?

This PR removes `import/no-default-export` override from linting rule, but adds it back just for storybook stories which require default exports, and fixes all the places where default exports and imports were used.

This rule is inherited from standard Guardian linting config, but is slightly subjective, as some people may prefer default exports. However it enforces some consistency.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,6 +23,12 @@ module.exports = {
 				'@typescript-eslint/ban-ts-comment': ['off'],
 			},
 		},
+		{
+			files: ['client/**/*.stories.tsx'],
+			rules: {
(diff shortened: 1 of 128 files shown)


andrewHEguardian:

jamesmockett (approved):
I think this is good for consistency 👍 It's one of those PRs that's tricky to thoroughly review due to the number of changes, but I don't see any glaring issues. Might be worth deploying to `CODE` if you haven't already done so, and rebasing from `main` before deploying (if there have been any merges since you last did so).

PR #959, opened 2023-01-12
Enable rule `'@typescript-eslint/unbound-method'`

andrewHEguardian:
## What does this change?

Removes override for  `'@typescript-eslint/unbound-method'` rule and suppress the error in the one place it failed, AsyncLoader, which should be deprecated soon 🤞.


jamesmockett (approved):
I'm looking forward to the demise of `AsyncLoader`, but until then this is a good fix 👍

andrewHEguardian:

PR #964, opened 2023-01-20
Change useAsyncLoader to use an enum to track state

andrewHEguardian:
Change promiseFromAsync fetch to be a callback instead of a promise

Co-Authored-By: Kent McClymont <111352918+KentGrm@users.noreply.github.com>

## What does this change?

Changes useAsyncLoader to use an enum to track state. Also change the fetch to be a callback instead of a promise to stop it firing indefinitely.


jamesmockett (approved):
This is a nice little improvement 🚀

PR #966, opened 2023-01-23
MMA: Randomise Cancellation Reasons

charleycampbell:
## What does this change?
When a customer wants to cancel the cancellation reasons are in a fixed order meaning that there could be bias to the options at the top of the list. This change shuffles the cancellation reasons whilst keeping 'another reason' fixed at the bottom.

## How to test
Checked on storybook and CODE that when trying to cancel the reasons provided are shuffled differently each time, but 'Other Reasons' always remains at the bottom of the list. On Prod when on the cancel recurring contributions page the reasons should be in a random order with another reason at the bottom. Refresh the same page and the reasons should be shuffled with another reason remaining at the bottom of the list.

## Images

Before image

After
image

After refresh
image

## Accessibility

### DIFF
--- a/client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
+++ b/client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
@@ -104,9 +104,19 @@ export const contributionsCancellationReasons: CancellationReason[] = [
 		linkLabel: 'I no longer read the Guardian enough to justify my support',
 		alternateFeedbackIntro: standardAlternateFeedbackIntro,
 	},
+];
+
+export const otherCancellationReason: CancellationReason[] = [
 	{
 		reasonId: 'mma_other',
 		linkLabel: 'Another reason (please specify)',
 		alternateFeedbackIntro: standardAlternateFeedbackIntro,
 	},
 ];
+
+const shuffledArray = contributionsCancellationReasons.sort(
+	() => 0.5 - Math.random(),
+);
+
+export const shuffledContributionsCancellationReasons: CancellationReason[] =
+	shuffledArray.concat(otherCancellationReason);
--- a/client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
+++ b/client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
@@ -114,9 +114,11 @@ export const otherCancellationReason: CancellationReason[] = [
 	},
 ];
 
-const shuffledArray = contributionsCancellationReasons.sort(
+const shuffledArray = [...contributionsCancellationReasons].sort(
 	() => 0.5 - Math.random(),
 );
 
-export const shuffledContributionsCancellationReasons: CancellationReason[] =
-	shuffledArray.concat(otherCancellationReason);
+export const shuffledContributionsCancellationReasons: CancellationReason[] = [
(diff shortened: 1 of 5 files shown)


twrichards:
this is a cool idea - nice!

might be worth augmenting the events...

- salesforce case tracking (if still there)
- GA events (if still there)
- ophan events (if still there, might be the wrong name for them)

... with the index/position of the reason they chose (to help prove the hypothesis that there is a bias for the first one/few)

client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
```diff
@@ -104,9 +104,19 @@ export const contributionsCancellationReasons: CancellationReason[] = [
...
 		alternateFeedbackIntro: standardAlternateFeedbackIntro,
 	},
 ];
+
+const shuffledArray = contributionsCancellationReasons.sort(
+	() => 0.5 - Math.random(),
+);
+
+export const shuffledContributionsCancellationReasons: CancellationReason[] =
+	shuffledArray.concat(otherCancellationReason);
```
twrichards, line 122: could be done with a spread operation, ideally with the value of `otherCancellationReason` inlined (if not reused, and if reused then it needn't be wrapped in array), something like... 
```suggestion
	[...shuffledArray , otherCancellationReason];
```

charleycampbell: Great suggestion!

andrewHEguardian (approved):
Looks good! Will be very interesting to see the results of this

client/components/mma/cancel/contributions/ContributionsCancellationReasons.tsx
```diff
@@ -104,9 +104,21 @@ export const contributionsCancellationReasons: CancellationReason[] = [
...
 	{
 		reasonId: 'mma_other',
 		linkLabel: 'Another reason (please specify)',
 		alternateFeedbackIntro: standardAlternateFeedbackIntro,
 	},
 ];
+
+const shuffledArray = [...contributionsCancellationReasons].sort(
+	() => 0.5 - Math.random(),
+);
```
johnduffell, line 119: this produces quite a biased distribution of ordering, e.g. the first element in the initial list is more than twice as likely to remain first than the second element
See discussion here:
https://github.com/guardian/manage-frontend/pull/1415/files#r1846461396

client/components/mma/cancel/supporterplus/SupporterplusCancellationReasons.tsx
```diff
@@ -84,9 +84,21 @@ export const supporterplusCancellationReasons: CancellationReason[] = [
...
 	{
 		reasonId: 'mma_other',
 		linkLabel: 'Another reason (please specify)',
 		alternateFeedbackIntro: standardAlternateFeedbackIntro,
 	},
 ];
+
+const shuffledArray = [...supporterplusCancellationReasons].sort(
+	() => 0.5 - Math.random(),
+);
```
johnduffell, line 99: as above this produces quite a biased distribution of ordering, e.g. the first element in the initial list is more than twice as likely to remain first than the second element
See discussion here:
https://github.com/guardian/manage-frontend/pull/1415/files#r1846461396

PR #969, opened 2023-01-26
Add feature switch for product switching flow

jamesmockett:
## What does this change?

Adds a feature switch to disable product switching, preventing access to the work in progress pages on production.

## How to test

Navigation to `/switch` or any of the child pages will not be possible.


andrewHEguardian (approved):
Good addition, I think it's fine to leave the routes out if the feature switch isn't there. At least for now

PR #972, opened 2023-01-30
Replace test Stripe setup intent with more obviously dummy data

andrewHEguardian:
## What does this change?

Changes the stripe setup intent used for tests with a more obviously dummy string.


rupertbates (approved):

andrewHEguardian:
The reason the string is in that form is that stripe JS is validating it:
image

PR #975, opened 2023-02-02
Improve Cypress cancellation spec reliability

jamesmockett:
## What does this change?

Updates cancellation tests to reference radio buttons by name rather than index to improve reliability.

Cancellation reasons are now randomised so if the reason radio button is selected by index a different reason will be selected for each test run. Not all tests are explicitly testing specific cancellation reasons, but the ones that are will randomly fail due to an incorrect option being selected. (Some cancellation reasons add additional steps or UI elements, changing the flow through the cancellation process.)

This change has previously been made to _some_ tests, but not all. This PR updates the remaining tests to ensure consistent test runs.

Selecting radio buttons by index:

```js
cy.findAllByRole('radio').eq(2).click();
```

Selecting radio buttons by name:

```js
cy.findAllByRole('radio', {
    name: 'I am unhappy with some editorial decisions',
}).click();
```

### DIFF
--- a/cypress/integration/parallel-2/cancelContribution.spec.ts
+++ b/cypress/integration/parallel-2/cancelContribution.spec.ts
@@ -106,7 +106,9 @@ describe('Cancel contribution', () => {
 		}).as('get_case');
 
 		setupCancellation();
-		cy.findAllByRole('radio').eq(0).click();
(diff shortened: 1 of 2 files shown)


charleycampbell (approved):
Good spot! Thank you for picking this up 👍

PR #980, opened 2023-02-03
Continue Reading The Guardian component

charleycampbell:
If a user has gone through the switch journey on desktop or via a mobile but **not** using the app, we want to show a 'continue reading the guardian' button to direct them back to the homepage of The Guardian.

## What does this change?
A new button has been added to the switch complete screen which when clicked directs the user to the home page of The Guardian. The button should only be visible on desktop/mobile  **if**  user hasn't come via the app

## How to test
The new button should appear in the Switch Complete Storybook test. Since this remains behind a feature switch it can't be tested in production but on CODE (if the feature flag has been lifted) follow the product switch journey and you will see the button on the switch complete page.

## Images

From testing in CODE:
image

## Accessibility

-   [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)

### DIFF
--- a/client/components/mma/switch/SwitchComplete.tsx
+++ b/client/components/mma/switch/SwitchComplete.tsx
@@ -8,12 +8,14 @@ import {
 	textSans,
 } from '@guardian/source-foundations';
 import {
+	Button,
 	Stack,
 	SvgClock,
 	SvgEnvelope,
 } from '@guardian/source-react-components';
 import { useContext } from 'react';
 import { Navigate, useLocation } from 'react-router';
+import { useNavigate } from 'react-router-dom';
 import type { PaidSubscriptionPlan } from '../../../../shared/productResponse';
 import { getMainPlan } from '../../../../shared/productResponse';
 import { calculateMonthlyOrAnnualFromBillingPeriod } from '../../../../shared/productTypes';
@@ -30,6 +32,8 @@ import { iconListCss, sectionSpacing } from './SwitchStyles';
 export const SwitchComplete = () => {
 	const switchContext = useContext(SwitchContext) as SwitchContextInterface;
 	const productDetail = switchContext.productDetail;
+	const navigate = useNavigate();
+
 	const mainPlan = getMainPlan(
 		productDetail.subscription,
 	) as PaidSubscriptionPlan;
@@ -75,6 +79,16 @@ export const SwitchComplete = () => {
 			{!switchContext.isFromApp && (
 				<section css={sectionSpacing}>
 					<SignInBanner />
+					<Button
+						css={css`
+							margin-top: ${space[6]}px;
+						`}
+						onClick={() => {
+							navigate('https://www.theguardian.com/uk');
+						}}
+					>
+						Continue to read the Guardian
+					</Button>
 				</section>
 			)}
 		</>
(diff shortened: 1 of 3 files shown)


client/components/mma/switch/SwitchReview.tsx
```diff
@@ -127,8 +127,8 @@ export const SwitchReview = () => {
 
 	const confirmSwitch = async (amount: number) => {
 		try {
-			setIsSwitching(true);
-			const response = await productMoveFetch(false);
+			setIsSwitching(false);
```
andrewHEguardian:
```suggestion
			setIsSwitching(true);
```
you switched the wrong one back :)

client/components/mma/switch/SwitchReview.tsx
```diff
@@ -127,8 +127,8 @@ export const SwitchReview = () => {
...
+			const response = await productMoveFetch(true);
```
andrewHEguardian:
```suggestion
			const response = await productMoveFetch(false);
```

andrewHEguardian (approved):
Nicely done 🚀

client/components/mma/switch/SwitchComplete.tsx
```diff
@@ -75,6 +76,14 @@ export const SwitchComplete = () => {
 			{!switchContext.isFromApp && (
 				<section css={sectionSpacing}>
 					<SignInBanner />
+					<LinkButton
+						css={css`
```
jamesmockett, line 80: This should be `cssOverrides` rather than `css`?

In general it's good to avoid adding style overrides to components if you can though. (Sometimes it _is_ the only way to do it though.)

It might be nicer to wrap the banner and button components with a `<Stack>` which will apply spacing between the components automatically:

```jsx
<Stack space={6}>
    <SignInBanner />
    <div>
        <LinkButton ...>
    </div>
</Stack>
```

(I've added a `<div>` around the button as `<Stack>` makes the width of all of the child elements 100% which you probably don't want for the button.)

charleycampbell: Ahh ok! Nice suggestions - I've changed it. I'll just chuck it on to CODE and double check it's working as expected 👀

charleycampbell: Tested successful in CODE

charleycampbell:
[Thank you page: NON APP ONLY: Continue Reading The Guardian component](https://trello.com/c/y6ZqEkhy/3810-thank-you-page-non-app-only-continue-reading-the-guardian-component)

client/components/mma/switch/SwitchComplete.tsx
```diff
@@ -74,7 +75,19 @@ export const SwitchComplete = () => {
...
-					<SignInBanner />
+					<Stack space={6}>
+						<SignInBanner />
+						<div>
+							<LinkButton
+								cssOverrides={css`
+									margin-top: ${space[6]}px;
+								`}
+								href="https://www.theguardian.com/"
+							>
```
jamesmockett, line 86: Sorry, might have confused things with my last comment! 😅 The `<Stack>` is applying spacing between the banner and button now so you can remove the override for `margin-top` completely.

```suggestion
							<LinkButton href="https://www.theguardian.com/">
```

charleycampbell: oh right I thought we needed it as you suggested changing the css to cssOverride, but that make sense if Stack is doing the work

PR #983, opened 2023-02-06
Return to app banner

jamesmockett:
## What does this change?

Adds a 'thank you' banner specifically for users who began the switching journey in the app. The banner includes a CTA to return users to the app so they can continue reading.

This is primarily for mobile users, although users on tablets may also go through the same switching journey via the app so this includes an interim desktop design which may be refined later.

## How to test

Start a switching journey from `/app/switch`. The banner will appear at the top of the final page.

## Images

Screenshot 2023-02-07 at 10 18 59

Screenshot 2023-02-07 at 10 19 14

Screenshot 2023-02-07 at 10 19 27

### DIFF
--- a/client/components/mma/switch/SwitchComplete.tsx
+++ b/client/components/mma/switch/SwitchComplete.tsx
@@ -5,6 +5,7 @@ import {
 	palette,
 	space,
 	textSans,
+	until,
 } from '@guardian/source-foundations';
 import {
 	Button,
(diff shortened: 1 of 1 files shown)


jamesmockett:
[Thank You Page - APP ONLY: Return to App Component](https://trello.com/c/6X8cySb0/3806-thank-you-page-app-only-return-to-app-component)

charleycampbell (approved):
Looks good to me 👍

PR #984, opened 2023-02-07
Add threshold per currency config and use new config in Switch pages

andrewHEguardian:
- refactor contributionsAmount since it is similar
- add Storybooks for variations

## What does this change?

Adds a config file for Supporter Plus benefit thresholds and uses it in switch pages. Also refactor ContributionsAmountOptions (similar config) to sit in the same folder. Fixes some errors where the pages were not using dynamic monthly/annual copy and adds storybooks with currency and yearly config to catch any regressions.

## How to test

Go to `/switch` with a contribution using a currency other than GBP and observe a different price point.

## Accessibility

### DIFF
--- a/client/components/mma/accountoverview/ContributionUpdateAmountForm.tsx
+++ b/client/components/mma/accountoverview/ContributionUpdateAmountForm.tsx
@@ -17,6 +17,11 @@ import type { PaidSubscriptionPlan } from '../../../../shared/productResponse';
 import { augmentBillingPeriod } from '../../../../shared/productResponse';
 import type { ProductType } from '../../../../shared/productTypes';
 import { trackEvent } from '../../../utilities/analytics';
+import type {
+	ContributionInterval} from '../../../utilities/contributionsAmount';
+import {
+	contributionAmountsLookup
+} from '../../../utilities/contributionsAmount';
 import { fetchWithDefaultParameters } from '../../../utilities/fetch';
 import { AsyncLoader } from '../shared/AsyncLoader';
 import { Button } from '../shared/Buttons';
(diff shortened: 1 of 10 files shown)


client/components/mma/accountoverview/ContributionUpdateAmountForm.tsx
```diff
@@ -34,127 +39,8 @@ interface ContributionUpdateAmountFormProps {
...
-	string,
-	{
-		month: ContributionAmountOptions;
-		year: ContributionAmountOptions;
-	}
->;
-
 class UpdateAmountLoader extends AsyncLoader<string> {}
 
-// TODO: make this dynamic (i.e. looks up api/config file agreed/shared by contributions team)
```
andrewHEguardian, line 56: I don't really know what this comment was suggesting so I think it's safe to remove it?

andrewHEguardian:
[Product Switching: Dynamic Ts and Cs price thresholds](https://trello.com/c/NWUqEnjv/3842-product-switching-dynamic-ts-and-cs-price-thresholds)

client/components/mma/accountoverview/ContributionUpdateAmountForm.tsx
graham228221: I think this is related to this idea to consolidate thresholds for products into one place: https://trello.com/c/2n9CYEEA/3800-spike-consolidate-thresholds-for-product-holdings-in-one-place currently these are separately hardcoded in 5 or 6 (or more) places

graham228221: Oh I see! Perhaps these could also be incorporated in a single place along with the thresholds.

andrewHEguardian: Added as a comment to that Trello ticket 👍

david-pepper:
conscious that we are adding yet another place that these thresholds live (alongside Salesforce and new-product-api)

Should we have a session to discuss where best to maintain the threshold values (S3 maybe?) and in what format, so that all systems that need to use them can access them when needed?

client/utilities/benefitsThreshold.ts
```diff
@@ -0,0 +1,45 @@
+import type { CurrencyIso } from './currencyIso';
+
+export const benefitsThresholdsByCountryGroup: Record<
```
jamesmockett, line 3: Would it make sense to combine the benefit thresholds and contribution amounts into a single object so the config is in a single place? (Both have monthly and annual amounts for the same list of countries.)

jamesmockett: Referring to David's previous comment, this is probably best informed by any discussion around storing these values in a central location.

andrewHEguardian:
> Should we have a session to discuss where best to maintain the threshold values (S3 maybe?) and in what format, so that all systems that need to use them can access them when needed?

Definitely, that would be valuable

client/utilities/benefitsThreshold.ts
andrewHEguardian: I thought about that - but given it's still somewhat TBD it's probably better not to combine them before we know what we want to extract to a central location.

charleycampbell (approved):
Looks good to me! I know there has been some conversation around the storing of the thresholds and there is a ticket so happy to approve to get this out 🚀

PR #987, opened 2023-02-09
Enable Product Switching feature flag

charleycampbell:
## What does this change?

Enables the Product Switching feature flag.

## How to test

`/switch` should be accessible on all environments, including production.

We're enabling this to allow beta testing of apps metering so the pages are not linked to from anywhere in MMA. It's possible somebody could stumble upon this URL and go through the switching journey, but this isn't a huge problem if it happens and is a risk we've accepted.


andrewHEguardian (approved):
Once we've finished testing we should remember to remove the code that checks the feature flag

jamesmockett:
[Product Switching: Release to PROD](https://trello.com/c/gFLSw9lD/3851-product-switching-release-to-prod)

PR #989, opened 2023-02-10
Fix product switching Cypress spec

jamesmockett:
## What does this change?

Updates Cypress spec to reflect latest copy changes. This was missed as the tests are only run when the feature switch is enabled.


andrewHEguardian (approved):
maybe slightly brittle test but good catch 👍

PR #994, opened 2023-02-16
Product switching payment method component

jamesmockett:
## What does this change?

Adds a new payment method component for product switching. The existing payment method components were unsuitable due to design differences and the requirement that they should be displayed inline with the surrounding text. (PayPal is a special case as we also need to truncate overly long email addresses.)

Refactoring the existing components for use here would have added unnecessary complexity, hence building a new version. As the design of MMA is updated we can hopefully standardise of a way to display payment details across the site.

### Desktop

Screenshot 2023-02-16 at 15 04 56
Screenshot 2023-02-16 at 15 05 07
Screenshot 2023-02-16 at 15 05 14
Screenshot 2023-02-16 at 15 05 21

### Mobile medium

Screenshot 2023-02-16 at 15 06 33
Screenshot 2023-02-16 at 15 06 43
Screenshot 2023-02-16 at 15 06 51
Screenshot 2023-02-16 at 15 07 00

### Mobile

Screenshot 2023-02-16 at 15 05 37
Screenshot 2023-02-16 at 15 05 46
Screenshot 2023-02-16 at 15 05 59
Screenshot 2023-02-16 at 15 06 10


jamesmockett:
[Product Switching: Update payment method components](https://trello.com/c/BlwN62YD/3877-product-switching-update-payment-method-components)

client/components/mma/switch/SwitchReview.stories.tsx
```diff
@@ -27,12 +31,52 @@ export default {
 export const Default: ComponentStory<typeof SwitchReview> = () => (
 	<SwitchReview />
 );
+// We're not setting `state` in the component-level parameters above due to how
```
andrewHEguardian, line 34: I'm torn on this one - not sure this comment will help us if we run into this issue in a different place. It seems like the problem is more with how we check for one payment method vs another. The data structure implies they are not mutually exclusive, while in practice they are. That said, the `subscription.card && <>foo</>` is a very neat little way to check it exists.

andrewHEguardian (approved):
Looks really good! 🥇

client/components/mma/switch/SwitchReview.stories.tsx
jamesmockett: The comment was more to explain how parameters are merged in stories which is not necessarily obvious and has caused confusion elsewhere. In this case it manifests itself as multiple payment methods being shown due to how we check the payment method, but, as you say, this wouldn't happen in practice.

I was hesitant to switch to using `paymentMethod` instead as I'm not sure how reliable that is as it doesn't seem to be used elsewhere in the codebase. It would be a neater solution than checking for properties associated with specific payment methods though.

PR #995, opened 2023-02-16
Make amendments to copy

andrewHEguardian:
## What does this change?

Makes the following amendments to copy:

  | Page | Copy | Should be
-- | -- | -- | --
All scenarios | Account Overview | Good Afternoon Reader | Good afternoon, supporter
  |   | Thanks for supporting... | Thank you for funding the Guardian since [month] [year]
  | Change your support / Review change | view benefits > | View extras >
  |   | hide benefits > | Hide extras >
  |   | Dive in and start enjoying your exclusive extras straight away | In just a couple of steps, you'll be able to start enjoying your exclusive extras
  | Thank you page | Remove full stops except for the banner |  
Above Threshold Specific | Change your support / Your current support | Add extras with no extra cost | Add extras
  |   | Your current payment entitles you to | In exchange for your current payment, you can choose to receive exclusive supporter extras ...
  |   | These exclusive supporter extras are unlocked for a minimum monthly support of £10 and a minimum annual support of £95. | These exclusive supporter extras are available when you pay £10 minimum on a monthly basis, or £95 minimum on an annual basis.
Below Threshold specific | Your current support | Upgrade | Upgrade to £X per month/year

## How to test

Navigate to each page in the above table, in the scenarios listed and witness the 'should be' copy.

## Images

Account Overview
image

Your current support
Below threshold
image
Above threshold
image

£0 edgecase copy
image
 
image

## Accessibility

### DIFF
--- a/client/components/mma/accountoverview/PersonalisedHeader.tsx
+++ b/client/components/mma/accountoverview/PersonalisedHeader.tsx
@@ -16,9 +16,8 @@ function calculateTimeOfDay() {
 	}
 	if (currentHour < 18) {
 		return 'Good afternoon';
-	} else {
-		return 'Good evening';
 	}
+	return 'Good evening';
 }
 
 export const PersonalisedHeader = ({
--- a/client/components/mma/accountoverview/PersonalisedHeader.tsx
+++ b/client/components/mma/accountoverview/PersonalisedHeader.tsx
@@ -1,10 +1,24 @@
 import { css } from '@emotion/react';
 import { headline, space } from '@guardian/source-foundations';
+import { dateString } from '../../../../shared/dates';
 import type { MembersDataApiResponse } from '../../../../shared/productResponse';
 import { isProduct, sortByJoinDate } from '../../../../shared/productResponse';
 
 interface PersonalisedHeaderProps {
 	mdapiResponse: MembersDataApiResponse;
+}
+
+function calculateTimeOfDay() {
+	const currentHour = new Date().getHours();
+
+	if (currentHour < 12) {
+		return 'Good morning';
+	}
+	if (currentHour < 18) {
+		return 'Good afternoon';
+	} else {
+		return 'Good evening';
+	}
 }
 
 export const PersonalisedHeader = ({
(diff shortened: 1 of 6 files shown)


andrewHEguardian:
[Product Switching: Copy Amends (Iona)](https://trello.com/c/AxhqvZxt/3882-product-switching-copy-amends-iona)

client/components/mma/switch/SwitchComplete.tsx
```diff
@@ -59,14 +59,14 @@ export const SwitchComplete = () => {
 	const amountPayableToday = routerState?.amountPayableToday;
 	const nextPaymentDate = routerState?.nextPaymentDate;
 
-	if (!amountPayableToday || !nextPaymentDate) {
+	if (amountPayableToday === undefined || !nextPaymentDate) {
```
jamesmockett, line 62: Great catch that this would redirect you if the amount payable is zero 👍

client/components/mma/accountoverview/PersonalisedHeader.tsx
```diff
@@ -1,12 +1,26 @@
...
 import { headline, space } from '@guardian/source-foundations';
+import { dateString } from '../../../../shared/dates';
 import type { MembersDataApiResponse } from '../../../../shared/productResponse';
 import { isProduct, sortByJoinDate } from '../../../../shared/productResponse';
 
 interface PersonalisedHeaderProps {
 	mdapiResponse: MembersDataApiResponse;
 }
 
+function calculateTimeOfDay() {
```
jamesmockett, line 11: Should we leave this as an arrow function for consistency? (I realise we're not _entirely_ consistent with this elsewhere…)

client/components/mma/accountoverview/PersonalisedHeader.tsx
```diff
@@ -1,12 +1,26 @@
...
+	const currentHour = new Date().getHours();
+
+	if (currentHour < 12) {
+		return 'Good morning';
+	}
+	if (currentHour < 18) {
+		return 'Good afternoon';
+	} else {
```
jamesmockett, line 19: I'm pretty sure you pointed this out before, but you don't really need an `else` here.

client/components/mma/accountoverview/PersonalisedHeader.tsx
andrewHEguardian: When I was pairing with @KentGrm he pointed out that if you define the function as an arrow function inside the component, every time the component is re-rendered the function will be defined again

andrewHEguardian: See also this SO answer. I don't think it's a big performance hit at all but might be a better practice going forward. https://stackoverflow.com/a/46138370

KentGrm: useCallback is useful if you want to access variables in the render/component function. But this is a code smell and edgecase

andrewHEguardian: Yeah not flagging the useCallback part of that answer just this part:

> In most cases you should declare the functions outside the component function so you declare them only once and always reuse the same reference. When you declare the function inside, every time the component is rendered the function will be defined again.

jamesmockett: Ah, that's an interesting point about functions being defined every time a component is rendered. In this case the function is no longer inside a component so it doesn't matter if it's an arrow function? (But it's fine either way!) This is an unexpected, but interesting, chat in a PR that's ostensibly about copy changes 😀

jamesmockett (approved):
Nice work 👍  Hopefully we can stop worrying about full stops now… 🤞

PR #996, opened 2023-02-17
Add routing and navigation for Saved Articles page behind feature switch

vlbee:
## Accessibility


shared/featureSwitches.ts
```diff
@@ -27,4 +27,5 @@ export const featureSwitches: Record<string, boolean> = {
 	cancellationProductSwitch: false,
 	accountOverviewNewLayout: false,
 	productSwitching: true,
+	savedArticles: false,
```
lindseydew, line 30: Nice :)

cypress/integration/parallel-6/savedArticles.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { featureSwitches } from '../../../shared/featureSwitches';
+
+if (!featureSwitches.savedArticles) {
```
lindseydew, line 3: Think it would be good to have a mechanism which toggles the feature switch state on / off actually in the test rather than reading what's in defined in the code base, so we can check both states at once

cypress/integration/parallel-6/savedArticles.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+		beforeEach(() => {
+			cy.session('auth', () => {
+				cy.setCookie('gu-cmp-disabled', 'true');
+			});
+		});
+
+		it('displays Saved Article page', () => {
+			cy.visit('/saved-articles');
+
+			cy.findAllByText('Saved articles').should('have.length', 2);
```
lindseydew, line 37: Small point, but might be tempted to be more general here and say length greater than 0 for an existence check.

cypress/integration/parallel-6/savedArticles.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+		});
+	});
+}
```
lindseydew, line 40: Is it worth having a test on the main nav to check the existence of the Saved Article list?

lindseydew: Oh wait I see we are testing this, all good 👍

cypress/integration/parallel-6/savedArticles.spec.ts
vlbee: I thought the length check would do that as well as indicate that the tab appears on the title AND LH navigation. I saw you added a comment in your PR which we can merge in 👍

cypress/integration/parallel-6/savedArticles.spec.ts
vlbee: I've [updated `featureSwitches.ts`](https://github.com/guardian/manage-frontend/pull/996/commits/c152b9607a4cc313b9db7d6259cc8f8448b370a8) to maintain Cypress environment switch states static where required.

client/components/shared/nav/NavConfig.tsx
```diff
@@ -20,13 +21,15 @@ export interface NavItem {
...
 	isDropDownExclusive?: boolean;
+	isExcludedByFeatureSwitch?: boolean;
 }
 
 interface NavLinks {
 	accountOverview: MenuSpecificNavItem;
 	billing: MenuSpecificNavItem;
 	profile: MenuSpecificNavItem;
 	settings: MenuSpecificNavItem;
+	savedArticles: MenuSpecificNavItem;
```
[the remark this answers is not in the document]
vlbee: I'm not sure if this is a better solution becuase I think you'd need to conditionally check for the presence of `savedArticles` in other places using the `NavLinks` such as `MMAPageSkeleton` and the new `SavedArticlesComponent`. Given it will be a hopefully temporary exclusion I think it's clearer to have the code doing the work in one file with a explicit nameing to explain what is is doing rather than returning undefined.  Maybe I am thinking about it wrong though? Happy to pair and review.

shared/featureSwitches.ts
```diff
@@ -27,4 +33,27 @@ export const featureSwitches: Record<string, boolean> = {
...
+		Cypress: unknown;
+	}
+}
+
+const cypressSwitchState: Record<string, boolean> = {
+	...featureSwitches,
+	savedArticles: false,
+};
+export const getFeatureSwitches = (): Record<string, boolean> => {
+	// If running within cypress, return the forced switches to keep tests consistent regardless of production state
```
lindseydew, line 55: Ah I see, so `window.Cypress` is written to when the tests run. 

Could potentially write the actual switch state on the `window.Cypress` and read this here to return the switches, rather than hardcoding `cypressSwitchState` above. There's an example of that here: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/cypress/e2e/parallel-1/sign-in-gate.cy.js#L141

Then we can be explicit in the test that `savedArticles` is being set to false.

lindseydew: Would be good to validate with someone with a more expert JS knowledge than me whether the use of a global object here is ok

client/components/shared/nav/NavConfig.tsx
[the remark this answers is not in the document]
lindseydew: Sure that makes sense. Looks like it's the same approach being taken with the other boolean field too

shared/featureSwitches.ts
vlbee: For review context, I was basing this solution on the [dotcom solution here](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/experiments/cypress-switches.ts)

shared/featureSwitches.ts
```diff
@@ -27,4 +33,27 @@ export const featureSwitches: Record<string, boolean> = {
...
 	productSwitching: true,
+	savedArticles: false,
+};
+
+// Cypress Testing:
+// We want to be able to write tests independent of switch state that test both states.
+// Use getFeatureSwitches to retrieve switches dependant on clientside environment.
+// For example you can set a switch to false
+// and opt in in Cypress via query search params in cy.visit(url).
+declare global {
```
andrewHEguardian, line 44: I think this could go in `globals.ts` where global is also declared

shared/featureSwitches.ts
```diff
@@ -27,4 +33,27 @@ export const featureSwitches: Record<string, boolean> = {
...
+}
+
+const cypressSwitchState: Record<string, boolean> = {
+	...featureSwitches,
+	savedArticles: false,
+};
+export const getFeatureSwitches = (): Record<string, boolean> => {
+	// If running within cypress, return the forced switches to keep tests consistent regardless of production state
+	if (window !== undefined && window.Cypress) return cypressSwitchState;
+	// Otherwise just return empty object
```
andrewHEguardian, line 57: We're not returning an empty object here

shared/featureSwitches.ts
lindseydew: Ah I see, yeah makes sense. I think my point is perhaps a bit of overkill.

PR #998, opened 2023-02-20
Refactor product switch pages to use common setup in context

andrewHEguardian:
## What does this change?

Refactor the product switching pages that use the same setup by moving that setup into the context from SwitchContainer instead of repeating the calculations on every page. Move each page into its own folder.

## Accessibility


client/components/mma/switch/SwitchContainer.tsx
```diff
@@ -30,6 +36,17 @@ export interface SwitchContextInterface {
...
+	monthlyOrAnnual: 'Monthly' | 'Annual';
+	supporterPlusTitle: string;
+	thresholds: Thresholds;
+}
+
+export interface Thresholds {
+	monthlyThreshold: number;
+	annualThreshold: number;
+	chosenThreshold: number;
+	aboveThreshold: boolean;
```
jamesmockett, line 49: Minor detail, but I wonder if it would be clearer if this property was renamed to distinguish it from the other 3 similarly named properties which are all threshold amounts? `isAboveThreshold` or similar?

andrewHEguardian: isAboveThreshold is better 👍 not sure 'chosenThreshold' makes that much sense since a user can't choose between monthly or annual in this flow

jamesmockett: It's essentially the threshold for the billing period, so maybe something like `billingPeriodThreshold`? (Not sure if that makes any more sense 🤷‍♂️ )

client/components/mma/switch/SwitchContainer.tsx
```diff
@@ -114,17 +131,53 @@ const RenderedPage = (props: {
...
 			>
 				<Outlet />
 			</SwitchContext.Provider>
 		</SwitchPageContainer>
 	);
 };
+
+function getThresholds(
+	mainPlan: PaidSubscriptionPlan,
+	monthlyOrAnnual: string,
```
jamesmockett, line 162: This might be better as a boolean as that's all it's being used for. (eg. `monthly: boolean`) The actual value of the string isn't being used for anything else in here.

client/components/mma/switch/SwitchContainer.tsx
jamesmockett: Or maybe we don't even need it as it's only used in a single place? That neatly avoids having to come up with a good name 😅

andrewHEguardian: is it? I thought that threshold was used in a number of places 🤔

jamesmockett: Only seems to be one reference to it in this PR in `SwitchComplete`.

jamesmockett (approved):
This is a really nice update and a good stepping stone to breaking these pages down further which will really help with future updates and maintenance 🎉

PR #1000, opened 2023-02-20
Remove redundant usage of  mdapiResponseReader

charleycampbell:
## Background
Work was done to add user details to the MMA response. For the work to be shipped successfully we had to enable the client side to accept the previous response (a JSON array of products) and the new response which contains the user

This ticket is to tidy up the code base to remove the old response type and do any refactoring which may be required.

## What does this change?
Removed mdapiResponseReader from productResponse which then cascaded into a refactor to remove where it was called and only pass in the new MembersDataApiResponse object. 

Some of the tests needed to be refactored to used the Test Utils object which is the new response object and irrelevant tests have been removed since it tested the mdapiResponseReader.

## How to test
On `local` or `CODE` make sure all of the screens and process are working as expected. We hope that there is no change in behaviour or errors in the logs.


i-hardy:
[Mobile optimisation: variant A build for global AB test: build](https://trello.com/c/zdqCSfuk/1080-mobile-optimisation-variant-a-build-for-global-ab-test-build)

client/__tests__/components/productResponse.test.tsx
```diff
@@ -11,25 +10,6 @@ const membersDataApiResponse = {
...
-	});
-
-	it('throws validation error if MDAPI Response is in incorrect format', () => {
-		const mdaResponse = '';
-		expect(() => mdapiResponseReader(mdaResponse as any)).toThrow(
-			'MDAPI Response is of invalid format',
-		);
-	});
+it('mdapiResponseReader returns mdaResponse when MembersDataApiResponse is passed in', () => {
+	expect(membersDataApiResponse).toEqual(membersDataApiResponse);
```
andrewHEguardian, line 14: I don't see any reason to keep this test

client/components/mma/cancel/CancellationContainer.tsx
```diff
@@ -33,9 +31,8 @@ import type {
 } from './productSwitch/productSwitchApi';
 
 const renderSingleProductOrReturnToAccountOverview =
-	(productType: ProductType) =>
-	(data: [MembersDataApiResponse | MembersDataApiItem[]]) => {
-		const mdaResponse = mdapiResponseReader(data);
+	(productType: ProductType) => (data: MembersDataApiResponse) => {
+		const mdaResponse = data;
```
andrewHEguardian, line 35: Could we name this `mdapiResponse` to keep it consistent?

client/components/mma/switch/review/SwitchReview.stories.tsx
```diff
@@ -18,6 +18,7 @@ export default {
 	parameters: {
 		layout: 'fullscreen',
 		reactRouter: {
+			state: { productDetail: contributionPayPal },
```
andrewHEguardian, line 21: I'm not sure why this is here?

andrewHEguardian: image

The storybook for Switch Review pages looks a bit weird now with two payment methods showing

jamesmockett: Yes, that `state` shouldn't be there as it's being set at the individual story level due to merging of state causing the above issue. (There's a lengthy comment a bit further down in the file explaining this.)

cypress/integration/parallel-2/updateContributionAmount.spec.ts
```diff
@@ -21,12 +24,12 @@ describe('Update contribution amount', () => {
 
 		cy.intercept('GET', '/api/me/mma?productType=Contribution', {
 			statusCode: 200,
-			body: [contributionCard],
+			body: toMembersDataApiResponse(contributionCard),
```
jamesmockett, line 27: Would it be better to update the fixtures to include the `user` and `products` properties rather than having to pipe them through a function to transform them?

client/components/mma/switch/review/SwitchReview.stories.tsx
charleycampbell: Great spot..I'm not sure how this snuck in here

cypress/integration/parallel-2/updateContributionAmount.spec.ts
charleycampbell: Interesting idea! I'm not sure how we would implement that as in productDetails.ts the products are all of type ProductDetails, so a const is needed to pass in the product into the MdapiResponse. Could you show me an example of what you mean? If it's a larger bit of refactoring to change all of the products I'm reluctant to do it part of this change as this should cover only what the refactor impacts

client/components/mma/delivery/address/DeliveryAddressChangeContainer.tsx
```diff
@@ -38,13 +34,11 @@ interface ContextAndOutletContainerProps {
 
 const renderContextAndOutletContainer =
 	(productType: ProductType) =>
-	(allProductDetails: [MembersDataApiResponse | MembersDataApiItem[]]) => {
-		const mdaResponse = mdapiResponseReader(allProductDetails);
-
+	(allProductDetails: MembersDataApiResponse) => {
```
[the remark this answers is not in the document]
charleycampbell: Good call

client/components/mma/cancel/CancellationContainer.tsx
charleycampbell: Good call - I've made it consistent through other classes too

client/__tests__/components/productResponse.test.tsx
charleycampbell: Deleted - although it's handy to keep in mind we don't have much code coverage for lower level functions

cypress/integration/parallel-2/updateContributionAmount.spec.ts
andrewHEguardian: I think this would make sense, but agree maybe a separate PR

andrewHEguardian (approved):
Very nicely done 👍

cypress/integration/parallel-2/updateContributionAmount.spec.ts
jamesmockett: I think it should be done as the format of the fixtures no longer match what the API returns so they always have to be piped through `toMembersDataApiResponse`, but yeah, it could be a separate PR. We were going to go through the fixtures to update things like `subscriberId` with clearly mocked data, so perhaps it would make sense to do this at the same time?

charleycampbell: As discussed offline line I'll raise a ticket to cover reworking fixtures

PR #1001, opened 2023-02-20
Add components for different states of the API response

lindseydew:
## What does this change?

This PR adds the basic components we'll need for rendering the Save for Later page. Since this feature is behind a switch, this is intended as a work in progress to give us the files we need to start iterating on the project. We still need to flesh out what the API request and response will be, which is why this is just testing mocking a call to an external service for now.  

## How to test

We should check that there is no visibility of these pages in PROD as is it behind a switch. 

## Accessibility

### DIFF
--- a/client/__tests__/components/savedarticles/savedArticlesPage.test.tsx
+++ b/client/__tests__/components/savedarticles/savedArticlesPage.test.tsx
@@ -66,7 +66,8 @@ describe('SavedArticlesPage', () => {
 		});
 	});
 
-	it('should display an error message if the API call is successful but the json response is not expected', async () => {
+	// eslint-disable-next-line jest/no-disabled-tests -- Add this test in when we have validated the response type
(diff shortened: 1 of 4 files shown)


client/components/mma/savedArticles/SavedArticles.tsx
```diff
@@ -1,13 +1,61 @@
...
+			/>
 		</PageContainer>
 	);
 };
+
+interface SavedArticlesPageProps {
+	saveForLaterAPICall: () => Promise<unknown>;
+}
+
+export function SavedArticlesPage(props: SavedArticlesPageProps) {
```
vlbee, line 36: Wonder if it's worth looking at the AsyncLoader utility used by the AccountOverview? There seem to be two async loaders in use in this repo - useAsyncLoader hook and the AsyncLoader class compononent. I would check with the rest of the MMA team which component is preferred and see if we can align our work with that.

PR #1003, opened 2023-02-21
Remove product switch code from main bundle

jamesmockett:
## What does this change?

Adds code splitting for product switch pages so they are dynamically imported as needed and not included in the main bundle. This matches the other sections of the site and ensure users are not downloading code that they do not need. This is especially important at the moment as product switching is only partially rolled out for beta app users, but all visitors to the site are having to download the code for this feature.

|Before|Before (gzip)|After|After (gzip)|
|---|---|---|---|
|818.32KB|243.23 KB|737.6 KB|213.78 KB|

## How to test

Manually navigate to the product switch page at `/switch` with dev tools open and you will see the product switch bundle being dynamically loaded.

Screenshot 2023-02-21 at 16 53 41

### DIFF
--- a/client/components/mma/MMAPage.tsx
+++ b/client/components/mma/MMAPage.tsx
@@ -39,10 +39,6 @@ import { Main } from '../shared/Main';
 import { DeliveryAddressUpdate } from './delivery/address/DeliveryAddressForm';
 import { Maintenance } from './maintenance/Maintenance';
 import { MMAPageSkeleton } from './MMAPageSkeleton';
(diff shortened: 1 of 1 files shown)


jamesmockett:
[Add code splitting for product switch pages](https://trello.com/c/wfOD6vfi/3893-add-code-splitting-for-product-switch-pages)

andrewHEguardian (approved):
Nice! Didn't really think about this when we created the pages, good spot 👍

PR #1008, opened 2023-02-23
Only call the `get-cancellation-date` API endpoint when necessary

jonflynng:
## What does this change?

 As part of the process to cancel a subscription, we make an API call to get a preview of the cancellation date. This is intended to provide the customer with an estimated timeframe for when their subscription will end, but it is only applicable for Guardian Weekly and Voucher products. We do not use this date for any other product in the cancellation process. This PR makes this API call conditionally based on the `startPageOfferEffectiveDateOptions `.

 Cypress tests have been amended to check the API is called (or not called) as expected.

Tested in DEV.

### DIFF
--- a/client/components/mma/cancel/CancellationReasonSelection.tsx
+++ b/client/components/mma/cancel/CancellationReasonSelection.tsx
@@ -329,17 +329,10 @@ const ReasonPickerWithCancellationDate = ({
 		return <GenericErrorScreen />;
 	}
 
-	// offer choice if not trial period or lead time, and startPageOfferEffectiveDateOptions config set to true
-	const shouldOfferEffectiveDateOptions =
-		!isNaN(
-			Date.parse(cancellationDateResponse.cancellationEffectiveDate),
-		) && productType.cancellation.startPageOfferEffectiveDateOptions;
-
-	const chargedThroughDateStr = shouldOfferEffectiveDateOptions
-		? parseDate(cancellationDateResponse.cancellationEffectiveDate).dateStr(
-				DATE_FNS_LONG_OUTPUT_FORMAT,
-		  )
-		: undefined;
+	// offer choice if not trial period or lead time
+	const chargedThroughDateStr = parseDate(
+		cancellationDateResponse.cancellationEffectiveDate,
+	).dateStr(DATE_FNS_LONG_OUTPUT_FORMAT);
 
 	return (
 		<ReasonPicker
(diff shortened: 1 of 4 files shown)


client/components/mma/cancel/CancellationReasonSelection.tsx
```diff
@@ -294,47 +291,95 @@ const ReasonPicker = (props: ReasonPickerProps) => {
...
+		shouldOfferEffectiveDateOptions &&
+		parseDate(cancellationDateResponse.cancellationEffectiveDate).dateStr(
+			DATE_FNS_LONG_OUTPUT_FORMAT,
+		);
+
+	return (
+		<ReasonPicker
+			productType={productType}
+			productDetail={productDetail}
+			chargedThroughDateStr={chargedThroughDateStr || undefined}
```
andrewHEguardian, line 348: would it be more readable if this was included on line 338 as 
`const chargedThroughDateStr = shouldOfferEffectiveDateOptions ? parseDate(...) : undefined;`

jonflynng: yep

andrewHEguardian (approved):
Another nice asyncLoader refactor 👍 checking the number of api calls in the cypress tests is a good idea

client/components/mma/cancel/CancellationReasonSelection.tsx
```diff
@@ -294,47 +291,95 @@ const ReasonPicker = (props: ReasonPickerProps) => {
...
+
+	if (cancellationDateResponse === null) {
+		return <GenericErrorScreen />;
+	}
+
+	// offer choice if not trial period or lead time, and startPageOfferEffectiveDateOptions config set to true
+	const shouldOfferEffectiveDateOptions =
+		!isNaN(
+			Date.parse(cancellationDateResponse.cancellationEffectiveDate),
+		) && productType.cancellation.startPageOfferEffectiveDateOptions;
```
jamesmockett, line 336: You don't really need this check now as this component is only rendered if `startPageOfferEffectiveDateOptions` is true.

jonflynng: amended

jamesmockett (approved):
This is really nice ❤️  It's so much easier to follow with `AsyncLoader` removed. (The fact it fixes a bug and removes an unnecessary API call is a bonus!)

PR #1010, opened 2023-02-28
Account Overview product banners

jamesmockett:
## What does this change?

Adds support for product banners on the _Account Overview_ page. The images themselves are hosted on Grid so this PR is relatively small. The codebase already includes some Grid image components borrowed from [`support-frontend`](https://github.com/guardian/support-frontend), but for the sake of simplicity these have not been used here as we don't require responsive image support. (It should be noted that these components are not used in any public facing features so could potentially be removed.)

_Note:_ this is a draft PR as designs for the banner images are still being finalised

## Images

Screenshot 2023-02-28 at 15 52 00

PR #1016, opened 2023-03-07
Show cancelled products on new Account Overview layout

jamesmockett:
## What does this change?

Adds cancellation detail summaries to the existing product cards (to cover products which are within the cancellation period and still accessible) as well as a redesigned cancelled product card that matches the new Account Overview card design.

The product details card has been renamed from `AccountOverviewCardV2` to `ProductCard` to match the naming of the other new product card components and to reflect that it will be the canonical product details card component when this feature is released.

The placeholder banner image code and styles have been removed. A simplified implementation is covered by #1010 which is awaiting final designs.

## How to test

With the `accountOverviewNewLayout` feature flag enabled, log into an account that has cancelled products. A cancellation detail summary should be shown above the cancelled product card as per the screenshots below. Storybook has also been updated with example of cancelled products.

## Images

Screenshot 2023-03-07 at 16 33 17

Screenshot 2023-03-07 at 16 34 04

Screenshot 2023-03-07 at 16 34 13

Screenshot 2023-03-07 at 16 34 24

### DIFF
--- a/client/components/mma/accountoverview/CancelledProductCard.tsx
+++ b/client/components/mma/accountoverview/CancelledProductCard.tsx
@@ -6,10 +6,11 @@ import {
 	space,
 	textSans,
 } from '@guardian/source-foundations';
-import { Stack } from '@guardian/source-react-components';
+import { LinkButton, Stack } from '@guardian/source-react-components';
 import { parseDate } from '../../../../shared/dates';
 import type { CancelledProductDetail } from '../../../../shared/productResponse';
 import { GROUPED_PRODUCT_TYPES } from '../../../../shared/productTypes';
+import { trackEvent } from '../../../utilities/analytics';
 import { InfoSummary } from '../paymentUpdate/Summary';
 import { Card } from '../shared/Card';
 import { productCardConfiguration } from './ProductCardConfiguration';
@@ -25,6 +26,10 @@ export const CancelledProductCard = ({
 
 	const cardConfig =
 		productCardConfiguration[specificProductType.productType];
+
+	const showSubscribeAgainButton =
+		productDetail.mmaCategory !== 'membership' &&
+		productDetail.mmaCategory !== 'recurringSupport';
 
 	const sectionHeadingCss = css`
 		${textSans.medium({ fontWeight: 'bold' })};
@@ -156,7 +161,25 @@ export const CancelledProductCard = ({
 							</dl>
 						</div>
 						<div css={buttonLayoutCss}>
-							{/* TODO: Resubscribe button */}
+							{showSubscribeAgainButton && (
+								<LinkButton
+									href="https://support.theguardian.com/uk/subscribe"
+									size="small"
+									cssOverrides={css`
+										justify-content: center;
+									`}
+									priority="primary"
+									onClick={() => {
+										trackEvent({
+											eventCategory: 'href',
+											eventAction: 'click',
+											eventLabel: 'subscribe_again',
+										});
+									}}
+								>
+									Subscribe again
+								</LinkButton>
+							)}
 						</div>
 					</div>
 				</Card.Section>
--- a/client/components/mma/accountoverview/CancelledProductCard.tsx
+++ b/client/components/mma/accountoverview/CancelledProductCard.tsx
@@ -1,0 +1,166 @@
+import { css } from '@emotion/react';
+import {
+	from,
+	headline,
+	palette,
+	space,
+	textSans,
+} from '@guardian/source-foundations';
+import { Stack } from '@guardian/source-react-components';
+import { parseDate } from '../../../../shared/dates';
+import type { CancelledProductDetail } from '../../../../shared/productResponse';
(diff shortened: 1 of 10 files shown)


jamesmockett:
[MMA: Account Overview Edgecase: Cancelled product messaging](https://trello.com/c/ByBMIGFX/3901-mma-account-overview-edgecase-cancelled-product-messaging)

client/fixtures/cancelledProductDetail.ts
```diff
@@ -0,0 +1,29 @@
...
+	mmaCategory: 'recurringSupport',
+	tier: 'Contributor',
+	joinDate: '2022-01-05',
+	subscription: {
+		subscriptionId: 'A-S00303370',
+		cancellationEffectiveDate: '2023-03-20',
+		start: '2021-12-10',
+		end: '2022-11-29',
+		readerType: 'Direct',
+		accountId: '8ad09f8a7e25bda3017e296317464818',
```
charleycampbell, line 13: Just double checking these are mock/test account ids?

shared/dates.ts
```diff
@@ -5,10 +5,13 @@ export const DATE_FNS_INPUT_FORMAT = 'yyyy-MM-dd'; // example: 1969-07-16
 export const DATE_FNS_LONG_OUTPUT_FORMAT = 'd MMMM yyyy'; // example: 1 July 2021
 export const DATE_FNS_SHORT_OUTPUT_FORMAT = 'd MMM yyyy'; // example: 5 Jan 2019
 
-export const cancellationFormatDate = (cancellationEffectiveDate?: string) => {
+export const cancellationFormatDate = (
+	cancellationEffectiveDate?: string,
+	outputFormat: string = DATE_FNS_SHORT_OUTPUT_FORMAT,
```
charleycampbell, line 10: Just curious chat does FNS stand for?

client/components/mma/accountoverview/ProductCardConfiguration.ts
```diff
@@ -0,0 +1,43 @@
...
+	},
+	supporterplus: {
+		headerColor: palette.brand[500],
+		showBenefitsSection: true,
+	},
+	digipack: {
+		headerColor: palette.brand[500],
+	},
+	digitalvoucher: {
+		headerColor: '#ff5943',
```
charleycampbell, line 23: Could there be a work done to put these hex codes into the palette library? Would be good to have the brand colours all in one place...not sure who that work would sit with though

client/components/mma/accountoverview/CancelledProductCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+					minHeightTablet
+				>
+					<h3 css={productTitleCss}>
+						{specificProductType.productTitle()}
+					</h3>
+				</Card.Header>
+				<Card.Section>
+					<div css={productDetailLayoutCss}>
+						<div>
+							<h4 css={sectionHeadingCss}>Billing and payment</h4>
```
charleycampbell, line 116: Minor but just wondering whether the **p** on **payment** should be capitalised?

charleycampbell (commented):
Just a few minor comments but otherwise looks good! 👍

shared/dates.ts
jamesmockett: `date-fns` is the name of the date formatting library being used. I don't know why that needs to be in the name of these constants, but I'm simply reusing what was already there so haven't renamed then.

client/components/mma/accountoverview/ProductCardConfiguration.ts
jamesmockett: If these colours are unique to MMA then they shouldn't be part of the Source colour palette. We could maybe add constants for these, but I've left them alone for the moment as Helene is still working on the final colour palette. Also, these colours aren't directly related to this PR, I've just extracted the config from the `ProductCard` component to tidy things up slightly as there's a lot going on in there!

client/components/mma/accountoverview/CancelledProductCard.tsx
jamesmockett: It's correct as per the designs which are using sentence case for the other headings.

client/components/mma/accountoverview/ProductCardConfiguration.ts
charleycampbell: That's a good point re the Source colour palette which I didn't know about!

andrewHEguardian (approved):
New design for cancelled products is really nice. Also glad to see the "V2" gone 👍

client/fixtures/cancelledProductDetail.ts
jamesmockett: I've updated the `subscriptionId` and `accountId` fields so that they're obviously mocked data to avoid any confusion in the future. I'll raise a separate PR to do the same for the existing fixtures. (Those will probably be slightly more work due to tests relying on specific account IDs being exposed in the UI.)

andrewHEguardian (approved):

PR #1017, opened 2023-03-09
Check if app subscription is iOS or Android from productId

andrewHEguardian:
## What does this change?

Adds a function to determine if an app subscription is Android or iOS based on rules from a mapping table (found [here](https://docs.google.com/spreadsheets/d/1I8XZnirRkIQ1RDAq9iSqVC5B9UzFqV4iJnC_dTfydfE/edit#gid=796953305)). Shows different copy on the Billing page depending on which app store the subscription comes from. If it is a puzzle subscription it falls back to default.

## How to test

See the Billing storybook or visit the Billing page with an in app subscription.

The naming convention of the productIds may change at some point - though not for a while. There is a fallback in place if we can not determine the origin from productId.

## Images

image

## Accessibility

### DIFF
--- a/shared/mpapiResponse.ts
+++ b/shared/mpapiResponse.ts
@@ -5,6 +5,7 @@ export interface AppSubscription {
 	cancellationTimestamp?: string;
 	from: string;
 	valid: boolean;
+	productId: string;
 }
 
 export interface MPAPIResponse {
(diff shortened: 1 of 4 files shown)


shared/mpapiResponse.ts
```diff
@@ -20,3 +21,22 @@ export const AppSubscriptionSoftOptInIds: string[] = [
...
+		return AppStore.ANDROID;
+	} else if (
+		subscription.productId.includes('guardian.gia') ||
+		subscription.productId.includes('guardian.gla')
+	) {
+		return AppStore.IOS;
+	} else {
+		return AppStore.UNKNOWN;
+	}
+}
```
jamesmockett, line 42: It's a stylistic preference, but each condition has a `return` so you don't really need `else`. (I think it might look a little cleaner though.)

jamesmockett (approved):
Looks good. Nice little improvement 👍

PR #1024, opened 2023-03-16
Prevent multiple calls to `product-move` API

jamesmockett:
## What does this change?

Prevents the `product-move` API being called again if the switch 'confirm' button is clicked more than once, and hides the switch button on the _Account Overview_ if the product has a pending cancellation. (This shouldn't happen in practice for a contribution, but we have a fixture that covers this scenario so this is a bit belt and braces.)

Screenshot 2023-03-16 at 16 13 54

## How to test

With the _Network_ tab in devtools open, click the 'confirm' button on the switch review page multiple times and check that the `product-move` API is only called once.

### DIFF
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -130,6 +130,7 @@ export const ProductCard = ({
 
 	const showSwitchButton =
 		isEligibleToSwitch &&
+		!hasCancellationPending &&
 		specificProductType.productType === 'contributions';
 
 	const productBenefits =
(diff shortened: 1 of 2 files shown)


jamesmockett:
[MMA: Prevent multiple calls to `product-move` API](https://trello.com/c/YC1AXgHn/3977-mma-prevent-multiple-calls-to-product-move-api)

client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -130,6 +130,7 @@ export const ProductCard = ({
 
 	const showSwitchButton =
 		isEligibleToSwitch &&
+		!hasCancellationPending &&
```
andrewHEguardian, line 133: unrelated to this PR, but just realised the switch button text is `Change to monthly + extras`, which should probably be dynamic?

jamesmockett: You're right! 😱 I'll fix that in a separate PR…

PR #1025, opened 2023-03-20
Inline 'price rise' alert and dynamic product name on switch button

jamesmockett:
## What does this change?

A couple of small tweaks and fixes ahead of enabling the new Account Overview layout: the existing price rise alert component has been replaced as it breaks the layout and now appears inline with the price, and the 'Change to x' button has been updated so that the product name is dynamic rather than hardcoded as 'monthly + extras'.

## Images

Screenshot 2023-03-17 at 17 54 57

### DIFF
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -4,6 +4,7 @@ import {
 	Button,
 	buttonThemeReaderRevenueBrand,
 	Stack,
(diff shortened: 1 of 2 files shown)


jamesmockett:
[MMA / Account Overview: Price rise alert](https://trello.com/c/pjP1Qmim/3692-mma-account-overview-price-rise-alert)

andrewHEguardian (approved):
👍

PR #1026, opened 2023-03-21
Add Quantum Metric script

jamesmockett:
## What does this change?

Dynamically loads Quantum Metric script if user has consented via the consent banner.

As we were using an old version of the `@guardian/consent-management-platform` package this was not aware of QM so this has been upgraded (along with `@guardian/libs`) so we can test for QM consent.

_Note:_ The QM script is currently restricted to capturing data from the product switch pages so this PR only includes masking of PII on those rather than the whole of MMA.

## How to test

If you have previously consented to cookies you will be able to see the QM script in the `head` element:

Screenshot 2023-03-21 at 10 20 00

If you clear the consent cookies the consent banner will be shown again. If you then select _Manage my cookies_ and disable tracking the QM script should not be present. You can also see the script being added when the banner is shown. Initially it won't be present as consent hasn't been given, but if you select _Yes, I'm happy_ you will see it added to the `head` element as before.

Session replays for the product switch journey will be available in the Quantum Metric dashboard.

## Images

Screenshot 2023-03-21 at 10 24 05


jamesmockett:
[MMA / Quantum Metric: Check for consent and dynamically load script](https://trello.com/c/ME0zmTrr/3982-mma-quantum-metric-check-for-consent-and-dynamically-load-script)

jamesmockett:
[MMA / Quantum Metric: Check for consent and dynamically load script](https://trello.com/c/ME0zmTrr/3982-mma-quantum-metric-check-for-consent-and-dynamically-load-script)

andrewHEguardian (approved):
Looks good 👍

PR #1029, opened 2023-03-22
Enable new Account Overview layout and product switching from MMA

jamesmockett:
## What does this change?

Enables the new Account Overview layout and allows product switches to be initiated from MMA.

## How to test

When signed in with an active recurring contribution, but no digital subscription, you should should see a _Change to monthly + extras_ button in the product card. (_annual_ if an annual contribution.)

All other subscription information shown in the old layout should be displayed in the new product cards, and it should be possible to manage each subscription and update payment details as before.

Apart from the addition of the _Change to monthly + extras_ button and a new look and feel, users should experience no change in functionality so we should not expect any additional contact with customer support as a result of this change. We should also monitor Sentry and ensure there are no new errors being logged.

## Before

manage thegulocal com_ (3)

manage thegulocal com_ (2)

## After

manage thegulocal com_ (4)

manage thegulocal com_

### DIFF
--- a/client/components/mma/accountoverview/AccountOverview.stories.tsx
+++ b/client/components/mma/accountoverview/AccountOverview.stories.tsx
@@ -26,6 +26,8 @@ import {
 import { user } from '../../../fixtures/user';
 import { AccountOverview } from './AccountOverview';
 
+featureSwitches['appSubscriptions'] = true;
+
 export default {
 	title: 'Pages/AccountOverview',
 	component: AccountOverview,
@@ -66,11 +68,9 @@ export const WithSubscriptions: ComponentStory<typeof AccountOverview> = () => {
 	return <AccountOverview />;
 };
 
-export const WithContributionNewLayout: ComponentStory<
+export const WithContributionAndSwitchPossible: ComponentStory<
 	typeof AccountOverview
 > = () => {
-	featureSwitches['accountOverviewNewLayout'] = true;
-
 	fetchMock
 		.restore()
 		.get('/api/cancelled/', { body: [] })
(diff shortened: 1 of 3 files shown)


jamesmockett:
[MMA: Enable new Account Overview layout and product switching](https://trello.com/c/nsMoEP2F/3998-mma-enable-new-account-overview-layout-and-product-switching)

client/components/mma/accountoverview/AccountOverview.stories.tsx
```diff
@@ -26,6 +26,8 @@ import {
 import { user } from '../../../fixtures/user';
 import { AccountOverview } from './AccountOverview';
 
+featureSwitches['appSubscriptions'] = true;
```
andrewHEguardian, line 29: why do we need this line? isn't it on by default now?

client/components/mma/accountoverview/AccountOverview.stories.tsx
```diff
@@ -109,11 +108,9 @@ export const WithContributionNewLayoutPaymentFailure: ComponentStory<
 	return <AccountOverview />;
 };
 
-export const WithContributionNewLayoutDigisubAndContribution: ComponentStory<
+export const WithContributionAndDigisub: ComponentStory<
```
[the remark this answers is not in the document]
jamesmockett: That's a good suggestion 👍

client/components/mma/accountoverview/AccountOverview.stories.tsx
andrewHEguardian: oh yes I misread it 🙈

andrewHEguardian (approved):
🥳  woo!

PR #1033, opened 2023-03-23
Use React Router for local links in dropdown nav

jamesmockett:
## What does this change?

Updates the dropdown navigation component to use React Router `<Link>` elements so that client-side routing is used. (External links continue to use regular anchor elements.)

The dropdown navigation is the only navigation available on mobile devices so mobile users are unnecessarily seeing full page refreshes as they navigate around the site.

## Images

Screenshot 2023-03-24 at 11 12 45


charleycampbell (approved):
Looks good to me 🚀

PR #1044, opened 2023-04-06
Allow multiple Recurring Contributions in Product Switch Journey

andrewHEguardian:
## What does this change?

Removes a check that a user only has one Recurring Contribution during Product Switching. There is a ticket in Conversions team to allow multiple contributions to be checked out - so we should not assume this will be the case downstream.

Also fixes an issue picked up where the RC productDetail selected for switching was not being passed through in routerState (meaning MDAPI was being called on the second step), so we are now passing through the existing state to every page.

## How to test

Take out multiple RCs (under the threshold amount) and product switch one of them via the account overview - the second switch page should load and not redirect to the Account Overview. 

Cypress tests have been updated to include a check of how many times MDAPI is called, on a full switch journey when coming directly to the URL or by clicking to switch a specific contribution.

## Accessibility

### DIFF
--- a/client/components/mma/switch/options/SwitchOptions.tsx
+++ b/client/components/mma/switch/options/SwitchOptions.tsx
@@ -7,13 +7,16 @@ import {
 } from '@guardian/source-react-components';
 import { ErrorSummary } from '@guardian/source-react-components-development-kitchen';
 import { useContext, useEffect, useRef, useState } from 'react';
-import { useNavigate } from 'react-router';
+import { useLocation, useNavigate } from 'react-router';
 import { Link } from 'react-router-dom';
 import { formatAmount } from '../../../../utilities/utils';
 import { Card } from '../../shared/Card';
 import { Heading } from '../../shared/Heading';
 import { SupporterPlusBenefitsSection } from '../../shared/SupporterPlusBenefits';
-import type { SwitchContextInterface } from '.././SwitchContainer';
+import type {
+	SwitchContextInterface,
+	SwitchRouterState,
+} from '.././SwitchContainer';
 import { SwitchContext } from '.././SwitchContainer';
 import {
 	buttonCentredCss,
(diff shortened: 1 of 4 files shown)


charleycampbell (approved):

andrewHEguardian:
[Product Switching - Two RCs Bug](https://trello.com/c/gXd78Rka/4041-product-switching-two-rcs-bug)

cypress/integration/parallel-2/productSwitch.spec.ts
```diff
@@ -270,6 +297,8 @@ describe('product switching', () => {
 			cy.findByText(
 				/Your first billing date is today and you will be charged £5/,
 			).should('exist');
+
+			cy.get('@mdapi_get_contribution.all').should('have.length', 1);
```
charleycampbell, line 301: Is it worth adding a test where there is more than one?

charleycampbell (commented):
Only one comment about adding in a test! But other than that it looks good to me 🚀

cypress/integration/parallel-2/productSwitch.spec.ts
andrewHEguardian: I actually meant to do this, but put it in the wrong setup

cypress/integration/parallel-2/productSwitch.spec.ts
```diff
@@ -221,11 +225,31 @@ describe('product switching', () => {
 		featureSwitches.accountOverviewNewLayout &&
 		featureSwitches.productSwitching
 	) {
-		it('navigates to product switching page from Account Overview', () => {
+		it('successfully completes product switching page from Account Overview', () => {
```
KentGrm, line 228: Can you add a test to demonstrate a single contribution and multiple contributions, selecting something other than the first.

client/components/mma/switch/options/SwitchOptions.tsx
```diff
@@ -68,6 +71,8 @@ const fromAppHeadingCss = css`
 
 export const SwitchOptions = () => {
 	const switchContext = useContext(SwitchContext) as SwitchContextInterface;
+	const location = useLocation();
+	const routerState = location.state as SwitchRouterState;
```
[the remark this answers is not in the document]
andrewHEguardian: the state is being validated at each step in `SwitchContainer` I would argue, it's merely being accessed here in order to pass it through again (where SwitchContainer will look at it)

PR #1054, opened 2023-04-20
Reminder creation page

tomrf1:


server/routes/api.ts
```diff
@@ -274,7 +274,7 @@ router.get('/help-centre/topic/:topic', getTopicHandler);
 
 router.post('/contact-us', contactUsFormHandler);
 
-router.post('/reminders', createReminderHandler);
+router.post('/reminders/create', createReminderHandler);
```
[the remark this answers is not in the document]
tomrf1: actually, I was wrong about this...

client/components/mma/cancel/cancellationContributionReminder.tsx
```diff
@@ -38,7 +38,7 @@ interface ReminderChoice {
 	thankYouMessage: string;
 }
 
-const REMINDER_ENDPOINT = '/api/reminders';
+const REMINDER_ENDPOINT = '/api/reminders/create';
```
tomrf1, line 41: for clarity I've renamed this endpoint for the existing "winback" reminder signup. We now have `/create` and `/create-public`

tomrf1: This existing one requires a user to be signed in

client/components/mma/identity/idapi/supportReminders.ts
```diff
@@ -69,3 +74,21 @@ const sendReminderReactivation = (reminderCode: string) =>
 		},
 		body: JSON.stringify({ reminderCode }),
 	});
+
+export const sendReminderCreation = (
```
tomrf1, line 78: this is making a request to the MMA backend

server/reminders/reminderApi.ts
```diff
@@ -0,0 +1,137 @@
+import crypto from 'crypto';
```
tomrf1, line 1: this file is moved from server/reminderApi.ts
I've added `publicCreateReminderHandler` and logic around verification

shared/requiresSignin.ts
```diff
@@ -5,14 +5,17 @@ const publicPaths = [
...
 	'/api/help-centre/',
 	'/cancel-reminders/',
+	'/create-reminder/',
 	'/help-centre/',
 	'/maintenance/',
 ];
 
 export const requiresSignin = (path: string) => {
-	const normalizedPath = normalize(path + '/');
+	const pathWithoutQuerystring = path.split('?')[0];
```
tjmw, line 17: Are there any unit tests for `requiresSignin`? It feels like the kind of scenario where a couple of tests might act nicely as documentation (i.e. a path with and without a query string). What do you think?

server/routes/api.ts
```diff
@@ -274,7 +275,15 @@ router.get('/help-centre/topic/:topic', getTopicHandler);
...
-router.post('/reminders', createReminderHandler);
+router.post('/reminders/create', createOneOffReminderHandler); // requires sign-in
+router.post(
+	'/reminders/create-public/one-off',
+	publicCreateReminderHandler('ONE_OFF'),
+); // does not require sign-in, uses verification token
+router.post(
+	'/reminders/create-public/recurring',
+	publicCreateReminderHandler('RECURRING'),
+); // does not require sign-in, uses verification token
```
tjmw, line 286: Feel free to ignore this, it's very nitpicky! I like using the RESTful pattern of using the HTTP method and the URL path to describe the action - i.e. POST = create (the verb) and `/reminders` = the noun. So the `/create` in `/reminders/create` feels slightly redundant as the POST already conveys that information. For the new endpoints, perhaps `/public` at the start of the path could be a way of scoping the endpoints which don't require sign-in, leaving us with `POST /public/reminders/one-off` and `POST /public/reminders/recurring` (or even `/public/one-off-reminders` and `/public/recurring-reminders`. What do you reckon? As I said do feel free to ignore!

server/reminders/reminderApi.ts
```diff
@@ -0,0 +1,137 @@
...
+		if (!response.ok) {
+			captureMessage('Reminder sign up failed at the point of request');
+		}
+		res.sendStatus(response.status);
+	});
+
+// Instead of requiring the user to be signed in, this endpoint verifies the token
+export const publicCreateReminderHandler =
+	(reminderType: ReminderType) => async (req: Request, res: Response) => {
+		const { reminderData, token } = JSON.parse(req.body);
```
tjmw, line 49: In the case of an invalid payload should we catch errors thrown by `JSON.parse` and return a 400?

server/reminders/reminderApi.ts
```diff
@@ -0,0 +1,137 @@
...
+
+		if (reminderData && token) {
+			const reminderHmacKey = await getReminderHmacKey();
```
tjmw, line 52: Is there a chance this will fail (if we can't get the key)? Do we need to catch and return an explicit error if that happens?

tjmw (approved):
Really nice! I left a handful of nitpicky comments/questions to take or leave, happy to chat about any of them if it's useful.

client/components/mma/reminders/CreateReminder.tsx
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * This is the destination for reminder signup links in Braze emails.
+ * Attempts to create a reminder and displays a progress message.
+ *
+ * Expects the querystring to contain two params:
+ * - reminderData - stringified JSON payload for the reminders API
+ * - token - hashed version of the payload (using a key shared with Braze)
+ */
+export const CreateReminder = ({ reminderType }: CreateReminderProps) => {
+	const [status, setStatus] = useState<Status>('PENDING');
```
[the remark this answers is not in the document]
tomrf1: yep, and it's stolen from the existing CancelReminders component!

server/reminderApi.ts
```diff
@@ -1,16 +1,74 @@
...
 			captureMessage('Reminder sign up failed at the point of request');
 		}
 		res.sendStatus(response.status);
 	});
 
+// Instead of requiring the user to be signed in, this endpoint verifies the token
+export const publicCreateReminderHandler =
+	(reminderType: 'ONE_OFF' | 'RECURRING') =>
+	async (req: Request, res: Response) => {
+		const { reminderData, token } = JSON.parse(req.body);
```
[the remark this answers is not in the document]
tomrf1: good point, I'll update this

shared/requiresSignin.ts
tomrf1: Great point, I've pushed a unit tests file

server/routes/api.ts
tomrf1: The reason I like having `create` in the path is that we also have:
`POST /reminders/cancel`
`POST /reminders/reactivate`

So `POST` feels a bit ambiguous.
I do however like your idea of scoping the endpoints to `/public/...`, as it's an important difference, so I'll make that change

server/reminders/reminderApi.ts
tomrf1: 👍

server/reminders/reminderApi.ts
tomrf1: I've refactored the error handling in this function. It's tricky to make it elegant but I think it's a bit better now

michaelbjacobson (approved):

PR #1068, opened 2023-05-05
Fix Navigation on Help Centre

andrewHEguardian:
## What does this change?

This PR fixes a bug where navigation via the dropdown does not work on Help Centre pages. The solution is to use normal links when on a help centre page since they are part of a different react router so the `<Link>` component does not work. The bug was introduced in this PR: https://github.com/guardian/manage-frontend/pull/1033. We will still use `<Link>` for links on non-Help Centre pages since it does not refresh the page.

## How to test

Navigate to `/help-centre` and use the dropdown navigation, the links should work (and reload the page).

We should be aware that only links within MMA (and not Help Centre pages) can be linked to with the property `local` in the interface `NavItem`, and furthermore that we cannot navigate using React router between MMA and Help Centre.
 

## Accessibility

### DIFF
--- a/client/components/helpCentre/HelpCentrePage.tsx
+++ b/client/components/helpCentre/HelpCentrePage.tsx
@@ -81,7 +81,7 @@ const HelpCentreRouter = () => {
 	];
 
 	return (
-		<Main signInStatus={signInStatus}>
+		<Main signInStatus={signInStatus} isHelpCentrePage>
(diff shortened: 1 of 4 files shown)


charleycampbell (approved):
Good catch and thanks for the explanation!

andrewHEguardian:
tested on `PROD`

PR #1069, opened 2023-05-05
Create a test product builder

andrewHEguardian:
## What does this change?

There is a > 1000 line file called `fixtures/productDetail.ts` which contains hardcoded examples of different products that can be returned from `members-data-api`. This PR introduces a new `ProductBuilder` class which should make creating new examples of products easier, requiring fewer repeated lines of code.  

The idea is to reduce test setup from 100s of lines, e.g. 
image

to one single base fixture for each product type
image

with functions that can be reused to add the variations for different tests, e.g.
image

to create a test product detail easily like so:
image

## Accessibility


client/fixtures/productBuilder/baseProducts.ts
```diff
@@ -0,0 +1,150 @@
+import type { ProductDetail } from '../../../shared/productResponse';
+
+// export type ProductTypeKeys =
```
charleycampbell, line 3: Does this need to remain commented out?

charleycampbell (commented):
This is **such** a good idea!!! I wonder if there is anywhere we could/should document how to use it/how to add a new product to it? excellent 🚀

client/fixtures/productBuilder/baseProducts.ts
andrewHEguardian: comment should be clearer - those are the product types the base products will be use

charleycampbell (approved):

PR #1071, opened 2023-05-09
Fix cancel-reminders page routing

andrewHEguardian:
## What does this change?

This PR fixes the routing to the cancel-reminders page, and uses the `useParams` hook to get the token passed to the page. The react-router api seems to have changed since this page was built and the route was added - so it's unclear how long this page has been broken. 

## How to test

Go to `/cancel-reminders/*token here*`, you should see the message "Unsubscribing" then a success or failure screen depending on if the token was valid.

The Storybook stories have been updated to use the param instead of component props.

## Accessibility


tomrf1 (approved):
👍

PR #1083, opened 2023-05-19
 Show single contributions on Account Overview

jonflynng:
Designs: https://www.figma.com/file/JCmtpsw8m3fpFYw3MH1cUz/product-switching-feature?type=design&node-id=2732-99357&t=hI3KHKN7z3Eccr3a-0

Adds this component:

Screenshot 2023-05-19 at 15 07 00

 - [x] Will set the feature switch back to false once I know all Cypress tests have passed.

### DIFF
--- a/client/components/mma/accountoverview/AccountOverview.tsx
+++ b/client/components/mma/accountoverview/AccountOverview.tsx
@@ -31,7 +31,7 @@ import {
 	LoadingState,
 	useAsyncLoader,
 } from '../../../utilities/hooks/useAsyncLoader';
-import { allProductsDetailFetcher } from '../../../utilities/productUtils';
+import { allRecurringProductsDetailFetcher } from '../../../utilities/productUtils';
 import { GenericErrorScreen } from '../../shared/GenericErrorScreen';
 import { NAV_LINKS } from '../../shared/nav/NavConfig';
 import { SupportTheGuardianButton } from '../../shared/SupportTheGuardianButton';
@@ -235,7 +235,7 @@ const AccountOverviewPage = () => {
 
 const accountOverviewFetcher = () =>
 	Promise.all([
-		allProductsDetailFetcher(),
+		allRecurringProductsDetailFetcher(),
 		fetchWithDefaultParameters('/api/cancelled/'),
 		fetchWithDefaultParameters('/mpapi/user/mobile-subscriptions'),
 	]);
--- a/client/components/mma/accountoverview/AccountOverview.tsx
+++ b/client/components/mma/accountoverview/AccountOverview.tsx
@@ -136,9 +136,9 @@ const AccountOverviewPage = () => {
 	}
 
 	if (
-		(allActiveProductDetails.length === 0 &&
-			appSubscriptions.length === 0) ||
-		(allActiveProductDetails.length === 0 &&
+		allActiveProductDetails.length === 0 &&
+		appSubscriptions.length === 0 &&
+		(singleContributions.length === 0 ||
 			!featureSwitches.singleContributions)
 	) {
 		return <EmptyAccountOverview />;
--- a/client/components/mma/accountoverview/AccountOverview.tsx
+++ b/client/components/mma/accountoverview/AccountOverview.tsx
@@ -137,10 +137,7 @@ const AccountOverviewPage = () => {
 
 	if (
 		(allActiveProductDetails.length === 0 &&
-			appSubscriptions.length === 0 &&
-			singleContributions.length === 0) ||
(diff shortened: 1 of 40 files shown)


client/utilities/currencyIso.ts
```diff
@@ -1,3 +1,22 @@
...
+	AUD: '$',
+	EUR: '€',
+	NZD: '$',
+	CAD: '$',
+};
+
+export function convertCurrencyToSymbol(currency: string): string {
+	const symbol = currencySymbols[currency as CurrencyIso];
+	if (symbol === undefined) {
+		Sentry.captureException(`Unrecognized currency code: ${currency}`);
```
jonflynng, line 19: Is this the right thing to do?

shared/dates.ts
```diff
@@ -133,3 +133,11 @@ export const getOldestDate = (dates: Date[]) =>
 	dates.reduce((dateA: Date, dateB: Date) =>
 		dateA.valueOf() <= dateB.valueOf() ? dateA : dateB,
 	);
+
+export function convertTimestampToDate(timestamp: number): string {
+	const date = new Date(timestamp);
+	const day = date.getDate();
+	const month = date.toLocaleString('default', { month: 'short' });
+	const year = date.getFullYear();
+	return `${day} ${month} ${year}`;
```
andrewHEguardian, line 142: there is definitely a format string to do this, would `DATE_FNS_LONG_OUTPUT_FORMAT` not do the trick?

charleycampbell:
Looks good! I think this would be a great scenario addition to storybook

client/components/mma/accountoverview/SingleContributionCard.tsx
```diff
@@ -0,0 +1,84 @@
...
+					>
+						<dl css={keyValueCss}>
+							{singleContributions.map((contribution) => (
+								<>
+									<div
+										css={css`
+											padding-top: ${space[3]}px;
+										`}
+									>
+										<dt>Single contribution of</dt>
```
charleycampbell, line 47: Really minor but I would check with Helene/UX on where they want to use colons at the end of each statement

client/components/mma/accountoverview/AccountOverview.tsx
```diff
@@ -225,6 +245,15 @@ const AccountOverviewPage = () => {
 										subscription={subscription}
 									/>
 								))}
+							{featureSwitches.singleContributions &&
+								category === 'recurringSupport' &&
```
charleycampbell, line 249: Minor the heading above the card on the design has 'Your Support' instead of 'Recurring Support'

charleycampbell (commented):
Just a few of minor comments:
- minor design changes (which are on the code as separate comments)
- Going by the design looks like we want to use the personalised header which I don't think is currently implemented for Single Contributions 
- adding this to the AccountOverview storybook (separate comment)
But looks good 👍

client/utilities/productUtils.ts
```diff
@@ -64,13 +64,7 @@ export const allRecurringProductsDetailFetcher = () =>
 	});
 
 export const allSingleProductsDetailFetcher = () =>
-	fetchWithDefaultParameters('/api/me/one-off-contributions', {
-		headers: {
-			[X_GU_ID_FORWARDED_SCOPE]: getScopeFromRequestPathOrEmptyString(
```
andrewHEguardian, line 69: just curious - what did this do? I assume it was from a different api call

client/components/mma/accountoverview/SingleContributionCard.tsx
jonflynng: Yep will do

client/components/mma/accountoverview/AccountOverview.tsx
jonflynng: Will ask Daniel as well about this

shared/dates.ts
andrewHEguardian: you could do `dateString(new Date(timestamp), DATE_FNS_LONG_OUTPUT_FORMAT)` instead

jonflynng:
[MMA Display Single Contributions](https://trello.com/c/czvkFAEK/4137-mma-display-single-contributions)

client/components/mma/accountoverview/AccountOverview.tsx
jonflynng: We're probably going to rename the "Recurring Support" header to "Your Support", but will do this in a subsequent PR on confirmation.

andrewHEguardian (dismissed):
Looks good 🚀  nice work

PR #1086, opened 2023-05-22
Fix braze reminder query string

tomrf1:
Braze can incorrectly append a new query string to a url that already has a query string, meaning we end up with a second `?` where there should be a `&`.
Braze do not seem likely to fix this any time soon, so we can defend against it here in MMA.


KaliedaRik (dismissed):
A necessary change that we can (hopefully) revert should Braze do the decent thing and fix on their end

KaliedaRik (approved):

PR #1090, opened 2023-05-24
Use CODE environment in AWS not DEV

rupertbates:
## What does this change?

There has been a lot of confusion over the years about what the different stages or environments (DEV/CODE/PROD) mean and which should be used for what. This PR is part of a piece of work intended to clear that up.

Full details are [in this document](https://docs.google.com/document/d/1wg1kMk7U7ht3AmPpsPUimtz1FkVbu5BWjijjme2598s/edit#) but in summary:
- The DEV environment refers only to a developers local machine and config settings here are things like local domain names eg. manage.thegulocal.com
- The CODE and PROD environments are for systems running in AWS or production and test versions of external systems

Supporter Revenue Environments(4)

## Changes
- Point all api gateway endpoints to their CODE versions not DEV (these will be deleted eventually)
- Remove permissions for manage to access DEV api gateway
- Rename uat to test in Stripe config, we are not using the term UAT any more as we don't have a separate environment for test users any more, test users just use the CODE environment

## How to test
In CODE carry out a product switch, if it works then all is good.

These changes only affect the CODE environment so risks are minimal, I have tested:
- changing contribution amount
- product switch
- cancellation

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ NOTE: nginx proxies CODE instances of [`identity-frontend`](https://github.com/g
 
 #### Running locally (each day you need use it)
 
-1.  You will also need to have [members-data-api](https://github.com/guardian/members-data-api) (`./start-api.sh`) running, to get subscription data (from DEV Zuora/Salesforce).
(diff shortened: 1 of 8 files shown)


andrewHEguardian (approved):
thank you for this 👏  so much simpler to understand now

PR #1110, opened 2023-06-12
Minimal recreation of cypress error

andrewHEguardian:
## This PR is for illustrative purposes only

This PR recreates an issue (found in #1109 ) with Cypress tests that fail because of an import from another file. Cursory googling suggests something to do with Webpack and or dynamic imports... ([see this issue](https://github.com/cypress-io/cypress/issues/18435))

### Error: 
`  > Automatic publicPath is not supported in this browser`

image

### Finding the problem

Neither the file changed nor the file referenced have any dynamic imports, but navigating through the imports of each file we eventually find the dynamic import in `Footer.tsx`. 

`
useEffect(() => {
		import('@guardian/consent-management-platform').then(({ cmp }) => {
			setImportedCmp(cmp);
		});
	}, []);
`

Through the very convoluted route - Cypress test -> `productBuilder.ts` -> `productResponse.ts` -> `import { GROUPED_PRODUCT_TYPES } from './productTypes';` -> `productTypes.ts` -> `PRODUCT_TYPES` -> `shuffledContributionsCancellationReasons` -> `contributionsCancellationReasons` -> `ContributionsCancellationFlowFinancialSaveAttempt` -> `CancellationContext` -> `PageContainer` -> `HasMinimalFooterContext` -> `Main.tsx` -> `Footer.tsx` -> dynamic import

If we comment these lines out the Cypress tests start running again.

### What does it mean?

We cannot reference anything that imports something from the chain above in a Cypress test (or we can [upgrade Cypress](https://github.com/cypress-io/cypress/issues/18435#issuecomment-1259788368)). Or else the files should be split up to have fewer responsibilities, maybe the functions/consts in that chain could live in separate files.

### DIFF
--- a/client/fixtures/productBuilder/productBuilder.ts
+++ b/client/fixtures/productBuilder/productBuilder.ts
@@ -1,5 +1,15 @@
-import type { Card, ProductDetail } from '../../../shared/productResponse';
+import type {
+	Card,
+	ProductDetail} from '../../../shared/productResponse';
+import {
+	isPaidSubscriptionPlan,
+} from '../../../shared/productResponse';
 import type { GroupedProductTypeKeys } from '../../../shared/productTypes';
+import type {
+	CurrencyIso} from '../../utilities/currencyIso';
+import {
+	convertCurrencyToSymbol,
+} from '../../utilities/currencyIso';
 
 export const cards = {
 	visaActive: () => {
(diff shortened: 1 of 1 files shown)


andrewHEguardian:
demo'd to the team

PR #1117, opened 2023-06-16
Default to exact package versions

andrewHEguardian:
## What does this change?

By default pin dependency to exact version, instead of ^version (following on from #1085). Pinning these is the recommendation and making it the default behaviour means we don't have to ask people on a PR if they haven't done so.

## Accessibility


charleycampbell (approved):
Nice follow up to the previous PR 👍

PR #1123, opened 2023-06-22
add new endpoint for updating subscription cancellation reason in Zuora

jonflynng:
This is part of the new cancellation saves work, which allows customers to optionally select the reason for cancelling at the end of the journey, after the cancellation is complete.

Previously, customers had to select the cancellation reason before cancelling. When confirming the cancellation, the reason would be sent to the API, which would perform both the reason update and cancellation. Now, these two actions must be done separately.

[see the PR in members-data-api](https://github.com/guardian/members-data-api/pull/1024)

- [x] Tested in CODE
- [x] merge server-side PR first

### DIFF
--- a/client/components/mma/cancel/cancellationSaves/SelectReason.tsx
+++ b/client/components/mma/cancel/cancellationSaves/SelectReason.tsx
@@ -151,6 +151,24 @@ function cancellationCaseFetch(
 	});
 }
 
+function updateZuoraCancellationReason(
+	selectedReasonId: string,
+	productDetail: ProductDetail,
+) {
+	return fetch(
+		'/api/update-cancellation-reason/' +
+			productDetail.subscription.subscriptionId,
+		{
+			method: 'POST',
+			body: JSON.stringify({ reason: selectedReasonId }),
+			headers: {
+				'Content-Type': 'application/json',
+				[MDA_TEST_USER_HEADER]: `${productDetail.isTestUser}`,
+			},
+		},
+	);
+}
+
 export const SelectReason = () => {
 	const navigate = useNavigate();
 	const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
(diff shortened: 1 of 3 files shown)


jonflynng:
[Membership Saves - Update reason](https://trello.com/c/p4JrQJyp/4174-membership-saves-update-reason)

client/components/mma/cancel/cancellationSaves/SelectReason.tsx
```diff
@@ -196,11 +214,15 @@ export const SelectReason = () => {
 
 		try {
 			setIsSubmitting(true);
-			const response = await cancellationCaseFetch(
-				selectedReasonId,
-				productType,
-				productDetail,
-			);
+			const response = await Promise.all([
```
andrewHEguardian, line 217: I'm still not 100% sure what happens when one of these calls fails - or what should happen

jonflynng: I'll test in code

andrewHEguardian: if the calls fail the user will still navigate to the next page

cypress/integration/parallel-2/membershipSave.spec.ts
```diff
@@ -56,6 +56,13 @@ if (featureSwitches.membershipSave) {
 				},
 			}).as('get_case');
 
+			cy.intercept('POST', '/api/update-cancellation-reason/**', {
```
andrewHEguardian, line 59: we should add the button click and `wait` for this into a test

andrewHEguardian (approved):
Looks good 👍 just one comment that it could have an extra test but happy with the manual testing

client/components/mma/cancel/cancellationSaves/SelectReason.tsx
jonflynng: Yep, just tested it. That's better UX than showing an error there.

PR #1127, opened 2023-06-26
Uncomment tests that should not be commented out

andrewHEguardian:
## What does this change?

When updating the test product #1100 - accidentally commented out most of the cypress test for holiday stops. This wasn't caught in the diff since it wasn't rendered because the diff was too large (`Large diffs are not rendered by default.`) 🤦 

## Accessibility


charleycampbell (approved):
Good catch!

PR #1153, opened 2023-08-01
Migrate from TeamCity to GitHub Actions

michaelbjacobson:
This PR migrates the repo from TeamCity to GH Actions for CI/CD. The branch is currently failing because I've removed the TeamCity build file. Once we've merged this I'll remove the repo from TeamCity and we won't have any issues.


johnduffell (approved):
nice work thanks Michael! 👏 
I don't claim to understand the detail but I'm assuming it's based off a template and significant problems will be clear pretty soon.
I believe this is the last one so well done getting it over the line

michaelbjacobson:
> nice work thanks Michael! 👏 I don't claim to understand the detail but I'm assuming it's based off a template and significant problems will be clear pretty soon. I believe this is the last one so well done getting it over the line

No template per se, but I previously migrated some of the Conversions repos from TeamCity to GH Actions, and this is fairly similar.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Env
+        run: env
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
akash1810, line 31: Let's use the [latest version](https://github.com/aws-actions/configure-aws-credentials/blob/main/CHANGELOG.md) of this action, v4.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+      id-token: write
+      contents: read
+    name: manage-frontend build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Env
+        run: env
+
+      - name: Checkout
+        uses: actions/checkout@v3
```
akash1810, line 28: Let's use the [latest version](https://github.com/actions/checkout/blob/main/CHANGELOG.md) of this action, v4.

script/cfn.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR="${DIR}/.."
+cd "$ROOT_DIR/cdk"
+
+yarn install
```
akash1810, line 10: Let's add the `--frozen-lockfile` flag here to increase determinism.

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass `--frozen-lockfile` flag.
> – https://classic.yarnpkg.com/lang/en/docs/cli/install/

script/ci.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR="${DIR}/.."
+cd "$ROOT_DIR"
+
+yarn install
```
akash1810, line 10: Let's add the `--frozen-lockfile` flag here to increase determinism.

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass `--frozen-lockfile` flag.
> – https://classic.yarnpkg.com/lang/en/docs/cli/install/

script/ci.sh
```diff
@@ -0,0 +1,24 @@
...
+yarn type-check
+yarn lint
+yarn bundle
+
+cp package.json dist
+cp riff-raff.yaml dist
```
akash1810, line 16: We shouldn't need to move the `riff-raff.yaml` file around, as we're configuring `guardian/actions-riff-raff@v2` to find it in it's original location.

script/ci.sh
```diff
@@ -0,0 +1,24 @@
...
+
+pushd dist
+yarn install --production
+popd
+
+cd dist
+
+zip -FSr "${ROOT_DIR}/manage-frontend.zip" ./*
```
akash1810, line 24: minor: the `pushd`, `popd` and `cd` can be simplified/combined here via:

```sh
(
  cd dist
  yarn install --production
  zip -FSr "${ROOT_DIR}/manage-frontend.zip" ./*
)
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+jobs:
+  CI:
+    if: >-
+      (github.actor != 'dependabot[bot]') &&
```
akash1810, line 14: Curious to see a filtering out of Dependabot PRs. Does the team have a particular process for reviewing Dependabot PRs?

michaelbjacobson:
Thanks for the comprehensive feedback @akash1810, I'll get to work on all that right away 🙂

.github/workflows/ci.yml
michaelbjacobson: This particular bit of filtering is some boilerplate I used in Conversions, but may not be relevant to Retention. @johnduffell, @rupertbates thoughts?

andrewHEguardian:
Previously our jest tests were running in the TeamCity build - how should we be doing this now?

akash1810 (commented):
LGTM. Still curious about the Dependabot filtering, but not blocking.

.github/workflows/ci.yml
michaelbjacobson: I've removed this filter, as I don't think it's needed/wanted for this repo.

johnduffell (approved):
I don't quite feel qualified to comment, but it looks pretty comprehensive, thanks for completing this 👍

PR #1159, opened 2023-08-04
Paolo castro/added subscription information

paologu:
## What does this change?

Added subscription information to Diagnostics Page

this is to help diagnose what subscriptions the user can see on their account without having to go to Zuora

image


client/components/helpCentre/diagnosticInformation/SubscriptionInformation.tsx
```diff
@@ -0,0 +1,68 @@
...
+import type { MembersDataApiResponse } from '../../../../shared/productResponse';
+import { isProduct } from '../../../../shared/productResponse';
+import { allRecurringProductsDetailFetcher } from '../../../utilities/productUtils';
+
+export const SubscriptionInformation = () => {
+	const [memberData, setMemberData] = useState<
+		MembersDataApiResponse | 'failed' | 'loading'
+	>('loading');
+
+	useEffect(() => {
```
andrewHEguardian, line 11: there is actually a hook in the codebase called `useAsyncLoader` which does an async fetch and returns a loading state that you could use here. It does a lot of the same things and its usage looks fairly similar to what you have here

andrewHEguardian (dismissed):
Left a couple of small notes but looks good 👍

andrewHEguardian (approved):

PR #1164, opened 2023-08-16
Create a specific URL for Manage product pages

andrewHEguardian:
## What does this change?

This PR changes the URL that the `Manage product` page exists on - and adds async loading if you go directly to this page through the URL.

The motivation for this change is to have a URL that readers can be linked to which will allow them to change the amount they support. I.e. `/support` in this case. In order to do this the routing for this page has changed from being at the `GROUPED_PRODUCT_TYPE` url to the specific `PRODUCT_TYPE`. For example, previously when managing a supporter plus subscription the URL would be `/recurringSupport`, it will now be `/support`. The same applies to other products, homedelivery, guardianweekly etc. 

Async loading: before, this page only worked when clicked through from another page that passes props. Now, there is logic to try and load the product from MDAPI, filtered by the productType specified in the URL. If there are multiple products returned the page loads the first in the list (unlikely that a user has multiple of the same product)

Using the specific product type also sidesteps the bug found in this PR: #1137 

## How to test

With a Supporter Plus subscription 

1) -> on account overview, click `Manage recurring support` and see the URL is `/support`

2) go directly to `/support` and see a "Loading your product..." and then see the page load (see image)

Potential risk: breaking existing URLs. This shouldn't make a difference since previously the page could not load via URL directly 

## Images

image

## Accessibility

### DIFF
--- a/shared/productTypes.ts
+++ b/shared/productTypes.ts
@@ -651,7 +651,6 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
 		groupedProductType: 'subscriptions',
 		allProductsProductTypeFilterString: 'GuardianPatron',
 		urlPart: 'guardianpatron',
-		legacyUrlPart: 'guardianpatron',
 		getOphanProductType: () => 'GUARDIAN_PATRON', //TODO: This value doesn't exist in Ophan yet
 		showTrialRemainingIfApplicable: true,
 		softOptInIDs: [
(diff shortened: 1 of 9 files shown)


shared/productTypes.ts
```diff
@@ -651,7 +651,6 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
 		groupedProductType: 'subscriptions',
 		allProductsProductTypeFilterString: 'GuardianPatron',
 		urlPart: 'guardianpatron',
-		legacyUrlPart: 'guardianpatron',
```
andrewHEguardian, line 654: having this legaclyUrlPart the same as urlPart caused an infinite redirect if you went to `/guardianpatron`

KentGrm (approved):

PR #1169, opened 2023-08-24
Cc/upsell/thanks switching v1

charleycampbell:
## What does this change?
New screen for the thank you page after the upgrading screen where the user has switched. This is the first version of the screen as final designs are still to be signed off so wording and design may change. And actual monetary amounts need to be passed in.

## How to test
Storybook sceneario added or head to upgrade-support/thank-you

## Images
image

## Accessibility


client/components/mma/upgrade/UpgradeSupportSwitchThankYou.tsx
```diff
@@ -0,0 +1,111 @@
...
+						<span>
+							Your new support will start today. It can take up to
+							an hour for your support to be activated
+						</span>
+					</li>
+				</ul>
+			</section>
+			<section>
+				<div css={signInCss}>
+					<SwitchSignInImage />
```
andrewHEguardian, line 80: I wonder if this banner would be worth making a component at this point - is it exactly the same as on the switch page?

client/components/mma/upgrade/UpgradeSupportSwitchThankYou.tsx
```diff
@@ -0,0 +1,111 @@
...
+	signInContentContainerCss,
+	signInCss,
+	signInHeadingCss,
+	signInParaCss,
+} from '../../shared/SignIn';
+import {
+	headingCss,
+	stackedButtonLayoutCss,
+	whatHappensNextCss,
+} from '../cancel/cancellationSaves/SaveStyles';
```
andrewHEguardian, line 22: I think either we need to duplicate these styles or make them common somewhere else - feels odd to reference files from the switch and save journeys in this folder

client/components/mma/upgrade/UpgradeSupportSwitchThankYou.tsx
charleycampbell: I have made a SignIn shared file and was starting to refactor

client/components/mma/upgrade/UpgradeSupportSwitchThankYou.tsx
charleycampbell: Yeah agreed. It's whether to start touching files which aren't part of this change or not

andrewHEguardian: In general I'd say it's fine, if it's just moving things around

andrewHEguardian (approved):
Looks good 👍  let's refactor more as we go, on to the next screen 🚀

PR #1175, opened 2023-09-01
Upgrade Support - add preview api call

andrewHEguardian:
## What does this change?

Adds the switch preview api call before showing step 2, to be able to render the payment amount.

The preview call is called every time the amount changes (with the threshold amount if a lower amount is chosen)

## How to test

Go to '/upgrade-support`, click 'Continue with xx' - see Loading bar where step 2 (confirm) should be

## Images

image

## Accessibility

### DIFF
--- a/client/components/mma/shared/SupporterPlusTsAndCs.tsx
+++ b/client/components/mma/shared/SupporterPlusTsAndCs.tsx
@@ -37,8 +37,8 @@ export const SupporterPlusTsAndCs = ({
 	const currencySymbol = convertCurrencyToSymbol(currencyISO);
 	const monthlyOrAnnual =
 		calculateMonthlyOrAnnualFromBillingPeriod(billingPeriod);
-	const monthlyThreshold = getBenefitsThreshold(currencyISO, 'Monthly');
-	const annualThreshold = getBenefitsThreshold(currencyISO, 'Annual');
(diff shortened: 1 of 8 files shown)


client/components/mma/shared/SupporterPlusTsAndCs.tsx
```diff
@@ -37,8 +37,8 @@ export const SupporterPlusTsAndCs = ({
 	const currencySymbol = convertCurrencyToSymbol(currencyISO);
 	const monthlyOrAnnual =
 		calculateMonthlyOrAnnualFromBillingPeriod(billingPeriod);
-	const monthlyThreshold = getBenefitsThreshold(currencyISO, 'Monthly');
```
charleycampbell, line 40: Just curious why this renaming?

charleycampbell (approved):
Looks good! I think we're right to flag that whenever the amount changes we will all the preview endpoint again... 👍

client/components/mma/shared/SupporterPlusTsAndCs.tsx
andrewHEguardian: the equivalent config for contributions amounts used month / year instead of monthly / annual, so I made it more consistent but it also avoids having to calculate `monthlyOrAnnual` to get the config

PR #1177, opened 2023-09-06
One product-move-api preview call in upgrade journey

andrewHEguardian:
## What does this change?

Refactors the upgrade journey to only call the `product-move-api` preview once. It uses the refund amount from the response to calculate the amount payable today. This change also puts the api call in the parent component of where it was previously (from `ConfirmForm` into `UpgradeSupport`), meaning it happens when the component first renders not when the `Continue` button is clicked

In effect we are calculating two response variables that are normally determined [in the product-move-api](https://github.com/guardian/support-service-lambdas/blob/main/handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala) on the client side instead

## How to test

1. Go to `/upgrade-support`.
2. In network tab see api call to fetch preview info
3. Select an amount 
4. Continue
5. See amount payable today
6. Select a different amount
7. See new amount payable today

This PR also adds a Cypress test that counts the number of api calls.

- The server side and client side could become out of sync if a change is performed and not duplicated 
- Inaccurate information could be passed to the server side if the client makes an error

## Images

Loading details if a user clicks Continue before the preview call finishes
image

## Accessibility

### DIFF
--- a/client/utilities/productMovePreview.ts
+++ b/client/utilities/productMovePreview.ts
@@ -1,0 +1,22 @@
+// These functions are client side versions of logic performed in the product-move-api to render preview details
+// They are duplicated here to avoid making many api calls if a user selects different amounts that require preview
+
+/**
+ * calculate the amount a user will pay at the time they switch product
+ */
+export function calculateAmountPayableToday(
+	userChosenAmount: number,
+	contributionRefundAmount: number,
+): number {
+	// contributionRefundAmount will be a negative number (eg if the refund is £5 the param will be -5)
+	return userChosenAmount - -1 * contributionRefundAmount;
+}
+
+/**
+ * calculate param to pass to non-preview call to check whether the amount is less than the minimum Stripe charge
+ */
+export function calculateCheckChargeAmountBeforeUpdate(
+	amountPayableToday: number,
+): boolean {
+	return amountPayableToday > 0 && amountPayableToday < 0.5;
+}
--- a/client/utilities/productMovePreview.ts
+++ b/client/utilities/productMovePreview.ts
@@ -9,7 +9,7 @@ export function calculateAmountPayableToday(
 	contributionRefundAmount: number,
 ): number {
 	// contributionRefundAmount will be a negative number (eg if the refund is £5 the param will be -5)
-	return userChosenAmount - -1 * contributionRefundAmount;
+	return userChosenAmount - Math.abs(contributionRefundAmount);
 }
 
 /**
--- a/cypress/integration/parallel-2/upgradeSupport.spec.ts
+++ b/cypress/integration/parallel-2/upgradeSupport.spec.ts
@@ -24,7 +24,7 @@ describe('upgrade support', () => {
 		}).as('product_move');
 	});
 
-	it('upgrades contribution to supporter plus', () => {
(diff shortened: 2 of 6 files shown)


charleycampbell (dismissed):
Looks good to me! Would be good to get Rupert's thoughts on it too!

cypress/integration/parallel-2/upgradeSupport.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+			),
+		}).as('mdapi_get_contribution');
+
+		cy.intercept('POST', '/api/product-move/**', {
+			statusCode: 200,
+			body: productMovePreviewResponse,
+		}).as('product_move');
+	});
+
+	it('upgrades contribution to supporter plus', () => {
```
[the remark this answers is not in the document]
andrewHEguardian: yes, definitely - was going to add more tests later

client/utilities/productMovePreview.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * calculate the amount a user will pay at the time they switch product
+ */
+export function calculateAmountPayableToday(
+	userChosenAmount: number,
+	contributionRefundAmount: number,
+): number {
+	// contributionRefundAmount will be a negative number (eg if the refund is £5 the param will be -5)
+	return userChosenAmount - -1 * contributionRefundAmount;
```
rupertbates, line 12: Is -1 * contributionRefundAmount just doing the same as `Math.abs()`? that would be easier to read for me, it took me a while to parse this!

client/utilities/productMovePreview.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * calculate param to pass to non-preview call to check whether the amount is less than the minimum Stripe charge
+ */
+export function calculateCheckChargeAmountBeforeUpdate(
+	amountPayableToday: number,
+): boolean {
+	return amountPayableToday > 0 && amountPayableToday < 0.5;
```
rupertbates, line 21: Is this definitely `> 0 && < 0.5` not `>= 0 && <= 0.5`?

andrewHEguardian: `collectPayment = !(amountPayableToday > BigDecimal(0) && amountPayableToday < BigDecimal(0.50))`
the scala has this calculation so I think it's exclusive

client/utilities/productMovePreview.ts
andrewHEguardian: yes I agree this is ugly to look at! I think prettier reformatted this to remove brackets I put around the multiplication...

Should it be `Math.abs` ... I guess I wasn't sure if there was an edgecase where the refundAmount would actually be a positive number. If we can confirm this won't happen then let's change to using absolute value

andrewHEguardian: ^ or alternatively just add the two together and let the comment explain why

rupertbates: Refund amount will never be positive unless there's a bug somewhere else, either way we want to end up with a positive number so I think abs is the way to go.

rupertbates (approved):
👍

PR #1186, opened 2023-09-14
Removing non serviceable countries from address dropdown

charleycampbell:
## What does this change?
Removing countries we are no longer able to service from country address drop downs. This includes the address filed where you can update your delivery address and correspondence address for all products.

## How to test
Check country drop downs in the Settings -> correspondence address OR on a deliverable product update your address

## Accessibility


client/components/mma/identity/models.ts
```diff
@@ -349,23 +349,17 @@ interface CountryValues {
 }
 export const COUNTRIES: CountryValues[] = [
 	{ iso: 'AF', name: 'Afghanistan' },
-	{ iso: 'AL', name: 'Albania' },
```
andrewHEguardian, line 352: Would it make more sense to have a list of these and filter them out from this master list of all countries?

KentGrm: The code would be clearer with a filter out. But it's more secure / guaranteed if we remove them. Either is fine with me

andrewHEguardian: just thought it might be easier if the list changes to keep track of it separately than to eg add a country back in

andrewHEguardian: it also might not be clear to someone coming to the codebase why there is an incomplete list of ISO countries

charleycampbell: We would need to remove or a add the country from the non-servicable list if anything was change. It just depends on how careful we want to be

PR #1187, opened 2023-09-14
Add National Delivery as a product type

rupertbates:
## What does this change?
[This change to members-data-api](https://github.com/guardian/members-data-api/pull/1036) makes National Delivery subs appear in the `user-attributes/me/mma` endpoint used by manage-frontend. This breaks manage however because National Delivery is not defined as a product so the site doesn't know how to display it.

This PR fixes that issue.

### DIFF
--- a/shared/productTypes.ts
+++ b/shared/productTypes.ts
@@ -38,7 +38,6 @@ type ProductFriendlyName =
 	| 'newspaper voucher subscription'
 	| 'newspaper subscription card'
 	| 'newspaper home delivery subscription'
-	| 'newspaper national delivery subscription'
 	| 'digital subscription'
 	| 'monthly + extras'
 	| 'annual + extras'
@@ -53,6 +52,7 @@ type ProductUrlPart =
 	| 'voucher'
 	| 'subscriptioncard'
 	| 'homedelivery'
+	| 'nationaldelivery'
 	| 'digital'
 	| 'support'
 	| 'guardianweekly'
@@ -449,12 +449,12 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
 	},
 	nationaldelivery: {
 		productTitle: calculateProductTitle('Newspaper Delivery'),
-		friendlyName: () => 'newspaper national delivery subscription',
-		shortFriendlyName: 'newspaper national delivery',
+		friendlyName: () => 'newspaper home delivery subscription',
+		shortFriendlyName: 'newspaper home delivery',
 		productType: 'nationaldelivery',
 		groupedProductType: 'subscriptions',
-		allProductsProductTypeFilterString: 'HomeDelivery', // Is this right?
-		urlPart: 'homedelivery', // Is this right?
+		allProductsProductTypeFilterString: 'HomeDelivery',
+		urlPart: 'nationaldelivery',
 		getOphanProductType: () => 'PRINT_SUBSCRIPTION',
 		productPageNewsletterIDs: [FRONT_PAGE_NEWSLETTER_ID],
 		softOptInIDs: [
--- a/shared/productTypes.ts
+++ b/shared/productTypes.ts
@@ -38,6 +38,7 @@ type ProductFriendlyName =
 	| 'newspaper voucher subscription'
 	| 'newspaper subscription card'
 	| 'newspaper home delivery subscription'
+	| 'newspaper national delivery subscription'
 	| 'digital subscription'
 	| 'monthly + extras'
 	| 'annual + extras'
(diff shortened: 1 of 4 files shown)


shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
 			productFilenamePart: 'Newspaper - Home Delivery',
 		},
 	},
+	nationaldelivery: {
```
rupertbates, line 450: Not 100% sure on the values in this object, please give your opinion!

shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
...
+		productTitle: calculateProductTitle('Newspaper Delivery'),
+		friendlyName: () => 'newspaper national delivery subscription',
```
andrewHEguardian, line 452: on the checkout I don't think we differentiated the national delivery with a different name, so maybe the same logic applies here?

shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
...
+		shortFriendlyName: 'newspaper national delivery',
+		productType: 'nationaldelivery',
+		groupedProductType: 'subscriptions',
+		allProductsProductTypeFilterString: 'HomeDelivery', // Is this right?
+		urlPart: 'homedelivery', // Is this right?
```
andrewHEguardian, line 457: I think probably this should be unique, because ReactRouter is going to pick one (not sure which? maybe the first) and if we specify different config options than for normal home delivery it could create problems

shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
 			productFilenamePart: 'Newspaper - Home Delivery',
 		},
 	},
+	nationaldelivery: {
+		productTitle: calculateProductTitle('Newspaper Delivery'),
+		friendlyName: () => 'newspaper national delivery subscription',
+		shortFriendlyName: 'newspaper national delivery',
+		productType: 'nationaldelivery',
+		groupedProductType: 'subscriptions',
+		allProductsProductTypeFilterString: 'HomeDelivery', // Is this right?
```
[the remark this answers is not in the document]
rupertbates: Will check

shared/productTypes.ts
andrewHEguardian: see `MMAPage.tsx`

shared/productTypes.ts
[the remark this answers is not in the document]
rupertbates: Yup works

shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
...
+					{
+						label: 'Instructions Not Followed',
+						messageIsMandatory: true,
+					},
+					...commonDeliveryProblemTypes,
+				],
+			},
+		},
+		fulfilmentDateCalculator: {
+			productFilenamePart: 'Newspaper - National Delivery',
```
andrewHEguardian, line 487: Do we need to update something in the fulfilmentDateCalculator to make this work?

shared/productTypes.ts
```diff
@@ -445,6 +447,46 @@ export const PRODUCT_TYPES: { [productKey in ProductTypeKeys]: ProductType } = {
...
+			SoftOptInIDs.SupporterNewsletter,
+		],
+		holidayStops: {
+			issueKeyword: 'paper',
+		},
+		delivery: {
+			showAddress: showDeliveryAddressCheck,
+			enableDeliveryInstructionsUpdate: true,
+			records: {
+				productNameForProblemReport: 'National Delivery',
```
andrewHEguardian, line 473: same for this - does this need to match something on an api somewhere?

andrewHEguardian (approved):
Looks good 👍  just a couple of questions about values used for apis but happy to approve if you know what they mean

shared/productTypes.ts
rupertbates: Yes, the change to support it is at the bottom of this PR: https://github.com/guardian/support-service-lambdas/pull/2055/files

shared/productTypes.ts
rupertbates: This value maps to a value in the delivery problem creation in support-service-lambdas here: https://github.com/guardian/support-service-lambdas/blob/b0194326780271679c71a63096cf4c863ae1aa52/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala#L55

I'm not completely sure what this value should be yet because the accompanying work hasn't been done, but I've discussed with John D who is doing that work, and will go with this for now and update later if necessary.

PR #1188, opened 2023-09-15
Revert "Migrate clientside routes serverside"

raphaelkabo:
Reverts guardian/manage-frontend#1163 because it created an issue where JSON bodies were being sent from the server unstringified, which is causing downstream lambdas to fail.


kelvin-chappell (approved):
👍

undergroundquizscene (approved):

PR #1190, opened 2023-09-15
Migrate clientside routes serverside - take 2

raphaelkabo:
## What does this change?

**Original PR:** https://github.com/guardian/manage-frontend/pull/1188

The original PR makes changes to the routing in MMA to move some clientside calls to external APIs to instead occur serverside.

In that PR, [we added an Express server feature which parsed JSON bodies in requests](https://github.com/guardian/manage-frontend/pull/1163/files#diff-cfbaa1474fd5b35981f53d787fd188be4fe43c9d9f68525c67ff6e0fe2934aadR38).

- That feature didn't exist on the server before
- As a result, all JSON requests being made to the server from the MMA clientside app started to be parsed.
- The JSON bodies of requests sent from the client to the server are all `JSON.stringify()`-ed at the point of being sent, which is necessary for the Fetch API, otherwise it will send the infamous `object Object`.
- In some server components, for instance [reminderApi.ts](https://github.com/guardian/manage-frontend/blob/c9ec5cdb910226820c1b67f327038975d4f2a257/server/reminders/reminderApi.ts#L163-L170), which receive requests coming from the clientside and send them on to downstream routes, the `fetch` call is _not_ being `JSON.stringify()`-ed (thanks to @undergroundquizscene for catching this quickly!). This worked fine before because the JSON bodies hitting the server weren't being automatically processed into JavaScript objects by Express. They remained as strings of stringified JSON and would be proxied unharmed all the way down the chain.
- As a result of the change, these serverside `fetch` calls started sending `object Object` instead of JSON strings.
- The bug was spotted by @undergroundquizscene because some Lambdas started showing a spike in errors after the previous PR was merged.

This PR fixes the bug by:

- Not putting `server.use(express.json());` into `server.ts`.
- [Adding code into `idapiProxyHandler()` to parse received JSON bodies as JSON when relevant](https://github.com/guardian/manage-frontend/pull/1190/files#diff-6e7a2e1dc7a77b9f42d0bfa57a1ba8298a58bdf8241375525fc627d56370fd3bR144-R154).

## Testing

- [ ] Tested in CODE
- [ ] Tested in PROD


undergroundquizscene:
Thanks for the fix, and thanks for putting in a separate commit! Nice and easy to review :)

coldlink (approved):

PR #1191, opened 2023-09-19
Upgrade journey: Create suggested amounts getter

andrewHEguardian:
## What does this change?

Adds a lookup object for suggested amounts for the current amount a reader is supporting with before entering the upgrade support journey. Each currency/monthly or annual combination is a function that calculates the 3 amounts to suggest.

[Source spreadsheet](https://docs.google.com/spreadsheets/d/1lxnOJhSPrddZmdjNe85sbTCtprqQlSy6_QcwY_n7AFk/edit#gid=0)

## How to test

Go to `/upgrade-support` and see the 3 amounts that come from a spreadsheet.

How do we test there isn't a mistake in this implementation... (not sure)

Creating this was a lot of manual data entry! Not good for maintainability in future

## Images

image

## Accessibility


client/utilities/supportPricing/suggestedAmounts.ts
```diff
@@ -0,0 +1,275 @@
...
+		year: getUsdAnnual,
+	},
+	international: {
+		month: getUsdMonthly,
+		year: getUsdAnnual,
+	},
+};
+
+function getGbpMonthly(amount: number) {
+	switch (true) {
```
rupertbates, line 64: Why a switch here rather than just 
```javascript
if(amount <= 3) 
 return [5, 7, 10];
```
?

client/utilities/supportPricing/supporterPlusPricing.ts
```diff
@@ -51,27 +50,17 @@ export function getBenefitsThreshold(
...
-	);
 
-	return getSuggestedAmounts(currentAmount, monthlyOrAnnual);
+	return mainPlan.billingPeriod === 'month'
+		? getSuggestedAmountsMonthly(currentAmount)
+		: getSuggestedAmountsAnnual(currentAmount);
 }
 
-function suggestedAmountsMonthly(currentAmount: number) {
+export function getSuggestedAmountsMonthly(currentAmount: number) {
```
rupertbates, line 63: What are these calculations used for compared to the new ones?

client/utilities/supportPricing/suggestedAmounts.ts
andrewHEguardian: I think that `if` might be better, I started with switching on the value and but the ranges were too big on the annuals

client/utilities/supportPricing/supporterPlusPricing.ts
andrewHEguardian: the suggested amounts when you update the amount of a supporter plus
image
confusing right? they're also the default fallback if someone goes over the range defined in the spreadsheet

andrewHEguardian: suggestions appreciated for how to name this in a way that makes it less confusing

rupertbates: So one is for RC upsell and the other is for S+? If so maybe put the product name in the function name

rupertbates: That's much clearer to me 👍

rupertbates (approved):

KentGrm (approved):

PR #1192, opened 2023-09-19
configure `assets-static` Riff-Raff deployment

michaelbjacobson:
This PR configures the `assets-static` Riff-Raff deployment.


johnduffell (approved):
let's give it a go! 👍

PR #1193, opened 2023-09-19
add justification for assets-static bucket

johnduffell:
This just adds a bit of context around the assets-static pattern for manage-frontend


andrewHEguardian (approved):
thanks for the explanation! 😄

PR #1201, opened 2023-09-20
Stop zero appearing under input box

charleycampbell:
## What does this change?
Reordered the condition by having a numeric value passed in on the right hand side of the &&. This stops the 0 from showing beneath the box when 0 is entered. 

Documentation for [reference](https://react.dev/learn/conditional-rendering#logical-and-operator-) of why this was happening:
_Pitfall
Don’t put numbers on the left side of &&.
To test the condition, JavaScript converts the left side to a boolean automatically. However, if the left side is 0, then the whole expression gets that value (0), and React will happily render 0 rather than nothing._

## Images

Before 
image

After
image

## Accessibility

### DIFF
--- a/client/components/mma/upgrade/UpgradeSupportAmountForm.tsx
+++ b/client/components/mma/upgrade/UpgradeSupportAmountForm.tsx
@@ -228,7 +228,7 @@ export const UpgradeSupportAmountForm = ({
 							threshold={threshold}
 						/>
 					)}
-					{!continuedToConfirmation && !errorMessage && chosenAmount && (
+					{!continuedToConfirmation && !errorMessage && !!chosenAmount && (
 						<section css={buttonContainerCss}>
 							<ThemeProvider
 								theme={buttonThemeReaderRevenueBrand}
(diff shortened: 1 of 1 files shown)


client/components/mma/upgrade/UpgradeSupportAmountForm.tsx
```diff
@@ -228,7 +228,7 @@ export const UpgradeSupportAmountForm = ({
 							threshold={threshold}
 						/>
 					)}
-					{!continuedToConfirmation && chosenAmount && !errorMessage && (
+					{!continuedToConfirmation && !errorMessage && chosenAmount && (
```
andrewHEguardian, line 231:
```suggestion
					{!continuedToConfirmation && !errorMessage && !!chosenAmount && (
```

changing the order here seems to work but you could add this for clarity. To be clear, chosenAmount is still on the left side of this && with regard to the react element (but now the error message would be set when the amount is 0)

andrewHEguardian:
image
just tested this in the linked storybook, and if the first number you type is 0 you don't see the error message. I think this is because some of the null checking has been done wrong (by me) since 0 evaluates to falsy. Happy to pair on this

andrewHEguardian (approved):
👍  great spot

PR #1202, opened 2023-09-21
Upgrade Journey- add mechanism to prevent going back

andrewHEguardian:
## What does this change?

As with product switching and membership cancellation, we don't want people to press back in the browser and repeat their confirmation step (triggering an api call that will fail). To prevent this behaviour this PR adds a parameter in routerState that gets checked within the container

## How to test

Go to `/upgrade-support`, confirm change and on thank you page press browser back. Expect to see account overview.

## Accessibility


charleycampbell (approved):
Well remembered to do this! I've just tested it locally and it looks good

PR #1203, opened 2023-09-21
replace reference to `TEAMCITY_BUILD` with `GITHUB_RUN_NUMBER`

michaelbjacobson:
This PR is a follow-up from the Teamcity to Github Actions migration. It removes the final reference to Teamcity from the repo.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,26 +1,16 @@
-const path = require('path');
-const { merge } = require('webpack-merge');
-const AssetsPlugin = require('assets-webpack-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
-const GitRevisionPlugin = require('git-revision-webpack-plugin');
-const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
-const babelLoaderExcludeNodeModulesExcept = require('babel-loader-exclude-node-modules-except');
-const webpack = require('webpack');
-const nodeExternals = require('webpack-node-externals');
+import path from 'path';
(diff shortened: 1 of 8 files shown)


webpack.common.js
```diff
@@ -1,20 +1,21 @@
...
 	path: path.resolve(__dirname, './dist/'),
 });
 
 const definePlugin = new webpack.DefinePlugin({
-	WEBPACK_BUILD: process.env.TEAMCITY_BUILD
-		? `'${process.env.TEAMCITY_BUILD}'`
+	WEBPACK_BUILD: process.env.GITHUB_RUN_NUMBER
+		? `'${process.env.GITHUB_RUN_NUMBER}'`
 		: `'DEV_${new Date().getTime()}'`,
 	GIT_COMMIT_HASH: process.env.BUILD_VCS_NUMBER
```
andrewHEguardian, line 20: what is `BUILD_VCS_NUMBER`? is that also a teamcity thing

jamesmockett: Yeah, that's another TeamCity environment variable, although it has a fallback. The commit hash is used by prout to determine when a change is live: https://manage.theguardian.com/_prout

[`GITHUB_SHA`](https://docs.github.com/en/actions/learn-github-actions/variables) looks like a suitable replacement?

michaelbjacobson: Good catch, I just assumed it was something else entirely. `GITHUB_SHA` looks perfect!

andrewHEguardian (approved):
Looks good to me 👍

PR #1206, opened 2023-09-22
Minor fixes and reactors to upgrade journey

charleycampbell:
## What does this change?

- Removing unnecessary const from validateChoice and replacing occurrences of it with chosenAmount
- Removing null type from chosenAmount as this will never be null when we show the button to continue
- Checking that otherAmountSelected is not null when selecting the chosenAmount options. This means that if the rounded toggle is chosen £10 is selected and when untoggled 'Other' amount is reselected.
- Refactored both thank you pages to align with the css in the `ConfirmForm`

## Accessibility


client/components/mma/upgrade/UpgradeSupportAmountForm.tsx
```diff
@@ -232,27 +230,29 @@ export const UpgradeSupportAmountForm = ({
...
-									{mainPlan.billingPeriod}
-								</Button>
-							</ThemeProvider>
-						</section>
-					)}
+									<Button
+										cssOverrides={buttonCentredCss}
+										onClick={() => {
+											setContinuedToConfirmation(
+												chosenAmount ? true : false,
```
andrewHEguardian, line 244: I know this wasn't changed here but not sure this makes sense anymore. Why would chosenAmount be falsy if we click? 🤔

andrewHEguardian, line 244:
```suggestion
												true
```

andrewHEguardian (dismissed):
👍

andrewHEguardian (dismissed):

andrewHEguardian (approved):
Looks great! ✨

PR #1207, opened 2023-09-22
Upgrade Cypress

andrewHEguardian:
## What does this change?

Upgrades Cypress to the latest version. This fixes the previously failing github actions, which seemed to fail after a new ubuntu version using Chrome 117 ([reference](https://github.com/cypress-io/cypress/issues/19122)).

This migration was done following the automatic migration through the Cypress UI (`yarn cypress:open`)

## How to test

- [x] GHA Cypress tests pass
- [x] Tests pass locally

## Accessibility


tsconfig.json
```diff
@@ -16,5 +16,5 @@
     },
     "esModuleInterop": true
   },
-  "exclude": ["cypress/**", "cdk/**"]
+  "exclude": ["cypress/**", "cdk/**", "cypress.config.ts"]
```
andrewHEguardian, line 19: https://stackoverflow.com/a/72663546
as suggested by this answer (the jest tests and TS check failed without this exclusion)

.github/workflows/cypress.yml
```diff
@@ -39,4 +39,4 @@ jobs:
           wait-on-timeout: 30
           quiet: true
           browser: chrome
-          spec: cypress/integration/parallel-${{ matrix.group }}/*.ts
+          spec: cypress/e2e/parallel-${{ matrix.group }}/*.ts
```
charleycampbell, line 42: Why do we need to rename this path?

cypress.config.ts
```diff
@@ -0,0 +1,36 @@
...
+		runMode: 2,
+		openMode: 0,
+	},
+	e2e: {
+		// We've imported your old cypress plugins here.
+		// You may want to clean this up later by importing these.
+		setupNodeEvents(on, config) {
+			return require('./cypress/plugins/index.ts')(on, config);
+		},
+		baseUrl: 'http://localhost:9234/',
```
charleycampbell, line 34: looks like this is a new port to run on?

charleycampbell (commented):
Looks good - just a couple of minor questions!

.github/workflows/cypress.yml
andrewHEguardian: the migration renamed the folder, I guess because Cypress started referring to these type of tests as end to end rather than integration

cypress.config.ts
andrewHEguardian: image

looks same as before?

charleycampbell: hmm my storybook goes to `localhost:6006` Is it a different port?

PR #1210, opened 2023-09-25
Upgrade Support - Use next payment date from preview response after switch

andrewHEguardian:
## What does this change?

Fixes a bug in the Switch Thank You page after upgrading to S+. The page was using the original billing date instead of the new one after switching (which comes from the product move preview response.)

## How to test

Go to `/upgrade-support` and select above the threshold, see the next payment date 
image

Confirm and see the **same date** on the thank you page (ignore the wrong charge in the screenshots they're from separate tests)

image

## Accessibility


charleycampbell (approved):
Great spot!

PR #1218, opened 2023-10-02
OAuth migration | IDAPI and AAPI proxies

raphaelkabo:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1223, opened 2023-10-12
OAuth migration | Core OAuth authentication middleware

raphaelkabo:
This PR migrates the authentication middleware used by the MMA Express server from 'classic' (calling IDAPI's `/auth/redirect` endpoint on every request) to Okta-powered OAuth2.

> [!WARNING]  
> The new OAuth mechanism this PR introduces is **disabled** by setting `useOAuth: false` in the `okta-STAGE.json` config file stored in S3. It should **remain disabled** until the upcoming 'server-side Okta auth' PR is merged in, because without that PR, if OAuth were enabled, we would be introducing a security vulnerability in accessing certain APIs. We want to merge this PR into main though, to prevent needing to keep it updated with main and to spot any hidden issues early in testing. While the `useOAuth` flag is set to `false`, the authentication behaviour will be identical to how it has always worked. 

## Changes

- Added some new packages: `@okta/jwt-verifier` (to verify OAuth tokens), `openid-client` (to set up an OIDC client in MMA), and `ms` (because counting is hard and that's why we have computers).
- The `withIdentity` authentication middleware now calls one of two functions: `authenticateWithIdapi`, which is the old middleware, or `authenticateWithOAuth`, which is the new middleware. This is controlled by a `useOkta` boolean in the Okta config file stored in S3 and retrieved by `oktaConfig.ts`.
    -  See the table below for a comparison of the 'classic' middleware, the OAuth middleware, and the process we'll adopt after in-progress PRs are finished.
    - The OAuth middleware runs on each call MMA makes to its server, whether that's a frontend call or an API proxy request.
    - With the OAuth middleware, we're going for a **complete replication of the behaviour of the previous Identity middleware** - users shouldn't notice anything different, except that everyone will be logged out of MMA when we deploy this to PROD and enable `useOAuth`. (This is okay because MMA sessions are only valid for 30 minutes anyway).
    - When `useOAuth` is true, we're still dual-running this migration with the classic authorization behaviour for API endpoints handled by MMA. **When `useOAuth` is true, users will  _authenticate_ via OAuth, but all API calls will still be _authorized_ with the classic IDAPI cookies**.
    - **Future PRs will progressively migrate API calls until none of them use IDAPI cookies**.
- Added a handler for `/oauth/callback`, the callback route to which the browser returns after the user gets a valid session.
- Added Jest tests to provide as much coverage as possible for the new middleware, with the proviso that at this stage, testing the OAuth flow end-to-end in Cypress would require a large time commitment to set Cypress up to run on a local domain so that cookies work, and would also require us to use Identity endpoints to generate test users, which may be deprecated in the next year (see the Cypress setup in [Gateway](github.com/guardian/gateway) for an example of this).

Huge thanks to @coldlink (I borrowed a lot of the OAuth code from his work in Gateway) and @pvighi, who both helped a lot with bringing this PR together.

<table>
<thead>
  <tr>
    <th></th>
    <th>Before</th>
    <th>This PR</th>
    <th>After server-side auth PR</th>
    <th>After full migration</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td><strong>Cookies</strong></td>
    <td><code>SC_GU_U</code>, <code>SC_GU_LA</code>, <code>GU_U</code></td>
    <td><code>GU_ACCESS_TOKEN</code>, <code>GU_ID_TOKEN</code>, <code>GU_SO</code>, <code>SC_GU_U</code>, <code>SC_GU_LA</code>, <code>GU_U</code></td>
    <td>" - "</td>
    <td><code>GU_ACCESS_TOKEN</code>, <br><code>GU_ID_TOKEN</code>, <br><code>GU_SO</code>, <br><code>GU_U</code></td>
  </tr>
  <tr>
    <td rowspan="2"><strong>Session validity</strong></td>
    <td rowspan="2">SC_GU_LA generated in the last 30 minutes</td>
    <td><code>withOAuth: false</code>: SC_GU_LA generated in the last 30 minutes</td>
    <td rowspan="2">" - "</td>
    <td rowspan="2">OAuth tokens which have expiry of 30 minutes<br></td>
  </tr>
  <tr>
    <td><code>withOAuth: true</code>: OAuth tokens which have expiry of 30 minutes</td>
  </tr>
  <tr>
    <td rowspan="2"><strong>Auth middleware</strong></td>
    <td rowspan="2">Calls IDAPI to validate cookies</td>
    <td><code>withOAuth: false</code>: Calls IDAPI to validate cookies</td>
    <td rowspan="2"><code>withOAuth: true</code>: After local validation, calls Okta to validate tokens for API-token-based API routes</td>
    <td rowspan="2">Locally validates OAuth tokens, calls Okta to validate tokens for API-token-based API routes</td>
  </tr>
  <tr>
    <td><code>withOAuth: true</code>: Locally validates OAuth tokens</td>
  </tr>
  <tr>
    <td><strong>OAuth-enabled API routes (IDAPI, AAPI, MDAPI)</strong></td>
    <td>Are sent IDAPI cookies</td>
    <td>" - "</td>
    <td>" - "</td>
    <td>Are sent the OAuth access token</td>
  </tr>
  <tr>
    <td><strong>API token-based API routes (Lambdas, etc.)</strong></td>
    <td>Are sent an admin-level API token</td>
    <td>Are sent an admin-level API token</td>
    <td>We call Okta to validate tokens, then are sent an admin-level API token</td>
    <td>" - " - "</td>
  </tr>
</tbody>
</table>

A quick recap of the new authentication flow:

- If a public route is being requested, we validate any existing OAuth tokens. If they're valid, they're passed to the frontend to render 'logged in' components. We don't do anything else here.
- If we have a `GU_SO` cookie, we need to log the user out. We delete the access and ID token cookies and redirect the user to Gateway for authorization.
- If we have valid access and ID tokens, the user is logged in. We set the frontend state, and let the request proceed.
- If we don't have tokens, or they're invalid, we redirect the user to Gateway for authorization.
- After the OAuth flow completes, the browser returns to `/oauth/callback`. Here we exchange the token we receive for access and ID tokens from Okta, and call IDAPI to give us a fresh set of classic IDAPI cookies.
- Finally, we redirect the user to the page they were originally on. Simples!

## Reviews and testing

It might be easiest to go through this PR commit-by-commit, which I've done my best to make easy to do. There are a couple of support commits, one large commit introduces the two parts of the OAuth middleware flow, and then a commit with a suite of Jest tests.

We'll test this by deploying it on CODE and running some manual tests, as noted in this [testing plan](https://docs.google.com/spreadsheets/d/1jzbuA780ygI-4Z0hduLH8bPoH1kqOExX06JkrxXAItc/edit#gid=0). Then we'll do the same on PROD.

- [ ] Tested on CODE
- [ ] Tested on PROD

## Remaining work

- [X] **Gateway:** respect `maxAge` parameter to force reauth (https://github.com/guardian/gateway/pull/2444)
- [x] **This PR:** validate `clientId` when validating access and ID tokens

## Related PRs

- https://github.com/guardian/manage-frontend/pull/1220
- https://github.com/guardian/manage-frontend/pull/1218


server/middleware/identityMiddleware.ts
```diff
@@ -1,156 +1,158 @@
...
+		if (
+			// check access token is valid
+			accessToken &&
+			// check that the id token is valid
+			idToken &&
+			// check that the access token is not expired
+			!accessToken.isExpired() &&
+			// check that the scopes are all the ones we expect
+			accessToken.claims.scp?.every((scope) =>
+				scopes.includes(scope as Scopes),
```
pvighi, line 42: isn't this backwards ?
don't we want to check that every scope on `scopes` is in `accesToken.claims.scp` ?

server/oauth.ts
```diff
@@ -0,0 +1,362 @@
...
+	const { values } = cookies;
+	values.forEach(({ key, value }) => {
+		res.cookie(key, value, {
+			...idapiCookieOptions(key),
+		});
+	});
+};
+
+// Returns either an object with the verified access and ID tokens, or an empty object.
+export const verifyOAuthCookies = async (
```
coldlink, line 285: might be worth mentioning that this is doing local verification, especially if you'll later be adding in remote verification too

coldlink (changes requested):
looking good so far from my side, a few changes required though i think

server/oauth.ts
```diff
@@ -0,0 +1,362 @@
...
+ * @param res - the express response object
+ * @param options - the options for the authorization code flow
+ * @returns 303 redirect to the okta /authorize endpoint
+ */
+export const performAuthorizationCodeFlow = async (
+	_req: Request,
+	res: Response,
+	{
+		sessionToken,
+		scopes = ['openid'],
```
pvighi, line 181: to avoid confusion I find it useful to give them a more specific name for variables holding scopes lists. For example instead of using a `scopes` variable we would name it `requiredScopes` or `requestedScopes` or `tokenScopes` and things like that.

server/middleware/identityMiddleware.ts
```diff
@@ -1,70 +1,161 @@
...
-	res: express.Response,
-	redirectURL: string,
-	statusCode?: number,
-) =>
-	statusCode
-		? res.status(statusCode).header('Location', redirectURL).send()
-		: res.redirect(redirectURL);
+			const { accessToken, idToken } = await verifyOAuthCookies(req);
+			const idapiCookies = allIdapiCookiesSet(req);
+			if (idapiCookies && accessToken && idToken) {
```
pvighi, line 92: maybe we could get rid of `idapiCookies` here ? 

it's actually a boolean not the cookies themselves, I would suggest renaming it to `areIdapiCookiesSet` or something, but if it's only used in one place we might us well inline the call to allIdapiCookies set in the condition:
```typescript
 if (allIdapiCookiesSet(req) & accessToken && idToken)
```

server/oauth.ts
raphaelkabo: I think this is pretty clear from the docblock above the function, so I'll leave it as is.

server/middleware/identityMiddleware.ts
raphaelkabo: Good idea!

server/oauth.ts
raphaelkabo: Done!

coldlink (approved):
amazing work!

PR #1225, opened 2023-10-16
Fix Holiday Stops Amend

andrewHEguardian:
## What does this change?

Fix amending an existing holiday stop - this was not checking the correct value and therefore calling the api with a `POST` instead of a `PATCH` request. 

Fix a secondary bug when amending a non-confirmed holiday stop. The wrong dates were being set when this button was clicked, which made the holiday stop to amend appear as an existing suspension. (See image below.) This PR removes the second Amend button.

## How to test

With a newspaper delivery product, click Manage suspensions, click "Amend", pick dates, click "Confirm", see loading message "Amending...", do not see error.

## Images

Before
image
Bug after pressing "Amend" on review page, the existing suspension should not appear as that is what is being amended
image

After
image

## Accessibility


michaelbjacobson (approved):
Looks good to me!

PR #1237, opened 2023-11-01
OAuth migration | Serverside Okta validation middleware

pvighi:
## What does this change?

Co-authored by @pvighi and @raphaelkabo.

### Server-side token validation middleware

As part of the MMA OAuth migration, this PR introduces a new middleware, `withOktaServerSideValidation`, which verifies an Okta OAuth access token against the [Okta API's `/userinfo` endpoint](https://developer.okta.com/docs/reference/api/oidc/#userinfo). This endpoint returns user data if a user with a valid, active session exists for the supplied access token. More importantly, if the access token is not associated with a valid, active session (i.e. if the user has been signed out), the endpoint returns an error code.

The middleware is always invoked after the standard `withIdentity` middleware function. `withIdentity` performs a **local** check on the access token, validating that it hasn't been tampered with, that it isn't expired, and that its scopes are what we expect. The local check cannot verify that the user to whom the token is connected still has an active Okta session, although MMA tokens last only 30 minutes to compensate for this. `withOktaServerSideValidation` performs an additional **server-side** check, ensuring that the final part of the authorization demand - that the token is connected to a valid session - is fulfilled.

We don't need to make this additional check everywhere. We add the new middleware selectively to API routes which fulfil the following criteria:

- They call a downstream API with an admin-level client access token which may allow the API to perform operations on various users, not just the currently signed in user
- They fetch or modify sensitive user data, such as payment or subscription information

In these cases, we want to be absolutely sure that the user is authorized to perform the operation.

### Unit tests

We add Jest unit tests to cover the new middleware functions.

### Helper functions and tweaks

- We add some functions to abstract setting and fetching local request state (objects which live in Express in `res.locals`).
- The middleware functions now send messages via the logger library, rather than direct to console.
- We add and use the `zod` package to validate responses in the ID and access tokens

## Tests

- [x] Tested on CODE (with Okta disabled)
- [x] Tested on CODE (with Okta enabled)
- [ ] Smoke tested on PROD

### DIFF
--- a/server/middleware/oktaServerSideAuthMiddleware.ts
+++ b/server/middleware/oktaServerSideAuthMiddleware.ts
@@ -1,0 +1,112 @@
+import type { NextFunction, Request, Response } from 'express';
+import {
+	clearIdentityLocalState,
+	getIdentityLocalState,
+	setIdentityLocalState,
+} from '@/server/identityLocalState';
+import { OAuthAccessTokenCookieName } from '@/server/oauthConfig';
+import { validateWithOkta } from '@/server/okta';
+import { getConfig as getOktaConfig } from '@/server/oktaConfig';
+import { requiresSignin } from '@/shared/requiresSignin';
+import { log } from '../log';
+
+declare const CYPRESS: string;
+
+export interface OktaValidationResponse {
+	ok: boolean;
+	valid: boolean;
+	userId?: string;
+	displayName?: string;
+	email?: string;
+}
+
+export const withOktaServerSideValidation = async (
+	req: Request,
+	res: Response,
+	next: NextFunction,
+) => {
+	if (CYPRESS === 'SKIP_IDAPI') {
+		return next();
+	}
+	const oktaConfig = await getOktaConfig();
+	if (!oktaConfig.useOkta) {
+		/**
+		 * OKTA NOT ENABLED
+		 *
+		 * If Okta is disabled, we will still run this middleware, but
+		 * it won't be able to do anything so we just pass through.
+		 */
+		return next();
+	}
+
+	const signinRequired = requiresSignin(req.originalUrl);
+
+	const locallyValidatedIdentityData = getIdentityLocalState(res);
+	const accessToken = req.signedCookies[OAuthAccessTokenCookieName];
+	if (!accessToken || !locallyValidatedIdentityData?.userId) {
+		if (signinRequired) {
+			/**
+			 * NO TOKEN OR USER - SIGN IN REQUIRED
+			 *
+			 * This is unexpected and should be impossible because the withIdentity
+			 * middleware should have run first and not allowed the request to get
+			 * here. Return a 500.
+			 */
+			log.error(
+				`OAuth / Serverside Validation Middleware / Error: no access token or user in request for a sign-in required endpoint! This should have failed local validation.`,
+			);
+			return res.sendStatus(500);
+		} else {
+			/**
+			 * NO TOKEN OR USER - SIGN IN NOT REQUIRED
+			 *
+			 * This is expected - continue.
+			 */
+			return next();
+		}
+	}
+
+	const oktaValidationResponse = await validateWithOkta({
+		oktaConfig,
+		accessToken,
+	});
+
+	if (!oktaValidationResponse.ok) {
+		/**
+		 * UNEXPECTED ERROR
+		 */
+		return res.sendStatus(500);
+	}
+
+	if (!oktaValidationResponse.valid) {
+		/**
+		 * INVALID TOKEN
+		 *
+		 * Clear the local state. If the endpoint requires sign-in,
+		 * return a 401 which can be handled down the line. Otherwise
+		 * continue (the user will see the page as if they are not
+		 * signed in).
+		 */
+		clearIdentityLocalState(res);
+		if (signinRequired) {
+			return res.sendStatus(401);
+		}
+		return next();
+	}
+
+	/**
+	 * VALID TOKEN
+	 *
+	 * Update the local state with the latest user info from Okta.
+	 */
+	setIdentityLocalState(res, {
+		// This is always 'signedInRecently' because we've just checked
+		// the token is valid with Okta, and it's only valid for 30 minutes.
+		signInStatus: 'signedInRecently',
+		userId: oktaValidationResponse.userId,
+		displayName: oktaValidationResponse.displayName,
+		email: oktaValidationResponse.email,
+	});
+
+	return next();
+};
--- a/server/oauth.ts
+++ b/server/oauth.ts
@@ -1,28 +1,24 @@
 import crypto from 'crypto';
 import { joinUrl } from '@guardian/libs';
 import OktaJwtVerifier from '@okta/jwt-verifier';
-import type { CookieOptions, Request, Response } from 'express';
+import type { Request, Response } from 'express';
 import ms from 'ms';
 import type { Client, IssuerMetadata } from 'openid-client';
 import { generators, Issuer } from 'openid-client';
 import { conf } from '@/server/config';
+import { setIdentityLocalState } from '@/server/identityLocalState';
+import { log } from '@/server/log';
+import type { Scopes, VerifiedOAuthCookies } from '@/server/oauthConfig';
+import {
+	IdTokenClaims,
+	OAuthAccessTokenCookieName,
+	oauthCookieOptions,
+	OAuthIdTokenCookieName,
+	OAuthStateCookieName,
+	scopes,
+} from '@/server/oauthConfig';
 import type { OktaConfig } from '@/server/oktaConfig';
 import { getConfig as getOktaConfig } from '@/server/oktaConfig';
-
-export const OAuthAccessTokenCookieName = 'GU_ACCESS_TOKEN';
-export const OAuthIdTokenCookieName = 'GU_ID_TOKEN';
-export const OAuthStateCookieName = 'GU_oidc_auth_state';
-
-export interface VerifiedOAuthCookies {
-	accessToken?: OktaJwtVerifier.Jwt;
-	idToken?: OktaJwtVerifier.Jwt;
-}
-
-export const oauthCookieOptions: CookieOptions = {
-	signed: true,
-	secure: true,
-	httpOnly: true,
-};
 
 /**
  * @function getOktaOrgUrl
@@ -97,7 +93,7 @@ export const verifyAccessToken = async (token: string) => {
 		).verifyAccessToken(token, joinUrl(getOktaOrgUrl(oktaConfig), '/'));
 		return jwt;
 	} catch (error) {
-		console.error('OAuth / Access Token / Verification Error', error);
+		log.error('OAuth / Access Token / Verification Error', error);
 	}
 };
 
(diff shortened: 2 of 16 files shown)


server/middleware/OktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+	console.log('Validating token server side ...');
+	const signinRequired = () => requiresSignin(req.originalUrl);
+
+	const locallyValidatedIdentityData = getIdentityLocalState(res);
+	const accessToken = req.signedCookies[OAuthAccessTokenCookieName];
+	if (!accessToken || !locallyValidatedIdentityData?.userId) {
+		if (signinRequired()) {
+			console.error(
+				'error: no access token or user in request for a sign-in required endpoint! this should have failed local validation',
+			);
```
pvighi, line 44: I tried to format this as 
```typescript
		console.error(	'error: no access token or user in request for a sign-in required endpoint! this should have failed local validation');
```
but there's some autoformatting when I commit that really wants me to put in it in multiple lines and with a comma and an empty second argument

raphaelkabo: That's fine, it just keeps lines to under 80 or 100 characters or something.

server/oauth.ts
```diff
@@ -301,8 +302,8 @@ export const performAuthorizationCodeFlow = async (
 export const verifyOAuthCookiesLocally = async (
 	req: Request,
 ): Promise<VerifiedOAuthCookies> => {
-	const accessTokenCookie = req.signedCookies['GU_ACCESS_TOKEN'];
-	const idTokenCookie = req.signedCookies['GU_ID_TOKEN'];
+	const accessTokenCookie = req.signedCookies[OAuthAccessTokenCookieName];
+	const idTokenCookie = req.signedCookies[OAuthIdTokenCookieName];
```
pvighi, line 306: this is not strictly related to my change, I just noticed we had constants defined for this

server/oauth.ts
```diff
@@ -346,12 +347,27 @@ export const setLocalStateFromIdTokenOrUserCookie = (
...
+
+	if (typeof idToken?.claims.legacy_identity_id === 'string') {
+		identityIdClaim = idToken?.claims.legacy_identity_id;
+	}
+	if (typeof idToken?.claims.name === 'string') {
+		nameClaim = idToken?.claims.name;
+	}
+	if (typeof idToken?.claims.email === 'string') {
+		emailClaim = idToken?.claims.email;
+	}
```
pvighi, line 363: all of this code is jut because the claims are declared as type `unknown`, I don't know if there's a better way to just convert them to  optional strings

server/middleware/OktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+	const oktaResponse = await fetch(`${issuerUrl}/v1/userinfo/`, {
+		method: 'GET',
+		headers: {
+			'Content-Type': 'application/json',
+			Authorization: `Bearer ${accessToken}`,
+		},
+	});
+
+	console.log(`okta response status: ${oktaResponse.status}`);
+	if (oktaResponse.status == 200) {
```
raphaelkabo, line 66: Equality checks should always be strict (`===`)!

server/middleware/OktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+			return res.sendStatus(401);
+		}
+		return next();
+	} else {
+		//unexpected return status
+		console.error(
+			`unexpected status from okta userinfo ${oktaResponse.status}`,
+		);
+		return res.sendStatus(500);
+	}
```
raphaelkabo, line 89: If you wanted to turn the fetch into a method as we were discussing, you could do something like this:

```javascript

interface OktaValidationResponse {
	ok: boolean;
	valid: boolean;
	userId?: string;
	name?: string;
	email?: string;
}

const validateInOkta = async ({
	oktaConfig,
	accessToken
}: {
	oktaConfig: OktaConfig;
	accessToken: string;
}): OktaValidationResponse => {
	const issuerUrl = joinUrl(
		oktaConfig.orgUrl,
		'/oauth2/',
		oktaConfig.authServerId,
	);

	try {
		const oktaResponse = await fetch(`${issuerUrl}/v1/userinfo/`, {
			method: 'GET',
			headers: {
				'Content-Type': 'application/json',
				Authorization: `Bearer ${accessToken}`,
			},
		});
		if (oktaResponse.status === 200) {
			//valid token
			const oktaUserInfo = await oktaResponse.json<OktaUserInfo>();
			return {
				ok: true,
				valid: true,
				userId: oktaUserInfo.legacy_identity_id,
				name: oktaUserInfo.name,
				email: oktaUserInfo.email,
			};
		}  else if ([403, 401].includes(oktaResponse.status)) {
			//invalid token
			return { ok: true, valid: false };
		} else {
			//unexpected return status
			return { ok: false, valid: false };
		}
	} catch (error) {
		console.error(error);
		return { ok: false, valid: false };
	}
}

const oktaValidationResponse = await validateInOkta({
	oktaConfig,
	accessToken,
});

if (!oktaValidationResponse.ok) {
	return res.sendStatus(500);
}

if (!oktaValidationResponse.valid) {
	return res.sendStatus(401);
}

return next();
```

raphaelkabo (commented):
This is shaping up really nicely! I've left a few comments.

server/oauth.ts
raphaelkabo: Good spot!

server/middleware/OktaServerSideAuthMiddleware.ts
raphaelkabo: That's just the ESLint setup here, I guess.

server/middleware/OktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+		return next();
+	}
+	const oktaConfig = await getOktaConfig();
+	if (!oktaConfig.useOkta) {
+		console.log('Okta disabled, skipping Okta server side validation...');
+		return next();
+	}
+
+	console.log('Validating token server side ...');
+	const signinRequired = () => requiresSignin(req.originalUrl);
```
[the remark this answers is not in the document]
pvighi: I can do that and it probably won't be much of a difference, but I would be executing it always instead of just having a function that I can call if I need it, right ?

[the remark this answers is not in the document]
raphaelkabo: Well, it's a function which checks if a string begins with something so I'd say the performance hit is negligible. It just feels cleaner to me this way, doesn't make much of a difference.

server/IdentityLocalState.ts
```diff
@@ -0,0 +1,24 @@
+import type { Response } from 'express';
+
+//could this be unified with IdentityDetails in globalstate.ts ?
```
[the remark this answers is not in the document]
pvighi: I didn't mean to include this comment.  I decided against doing this at least in this PR because a lot of things are using res.locals.identity.  I would have to make sure nothing is using the old name  and  typescript won't help me here (as usual).
I can always do a text search and replace like a cowboy (or ever more reckless, a js developer)

server/oauth.ts
```diff
@@ -346,12 +347,27 @@ export const setLocalStateFromIdTokenOrUserCookie = (
...
+	setIdentityLocalState(res, {
 		signInStatus: hasIdTokenOrUserCookie ? 'signedInRecently' : undefined,
-		userId: idToken?.claims.legacy_identity_id,
-		name: idToken?.claims.name,
-		email: idToken?.claims.email,
-	};
+		userId: identityIdClaim,
+		name: nameClaim,
+		email: emailClaim,
+	});
```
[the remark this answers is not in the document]
pvighi: this doesn't work because `idToken.claims` doesn't have the actual claims defined in the type. 
I am trying to assigning to a string so I guess typescript has to know it's going to be a string ?
```
   server/oauth.ts:354:10 - error TS2339: Property 'legacy_identity_id' does not exist on type 'JwtClaims | undefined'.

    354  const { legacy_identity_id: userId, name, email } = idToken?.claims;
                 ~~~~~~~~~~~~~~~~~~
    server/oauth.ts:354:38 - error TS2339: Property 'name' does not exist on type 'JwtClaims | undefined'.

    354  const { legacy_identity_id: userId, name, email } = idToken?.claims;
                                             ~~~~
    server/oauth.ts:354:44 - error TS2339: Property 'email' does not exist on type 'JwtClaims | undefined'.

    354  const { legacy_identity_id: userId, name, email } = idToken?.claims;
                                                   ~~~~~

```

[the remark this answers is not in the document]
raphaelkabo: It looks like the TS error was just because `claims` could have been undefined; this should work:

```
	if (!idToken?.claims) {
		throw new Error('ID token is missing claims');
	}

	const { legacy_identity_id: userId, name, email } = idToken.claims;

	setIdentityLocalState(res, {
		signInStatus: hasIdTokenOrUserCookie ? 'signedInRecently' : undefined,
		userId,
		name,
		email,
	});
```

You may need to handle the `throw` upstream.

server/middleware/OktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,90 @@
+import { joinUrl } from '@guardian/libs';
+import type { NextFunction, Request, Response } from 'express';
+import fetch from 'node-fetch';
```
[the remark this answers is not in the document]
pvighi: I was just following the example of other  places we use fetch in this app, like in Api.ts:
https://github.com/guardian/manage-frontend/blob/c28f320713485888bd5778fb6c7ffcbb9e61a3cd/server/apiProxy.ts#L3-L5
or identityMiddleware:
https://github.com/guardian/manage-frontend/blob/c28f320713485888bd5778fb6c7ffcbb9e61a3cd/server/middleware/identityMiddleware.ts#L2-L4

server/middleware/oktaServerSideAuthMiddleware.ts
```diff
@@ -0,0 +1,168 @@
...
+		}
+	} catch (error) {
+		console.error(
+			`OAuth / Serverside Validation Middleware / Error: ${error.message}`,
+		);
+		return { ok: false, valid: false };
+	}
+};
+
+export const withOktaSeverSideValidation = async (
```
rupertbates, line 79: Typo in Server

server/routes/api.ts
```diff
@@ -163,27 +168,20 @@ router.post(
...
 );
 router.patch(
 	'/case/:caseId?',
+	withOktaServerSideValidation,
 	cancellationSfCasesAPI('case/:caseId', 'UPDATE_CANCELLATION_CASE', [
 		'caseId',
 	]),
 );
 
-router.post(
```
andrewHEguardian, line 175: possibly from a merge but not sure why this has been removed?

raphaelkabo: Yeah, I think this must have been from a merge but it definitely shouldn't have been deleted! Good catch, thank you.

server/oauth.ts
```diff
@@ -24,6 +27,13 @@ export const oauthCookieOptions: CookieOptions = {
 	httpOnly: true,
 };
 
+// Zod schema for the claims we expect to find in the ID token
+const IdTokenClaims = z.object({
+	legacy_identity_id: z.optional(z.string()),
+	name: z.optional(z.string()),
+	email: z.optional(z.string()),
+});
```
coldlink, line 35: this zod schema would pass for any given object/json passed into this as all the types are optional

server/okta.ts
```diff
@@ -0,0 +1,56 @@
...
+import { z } from 'zod';
+import { log } from './log';
+import type { OktaValidationResponse } from './middleware/oktaServerSideAuthMiddleware';
+import type { OktaConfig } from './oktaConfig';
+
+const OktaUserInfo = z.object({
+	legacy_identity_id: z.optional(z.string()),
+	email: z.optional(z.string()),
+	name: z.optional(z.string()),
+});
```
coldlink, line 11: same problem here: this zod schema would pass for any given object/json passed into this as all the types are optional

also this is the same as IdTokenClaims

coldlink (dismissed):

server/oauth.ts
```diff
@@ -346,12 +325,14 @@ export const setLocalStateFromIdTokenOrUserCookie = (
...
+	setIdentityLocalState(res, {
 		signInStatus: hasIdTokenOrUserCookie ? 'signedInRecently' : undefined,
-		userId: idToken?.claims.legacy_identity_id,
-		name: idToken?.claims.name,
-		email: idToken?.claims.email,
-	};
+		userId: result.success ? result.data.legacy_identity_id : undefined,
+		displayName: result.success ? result.data.name : undefined,
+		email: result.success ? result.data.email : undefined,
+	});
```
raphaelkabo, line 335: @coldlink I changed this after your review so that Zod doesn't throw when the `idToken` isn't set, and `setIdentityLocalState()` is called with the 'best possible' data - either the result of the Zod parse, or `undefined`.

andrewHEguardian (approved):
👍  this looks good to me from an SR perspective

coldlink (approved):

PR #1242, opened 2023-11-07
Remove message about covid-19 and inaccurate holidays

andrewHEguardian:
## What does this change?

Removes a message in newspaper voucher cancellation flow about covid-19, and an inaccurate number of holiday stops.

## How to test

Vouchers are only available in the UK, and self serve cancellation is not available in the UK...so hard to test.

## Images

Before
image

After
image

## Accessibility


daniel-guardian:
Looks good!

graham228221 (approved):
> ## How to test
> Vouchers are only available in the UK, and self serve cancellation is not available in the UK...so hard to test.

Good point! I was going to point out that the next paragraph probably also needs updating (we'd want to point people towards our own delivery service) but I doubt that anyone sees this page so probably not worth it.

PR #1247, opened 2023-11-13
Upgrade from guardian/commercial-core to guardian/commercial

emma-imber:
## What does this change?
Previously, a version of the **guardian/commercial-core** package was being used, which was 8 months old and is no longer maintained. This PR upgrades to use the latest version of **guardian/commercial**, which is the actively maintained package for the commercial code.


andrewHEguardian (approved):
👍

PR #1253, opened 2023-11-17
Create call centre numbers accordion component

andrewHEguardian:
## What does this change?

Create a new call centre numbers component using Source's `Accordion`. This is closer to the design in the figma 
image 

and has better accessibility than the existing component but has some drawbacks - we cannot open a section by default and we have to use a cssOverride to style the label.

## Images

New component:
image

image

Old component: 
image

## Accessibility


charleycampbell (approved):
This looks like a good way of using the source component. Great consideration around accessibility too 👍

PR #1256, opened 2023-11-20
Add `similar_guardian_products` to `marketingEmailConsents` list

coldlink:
## What does this change?

Adds the `similar_guardian_products` consent into the `marketingEmailConsents` list so it always shows up in on the emails and marketing page.

It's also no longer classified as a "product" consent, so we've removed from anywhere which said it was a product consent.

This consent used to be a "product" consent, so would previously only be given if a user has purchased a product. However this consent is being renamed, and repurposed into a new soft opt in for newly registered users also. So it's being moved out from being a "product" consent, and into the "main" marketing consents area. It will continue to use the same ID.

New copy:

**Guardian products and support**
Information on our products and ways to support and enjoy our independent journalism.

Other related work elsewhere:

Firstly Gateway to add it to the registration journey: https://github.com/guardian/gateway/pull/2455
And Identity API to rename the consent and description: https://github.com/guardian/identity/pull/2462


kelvin-chappell (dismissed):
👍

kelvin-chappell (approved):

kelvin-chappell (approved):

PR #1265, opened 2023-11-24
HACK: code test payment failure

andrewHEguardian:
add a hack for testing payment failure on `CODE` by inserting 'alertText' into the product in the MDAPI response in the server, which will make the product appear to be in payment failure

### DIFF
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -64,14 +64,22 @@ router.get(
 				),
 		)
 			.then((productDetails) => {
-				mdapiResponse.products = productDetails;
+				mdapiResponse.products = productDetails.map((productDetail) => {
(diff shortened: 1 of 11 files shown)


server/routes/api.ts
```diff
@@ -64,14 +64,22 @@ router.get(
 				),
 		)
 			.then((productDetails) => {
-				mdapiResponse.products = productDetails;
+				mdapiResponse.products = productDetails.map((productDetail) => {
+					productDetail.alertText = 'payment failure';
```
andrewHEguardian, line 68: this is the hack

andrewHEguardian:
didn't need this, build the build should still be in riff-raff in case we do

PR #1270, opened 2023-12-01
Add discount prices and number of months in save screens

andrewHEguardian:
## What does this change?

Adds the discounted and non-discounted prices to the digisub saves journeys, and how many months the discount applies for. Using data stored in a config object.

If the api endpoint can return this information we will use that instead, but it's yet to be built as of this PR.

## How to test

Digisub saves storybook screens

## Images

image

image

## Accessibility

### DIFF
--- a/client/components/mma/cancel/cancellationSaves/digipack/DigiSubDiscountConfirm.tsx
+++ b/client/components/mma/cancel/cancellationSaves/digipack/DigiSubDiscountConfirm.tsx
@@ -23,6 +23,12 @@ import {
 	buttonCentredCss,
 	stackedButtonLayoutCss,
 } from '@/client/styles/ButtonStyles';
+import {
+	getDiscountMonthsForDigisub,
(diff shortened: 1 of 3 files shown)


client/components/mma/cancel/cancellationSaves/digipack/DigiSubDiscountConfirm.tsx
```diff
@@ -135,7 +148,8 @@ export const DigiSubDiscountConfirm = () => {
 								</strong>
 								<br />
 								From {nextBillingDate}, your ongoing{' '}
-								{billingPeriod} payment will be {priceDisplay}.
+								{billingPeriod} payment will be {currencySymbol}
+								{formatAmount(discountedPrice)}.
```
andrewHEguardian, line 152: is this the wrong way round? not sure if the copy here is right, the figma doesn't have much there?

PR #1273, opened 2023-12-04
Remove node-fetch

andrewHEguardian:
## What does this change?

`fetch` is included in [Node 18](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility), so we potentially don't need to specifically use `node-fetch`. Testing out if we can remove this. 

## Accessibility


andrewHEguardian:
not sure about this - going to close until further investigation

PR #1274, opened 2023-12-05
OAuth migration | Add E2E tests for OAuth authentication

raphaelkabo:
## What does this change?

This adds a small suite of end-to-end (E2E) Cypress tests to manage-frontend. The existing Cypress tests all mock various API endpoints to allow testing specific user and supporter conditions. However to do this, they entirely skip the authentication middleware, and for our OAuth migration we really want to have that middleware tested end-to-end. Hence these new tests!

A lot of the changes here are refactoring file locations; most of the rest is support to allow end-to-end testing to work on local machines and the GHA runners. Specifically, we borrow from the general strategy in [Gateway](https://github.com/guardian/gateway), using Nginx configs to proxy the CODE locations of Gateway, IDAPI and MDAPI to `.thegulocal.com` domains. This allows a locally running MMA on `manage.thegulocal.com` to interact with, and read cookies set on, CODE services.

- We move the original mocked Cypress tests to `cypress/tests/mocked`, and rename the GitHub Actions workflow file to `cypress-mocked.yml`.
- The new tests live in `cypress/tests/e2e`, and use the `cypress-e2e.yml` workflow file.
- We add an Nginx config file, `cypress/cypress-nginx.conf`, which is used in GitHub Actions.
- We add a basic suite of E2E Cypress tests which we will expand on down the line, and some docs explaining how to set them up and run them locally.
- When running in Cypress (defined by the presence of the `RUNNING_IN_CYPRESS` environment variable), we skip metrics as these aren't authorised to be added in GHA, and always run the authentication middleware via OAuth/Okta.
- We change some Yarn commands and add some new ones:
    - `yarn cypress:mocked:server`: starts a dev server (essentially the same as the standard `yarn watch` command but with the `CYPRESS` variable set to `'SKIP_IDAPI'`, and doesn't open a new browser tab).
    -   `yarn cypress:mocked:open`: opens the Cypress test runner in mocked mode.
    -   `yarn cypress:mocked:run`: runs the Cypress tests headless in mocked mode.
    -   `yarn cypress:e2e:server`: starts a dev server without setting the `CYPRESS` variable.
    -   `yarn cypress:e2e:open`: opens the Cypress test runner in E2E mode.
    -   `yarn cypress:e2e:run`: runs the Cypress tests headless in E2E mode.

## Notes

- The E2E tests only run against the new, Okta-enabled authentication middleware, by overwriting the Okta config when Cypress is running.

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -1,0 +1,64 @@
+name: manage-frontend cypress (E2E)
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  cypress:
+    name: Manage-frontend Cypress (E2E)
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ secrets.MANAGE_FRONTEND_IAM_ROLE }}
+          aws-region: eu-west-1
+
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          cache: yarn
+
+      - name: Setup OS, Nginx, and Certs
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libnss3-tools
+          sudo service nginx restart
+          wget -q https://github.com/FiloSottile/mkcert/releases/download/v1.4.3/mkcert-v1.4.3-linux-amd64
+          wget -q https://github.com/guardian/dev-nginx/releases/latest/download/dev-nginx.tar.gz
+          sudo cp mkcert-v1.4.3-linux-amd64 /usr/local/bin/mkcert
+          sudo chmod +x /usr/local/bin/mkcert
+          sudo mkdir -p /usr/local/bin/dev-nginx
+          sudo tar -xzf dev-nginx.tar.gz -C /usr/local
+          sudo chmod +x /usr/local/bin/dev-nginx
+          sudo dev-nginx setup-cert "profile.thegulocal.com"
+          sudo dev-nginx setup-cert "manage.thegulocal.com"
+          sudo dev-nginx setup-cert "members-data-api.thegulocal.com"
+          sudo cp ./cypress/cypress-nginx.conf /etc/nginx/nginx.conf
+          sudo dev-nginx restart-nginx
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v6
+        env:
+          CYPRESS_IDAPI_CLIENT_ACCESS_TOKEN: ${{ secrets.IDAPI_CLIENT_ACCESS_TOKEN }}
+          # Required for the Cypress tests to run as we're unable to verify the created certs
+          NODE_TLS_REJECT_UNAUTHORIZED: 0
+        with:
+          start: yarn cypress:e2e:server
+          wait-on: 'https://manage.thegulocal.com'
+          wait-on-timeout: 30
+          quiet: true
+          browser: chrome
+          spec: cypress/tests/e2e/*.cy.ts
+
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -47,7 +47,9 @@ jobs:
         uses: cypress-io/github-action@v6
         env:
           CYPRESS_IDAPI_CLIENT_ACCESS_TOKEN: ${{ secrets.IDAPI_CLIENT_ACCESS_TOKEN }}
-          # Required for the Cypress tests to run as we're unable to verify the created certs
+          # This env variable prevents Node from rejecting self-signed TLS certificates. It's
+          # required for the Cypress tests to run as we're unable to verify the created certs.
+          # See: https://nodejs.org/api/cli.html#node_tls_reject_unauthorizedvalue
           NODE_TLS_REJECT_UNAUTHORIZED: 0
         with:
           start: yarn cypress:e2e:server
--- a/.github/workflows/cypress-mocked.yml
+++ b/.github/workflows/cypress-mocked.yml
@@ -31,8 +31,6 @@ jobs:
 
       - name: Cypress run
         uses: cypress-io/github-action@v6
-        env:
-          IDAPI_CLIENT_ACCESS_TOKEN: ${{ secrets.IDAPI_CLIENT_ACCESS_TOKEN }}
         with:
           start: yarn cypress:mocked:server
           wait-on: 'http://localhost:9234, http://localhost:9233'
--- a/.github/workflows/cypress-mocked.yml
+++ b/.github/workflows/cypress-mocked.yml
@@ -1,4 +1,4 @@
-name: manage-frontend cypress
+name: manage-frontend cypress (mocked)
 on:
   push:
     branches-ignore:
@@ -6,7 +6,7 @@ on:
 
 jobs:
   cypress:
-    name: Manage-frontend Cypress
+    name: Manage-frontend Cypress (mocked)
     runs-on: ubuntu-latest
 
     permissions:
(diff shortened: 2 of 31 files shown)


.github/workflows/cypress-mocked.yml
```diff
@@ -34,9 +34,10 @@ jobs:
         env:
           IDAPI_CLIENT_ACCESS_TOKEN: ${{ secrets.IDAPI_CLIENT_ACCESS_TOKEN }}
         with:
-          start: yarn cypress:server
+          start: yarn cypress:mocked:server
```
andrewHEguardian, line 37: I feel I should probably know this but why do the mocked tests need the IDAPI secret and IAM role?

cypress/lib/signInOkta.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const signInOkta = () => {
+	// When this function runs, the browser will already be showing the Gateway sign-in page
+	// because MMA will have redirected to it when it loads the first page of the test.
+	cy.setCookie('gu-cmp-disabled', 'true', {
+		domain: '.thegulocal.com',
+	});
+
+	// Necessary otherwise we get a 502 error back from MMA for some reason, perhaps a race condition?
+	// TODO: Make this not suck
```
andrewHEguardian, line 13: 🤣 great TODO

.github/workflows/cypress-mocked.yml
raphaelkabo: Good question - I don't think they do! I'm going to remove them and see if anything breaks.

raphaelkabo: Ah, I suspect the IAM role lets the server actually fetch the config it needs to run from S3.

raphaelkabo: Have removed the IDAPI secret and the mocked tests work just fine.

andrewHEguardian (commented):
As a Retention dev I'm happy with this 👍  but will let someone from Identity approve

.github/workflows/cypress-e2e.yml
```diff
@@ -0,0 +1,64 @@
...
+          sudo dev-nginx setup-cert "members-data-api.thegulocal.com"
+          sudo cp ./cypress/cypress-nginx.conf /etc/nginx/nginx.conf
+          sudo dev-nginx restart-nginx
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v6
+        env:
+          CYPRESS_IDAPI_CLIENT_ACCESS_TOKEN: ${{ secrets.IDAPI_CLIENT_ACCESS_TOKEN }}
+          # Required for the Cypress tests to run as we're unable to verify the created certs
+          NODE_TLS_REJECT_UNAUTHORIZED: 0
```
kelvin-chappell, line 51: Can you link to what this does?

cypress/cypress-nginx.conf
```diff
@@ -0,0 +1,131 @@
...
+    worker_connections  1024;
+}
+
+http {
+    include       mime.types;
+    default_type  application/octet-stream;
+
+    sendfile        on;
+
+    keepalive_timeout  65;
```
kelvin-chappell, line 18: Strange number!

cypress/cypress-nginx.conf
```diff
@@ -0,0 +1,131 @@
...
+			proxy_set_header 		Host members-data-api.code.dev-theguardian.com;
+			proxy_set_header        Accept-Encoding "";
+			proxy_hide_header 		Content-Security-Policy;
+
+			proxy_cookie_domain 	members-data-api.code.dev-theguardian.com members-data-api.thegulocal.com;
+			proxy_cookie_domain 	.code.dev-theguardian.com .thegulocal.com;
+
+			sub_filter_types        application/json;
+			sub_filter_once         off;
+			sub_filter              'members-data-api.code.dev-theguardian.com'  'members-data-api.thegulocal.com';
```
kelvin-chappell, line 76: Nitpick: Formatting's a bit inconsistent here

cypress/cypress-nginx.conf
```diff
@@ -0,0 +1,131 @@
...
+	# ======================
+    server {
+      listen 443 ssl;
+      server_name profile.thegulocal.com;
+      proxy_http_version 1.1; # this is essential for chunked responses to work
+
+      ssl_certificate profile.thegulocal.com.crt;
+      ssl_certificate_key profile.thegulocal.com.key;
+      ssl_session_timeout 5m;
+      ssl_protocols TLSv1.2 TLSv1.3;
```
kelvin-chappell, line 91: Why is this different to L59?

cypress/lib/signInOkta.ts
```diff
@@ -0,0 +1,24 @@
...
+	cy.wait(1000);
```
kelvin-chappell, line 14: Is there a smarter way to wait for a particular event?  I know that often with these kind of tests the answer is no.

cypress/tests/e2e/e2e.cy.ts
```diff
@@ -0,0 +1,97 @@
+import { signInOkta } from '../../lib/signInOkta';
+
+describe('E2E with Okta', () => {
+	beforeEach(() => {
+		cy.clearAllCookies();
+		signInOkta();
+		cy.get('h1', {
+			timeout: 30000,
```
kelvin-chappell, line 8: Seems like a very long timeout.  And could this value be extracted or do you foresee this varying independently in different places?

nginx/identity-frontend-CODE-fallback.conf
```diff
@@ -16,23 +14,83 @@ server {
...
+        proxy_set_header 		Upgrade $http_upgrade;
+        proxy_set_header 		Connection "Upgrade";
+		proxy_set_header 		"X-GU-Okta-Env" "profile.code.dev-theguardian.com";
+
+        ######
+        # remove `sid` cookie in requests to Gateway
+        # save original "Cookie" header value
+        set $altered_cookie $http_cookie;
+        # check if the "sid" cookie is present
+        if ($http_cookie ~ '(.*)(^|;\s)sid=("[^"]*"|[^\s]*[^;]?)(\2|$|;$)(?:;\s)?(.*)') {
```
kelvin-chappell, line 36: Maybe say where you got this regex from

nginx/identity-frontend-CODE-fallback.conf
```diff
@@ -16,23 +14,83 @@ server {
...
+          set $altered_cookie $1$4$5;
+        }
+        # hide original "Cookie" header
+        proxy_hide_header Cookie;
+        # set "Cookie" header to the new value
+        proxy_set_header  Cookie $altered_cookie;
+        ######
 
-        proxy_intercept_errors on;
+        proxy_intercept_errors 	on;
```
kelvin-chappell, line 46: Is this formatting aligning with something?

nginx/identity-frontend-CODE-fallback.conf
```diff
@@ -16,23 +14,83 @@ server {
...
+        proxy_set_header 		Host profile.code.dev-theguardian.com;
+        proxy_set_header 		Origin https://profile.code.dev-theguardian.com;
+        proxy_set_header        Accept-Encoding "";
+        proxy_hide_header 		Content-Security-Policy;
+        proxy_cookie_domain 	profile.code.dev-theguardian.com profile.thegulocal.com;
+        proxy_cookie_domain 	.code.dev-theguardian.com .thegulocal.com;
+
+        sub_filter_types        application/json;
+        sub_filter_once         off;
+        sub_filter              'profile.code.dev-theguardian.com'  'profile.thegulocal.com';
```
kelvin-chappell, line 94: Nitpick: Some inconsistent formatting

kelvin-chappell (commented):
I can understand a lot more of this than I expected and I'm relieved that a large part of it was references to moved code.  It's very clear and well structured.   Great work!

cypress/cypress-nginx.conf
raphaelkabo: Good catch! @coldlink we should remove `TLSv1` and `TLSv1.1` from all our local Nginx configs, it'll help debug mystery TLS errors.

cypress/tests/e2e/e2e.cy.ts
raphaelkabo: This turns out to be quite a safe value for the headless Cypress browser running in GitHub Actions, communicating through an Nginx proxy with the CODE APIs 🙃. It's _slow_. I've extracted it out to a Cypress.config call though, good idea.

cypress/cypress-nginx.conf
raphaelkabo: I've added a comment here. It _is_ a strange number.

nginx/identity-frontend-CODE-fallback.conf
raphaelkabo: Added.

nginx/identity-frontend-CODE-fallback.conf
raphaelkabo: Fixed!

kelvin-chappell (approved):
👍

kelvin-chappell (approved):

PR #1279, opened 2023-12-18
Integrate update to discount-api

andrewHEguardian:
## What does this change?

Update the response from discount-api as of [this PR](https://github.com/guardian/support-service-lambdas/pull/2133). No longer calculate the discount rate plan or number of months it's applicable for, but let the api tell us. The api also returns a 400 instead of a boolean when the subscription is ineligible for a discount.

## How to test

With a digisub, go to `/digital?withFeature=digisubSave` and cancel - should see discount including how many months it's valid for

## Accessibility


client/components/mma/cancel/cancellationSaves/digipack/DigiSubThankYouOffer.tsx
```diff
@@ -149,42 +142,39 @@ export const DigiSubThankYouOffer = () => {
...
 	const location = useLocation();
 	const routerState = location.state as CancellationRouterState;
 
-	const [isDiscountLoading, setIsDiscountLoading] = useState<boolean>(false);
+	const [isApplyDiscountLoading, setIsApplyDiscountLoading] =
+		useState<boolean>(false);
 	const [hasDiscountFailed, setHasDiscountFailed] = useState<boolean>(false);
 
-	if (loadingState == LoadingState.IsLoading) {
+	const [isPreviewDiscountLoading, setIsPreviewDiscountLoading] =
```
andrewHEguardian, line 153: I removed the asyncLoader as it will handle a 400 response as an error and log it as such - maybe that's fine? But given we expect this endpoint to return 400s for some users I thought we shouldn't do that

charleycampbell: Good point. How do differentiate the difference between the 400s we expect and actual "bad" 400s?

charleycampbell: Great shout to remove another AsyncLoader!

charleycampbell (approved):

client/components/mma/cancel/cancellationSaves/digipack/DigiSubThankYouOffer.tsx
andrewHEguardian: don't think so - slight risk of getting a 404 if the URLs change but any validation errors I don't think we need to log

PR #1282, opened 2023-12-20
Remove `any` from `useAsyncLoader`

andrewHEguardian:
## What does this change?

Trying to refactor to make sure we have no `any`s in the the `useAsyncLoader` hook. 

## Accessibility


client/utilities/hooks/useAsyncLoader.ts
```diff
@@ -39,7 +39,7 @@ export function useAsyncLoader<T>(
 			asyncFetch()
 				.then((response) => responseProcessor(response))
 				.then((data) => {
-					setData(data);
+					setData(data as T);
```
andrewHEguardian, line 42: this is basically the crux of the refactor - using `as` to assert it's the type, since we're not parsing it at all

tjmw: Would it make sense to add some validation here before casting to `T`? Is there a pattern for doing that in this codebase already? We use zod in some other projects.

andrewHEguardian: yes, absolutely. Seemed like a step too far for this PR - but I think zod was just added to the codebase recently

PR #1297, opened 2024-01-19
Revert "OAuth migration | Remove SC_GU_LA cookie check | Migrate MDAPI calls to use OAuth tokens"

coldlink:
Reverts guardian/manage-frontend#1294

This broke MMA when using IDAPI cookies, so reverting this change.

The `authorizationOrCookieHeader` function only works in Okta, and does not take into account cases where Okta is disabled! So when Okta was turned off, it broke as it was trying to use non-existant okta tokens.


kelvin-chappell (approved):
👍

PR #1311, opened 2024-02-07
OAuth migration | Use new set username route

raphaelkabo:
## What does this change?

As part of the OAuth migration, we've removed an old bug in IDAPI - being able to set your username more than once, if you happened to know the API endpoint to hit (because setting username just used `POST /user/me`). We've already introduced a new IDAPI endpoint, `POST /user/me/username`, which only allows setting username once and only if it's not been set before. Once OAuth authorization to IDAPI is enabled in MMA, we will also be preventing the username being changed via `POST /user/me`. This PR modifies the MMA frontend to use the new endpoint, and adds an E2E test for username updating.

## Dependencies

- https://github.com/guardian/identity/pull/2492, because we need to let IDAPI create test users without a preset username before the E2E tests here pass.

## Tests

- [x] Tested on CODE

### DIFF
--- a/client/components/mma/identity/publicProfile/PublicProfile.tsx
+++ b/client/components/mma/identity/publicProfile/PublicProfile.tsx
@@ -60,7 +60,9 @@ export const PublicProfile = (_: { path?: string }) => {
 	const usernameDisplay = (u: User) => (
 		<>
 			<WithStandardTopMargin>
-				<PageSection title="Username">{u.username}</PageSection>
+				<PageSection title="Username">
+					<span data-cy="username-display">{u.username}</span>
+				</PageSection>
 			</WithStandardTopMargin>
 			<WithStandardTopMargin>
 				<Lines n={1} />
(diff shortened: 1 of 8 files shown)


client/components/mma/identity/publicProfile/PublicProfile.tsx
```diff
@@ -63,7 +60,9 @@ export const PublicProfile = (_: { path?: string }) => {
 	const usernameDisplay = (u: User) => (
 		<>
 			<WithStandardTopMargin>
-				<PageSection title="Username">{u.username}</PageSection>
+				<PageSection title="Username">
+					<span data-cy="username-display">{u.username}</span>
```
pvighi, line 64: just curious why do we need this span here now? is it just for cypress?

pvighi (approved):
looks good to me

client/components/mma/identity/publicProfile/PublicProfile.tsx
raphaelkabo: Yeah, it's just for Cypress. Makes it much easier to target it.

PR #1312, opened 2024-02-15
OAuth migration | Stop sending Cookies header

raphaelkabo:
## Changes

We stop sending the `Cookies` header to upstream APIs called by MMA. At last, we are free of the terror of IDAPI cookies.


kelvin-chappell (approved):
🚀

PR #1323, opened 2024-03-14
use cmp from libs

vlkerag:
## What does this change?
The cmp has moved to libs. Hence pointing to libs instead of cmp.

## Accessibility


rBangay (approved):
👍  Thanks for looking into this

PR #1329, opened 2024-04-05
Use new product-switch-api for recurring contribution to supporter plus

rupertbates:
## What does this change?
This PR changes manage-frontend so that it uses the [new product-switch api](https://github.com/guardian/support-service-lambdas/pull/2201) for recurring contribution to supporter plus product switches.

There is another type of product switch from membership to recurring contribution, but this is not currently handled by the new api and so will still call the old one.

### DIFF
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -10,6 +10,7 @@ import {
 	holidayStopAPI,
 	invoicingAPI,
 	productMoveAPI,
+	productSwitchAPI,
 } from '../apiGatewayDiscovery';
 import {
 	customMembersDataApiHandler,
@@ -194,7 +195,7 @@ router.post(
 router.post(
 	'/product-move/:switchType/:subscriptionName',
 	withOktaServerSideValidation,
-	productMoveAPI(
+	productSwitchAPI(
 		'product-move/:switchType/:subscriptionName',
 		'MOVE_PRODUCT',
 		['switchType', 'subscriptionName'],
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -198,7 +198,7 @@ router.post(
 	'/product-move/to-recurring-contribution/:subscriptionName',
 	withOktaServerSideValidation,
 	productMoveAPI(
-		'product-move/:switchType/:subscriptionName',
+		'product-move/to-recurring-contribution/:subscriptionName',
 		'MOVE_PRODUCT',
 		['switchType', 'subscriptionName'],
 	),
@@ -208,7 +208,7 @@ router.post(
 	'/product-move/recurring-contribution-to-supporter-plus/:subscriptionName',
 	withOktaServerSideValidation,
 	productSwitchAPI(
-		'product-move/:switchType/:subscriptionName',
+		'product-move/recurring-contribution-to-supporter-plus/:subscriptionName',
 		'MOVE_PRODUCT',
 		['switchType', 'subscriptionName'],
 	),
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -200,7 +200,7 @@ router.post(
 	productMoveAPI(
 		'product-move/to-recurring-contribution/:subscriptionName',
 		'MOVE_PRODUCT',
-		['switchType', 'subscriptionName'],
+		['subscriptionName'],
 	),
 );
 // recurring contribution to supporter plus is using the new api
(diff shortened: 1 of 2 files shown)


server/routes/api.ts
```diff
@@ -1,15 +1,16 @@
 import * as Sentry from '@sentry/node';
 import { Router } from 'express';
-import { featureSwitches } from '../../shared/featureSwitches';
-import type { MembersDataApiResponse } from '../../shared/productResponse';
-import { isProduct, MDA_TEST_USER_HEADER } from '../../shared/productResponse';
+import { featureSwitches } from '@/shared/featureSwitches';
+import type { MembersDataApiResponse } from '@/shared/productResponse';
+import { isProduct, MDA_TEST_USER_HEADER } from '@/shared/productResponse';
```
johnduffell, line 5: what does `@` do?

johnduffell (approved):
good and I don't want to minimise the effort you put in, but a small point is I don't think the difference between "move" -> "switch" captures the difference between the two APIs.
It reminds me of the change from identity-play-auth to identity-auth-play (or was it the other way around!)
I can see the plausible deniability attraction of not having any overt estra stuff in the naming, but I think versioning it might have been clearer if we don't want to put "ts" in the name etc.

server/routes/api.ts
```diff
@@ -191,13 +192,25 @@ router.post(
...
-		'product-move/:switchType/:subscriptionName',
+		'product-move/to-recurring-contribution/:subscriptionName',
 		'MOVE_PRODUCT',
-		['switchType', 'subscriptionName'],
+		['subscriptionName'],
+	),
+);
+// recurring contribution to supporter plus is using the new api
+router.post(
+	'/product-move/recurring-contribution-to-supporter-plus/:subscriptionName',
```
tjmw, line 208: Super minor - would it be worth pulling this route string out to a const so it can be referred to here and in the call to `productSwitchAPI`? They seem intrinsically linked looking at this and the other cases.

tjmw (approved):
Nice!

server/routes/api.ts
rupertbates: It's just some syntactic sugar to make imports neater

rupertbates:
> good and I don't want to minimise the effort you put in, but a small point is I don't think the difference between "move" -> "switch" captures the difference between the two APIs. It reminds me of the change from identity-play-auth to identity-auth-play (or was it the other way around!) I can see the plausible deniability attraction of not having any overt estra stuff in the naming, but I think versioning it might have been clearer if we don't want to put "ts" in the name etc.

I'm not sure I understand what you mean, the api takes exactly the same inputs and gives the same outputs, what differences are you thinking of?

server/routes/api.ts
jamesgorrie: It's 2¢ on this, as I know we use it everywhere, but this does add "one-more-thing-you-need-to-know" and usually has it's own bespoke config which adds cognitive overhead over using standard config.

jamesgorrie (approved):
Looks nice and simple and easy to understand.

server/routes/api.ts
tjmw: Those all sound like fine reasons to leave it alone!

server/routes/api.ts
rupertbates: Maybe? On the other hand it's quite a lot easier to read, the relative imports can look very messy and hard to parse.
I do agree the less magic approach is generally a good idea though

johnduffell: yeah I'm not complaining, (as a non CS expert) I'm happy for us to use it, I was actually mainly worried it was an accidental change as it didn't seem releated to the main PR.

rupertbates: It was giving me a warning so I changed it

rupertbates:
> > > good and I don't want to minimise the effort you put in, but a small point is I don't think the difference between "move" -> "switch" captures the difference between the two APIs. It reminds me of the change from identity-play-auth to identity-auth-play (or was it the other way around!) I can see the plausible deniability attraction of not having any overt estra stuff in the naming, but I think versioning it might have been clearer if we don't want to put "ts" in the name etc.
> > 
> > 
> > I'm not sure I understand what you mean, the api takes exactly the same inputs and gives the same outputs, what differences are you thinking of?
> 
> I just mean I'd rather call it product-move-api-v2 or product-move-api-ts, then the name makes it clear which is which. Otherwise we might have the constant problem where someone mentions "product-something-api" and we have to try to remember which is which. AFAIK the main (only?) difference is one is TS and the other is scala.
> 
> However it's a minor point compared with everything else so I don't want to overstate it!

Ah I see what you mean, I'll just need to crack on and decommission the old version then!

PR #1331, opened 2024-04-11
content security policy audit

rBangay:
## What does this change?
Adds a content security policy report header to all requests so that we can audit which endpoints need adding to a future content-security-policy header

as per docs at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only

### Example of csp report in cloudwatch logs
Screenshot 2024-04-19 at 13 04 22

### DIFF
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -1,5 +1,6 @@
 import * as Sentry from '@sentry/node';
 import { Router } from 'express';
+import { featureSwitches } from '../../shared/featureSwitches';
 import type { MembersDataApiResponse } from '../../shared/productResponse';
 import { isProduct, MDA_TEST_USER_HEADER } from '../../shared/productResponse';
 import {
@@ -339,10 +340,12 @@ router.get(
 router.post('/reminders/cancel', cancelReminderHandler);
 router.post('/reminders/reactivate', reactivateReminderHandler);
 
-router.post('/csp-audit-report-endpoint', (req, res) => {
-	const parsedBody = JSON.parse(req.body.toString());
-	log.warn(JSON.stringify(parsedBody));
-	res.status(204).end();
-});
+if (featureSwitches.cspSecurityAudit) {
+	router.post('/csp-audit-report-endpoint', (req, res) => {
+		const parsedBody = JSON.parse(req.body.toString());
+		log.warn(JSON.stringify(parsedBody));
+		res.status(204).end();
+	});
+}
 
 export { router };
--- a/server/routes/api.ts
+++ b/server/routes/api.ts
@@ -339,4 +339,10 @@ router.get(
 router.post('/reminders/cancel', cancelReminderHandler);
 router.post('/reminders/reactivate', reactivateReminderHandler);
 
+router.post('/csp-audit-report-endpoint', (req, res) => {
+	const parsedBody = JSON.parse(req.body.toString());
+	log.warn(JSON.stringify(parsedBody));
+	res.status(204).end();
+});
+
 export { router };
--- a/server/routes/core.ts
+++ b/server/routes/core.ts
@@ -8,11 +8,6 @@ import { log } from '../log';
 
 const router = Router();
 
-router.get('/csp-audit-report-endpoint', (req: Request, res: Response) => {
-	log.info(`CSP_AUDIT: ${req.body}`);
-	res.status(204).send();
-});
-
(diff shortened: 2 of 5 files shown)


server/routes/core.ts
```diff
@@ -8,6 +8,11 @@ import { log } from '../log';
 
 const router = Router();
 
+router.get('/csp-audit-report-endpoint', (req: Request, res: Response) => {
+	log.info(`CSP_AUDIT: ${req.body}`);
+	res.status(202).send();
```
johnduffell, line 13: nit pick
```suggestion
	res.status(204).send();
```

server/routes/core.ts
```diff
@@ -8,6 +8,11 @@ import { log } from '../log';
 
 const router = Router();
 
+router.get('/csp-audit-report-endpoint', (req: Request, res: Response) => {
+	log.info(`CSP_AUDIT: ${req.body}`);
```
johnduffell, line 12: sometimes cloudwatch is a bit irritating and splits multi line log statements across more than one log line.  It does have smarts to spot json, but if we hit this problem, we can always indent subsequent lines (which should help)

server/server.ts
```diff
@@ -39,6 +39,14 @@ if (conf.DOMAIN === 'thegulocal.com') {
 
 server.use(helmet());
 
+server.use(function (_: Request, res: Response, next: NextFunction) {
+	res.setHeader(
+		'Content-Security-Policy-Report-Only',
+		'report-to /csp-audit-report-endpoint',
```
johnduffell, line 45: the documentation is inconsistent about how report-to works https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to
it sounds like report-uri has recently been deprecated so it may not be completely up to date in the docs?

rBangay: I've updated it to send both report-uri and report-to ... which should give us the best browser coverage (report-uri is ignored for browsers that support report-to apparently, also firefox at the time of writing doesn't support report-to)

tjmw (approved):
Nice, this looks great 👍  I've not used `Content-Security-Policy-Report-Only` before so it's cool to see this in practice.

At the moment the policy added here is specifying https for all content and reporting violations, is the next step to add the more exhaustive policy defined in #1330 (but report only, not enforce)?

johnduffell:
> At the moment the policy added here is specifying https for all content and reporting violations, is the next step to add the more exhaustive policy defined in #1330 (but report only, not enforce)?

Good question - when we discussed it we were imagining to start from zero and only add the ones we saw used, because we didn't even know why some of them were there, and they may only be relevant to acquisition (and may not even be relevant there  ay more - after all, who would think to delete things?)
But once we see some results we could take a judgement call in the name of efficiency?

tjmw (approved):
This looks great!

One thought I had about the potential noise from the new endpoint - does MMA have an existing feature flag or switch mechanism? If so I wonder if it'd be worth wrapping the log line in a switch that we can easily toggle the logging back and forth without deploying. No worries if not.

rBangay:
> This looks great!
> 
> One thought I had about the potential noise from the new endpoint - does MMA have an existing feature flag or switch mechanism? If so I wonder if it'd be worth wrapping the log line in a switch that we can easily toggle the logging back and forth without deploying. No worries if not.

Good question ... there are feature switches of sort in manage-frontend, but they are not controlled by RRCP or any sort of CMS, they're just variables in a feature switch ts file. I'll put the audit stuff behind a switch though 👍

server/routes/api.ts
```diff
@@ -339,4 +340,12 @@ router.get(
 router.post('/reminders/cancel', cancelReminderHandler);
 router.post('/reminders/reactivate', reactivateReminderHandler);
 
+if (featureSwitches.cspSecurityAudit) {
+	router.post('/csp-audit-report-endpoint', (req, res) => {
+		const parsedBody = JSON.parse(req.body.toString());
+		log.warn(JSON.stringify(parsedBody));
```
johnduffell, line 346: looking at the screenshot in the description, it ends up as double nested JSON.  Is there a way to inline this JSON? or at least prettyPrint i.e. 2 space indent etc?
No worries if not but it might help!

johnduffell (approved):
fab I'm impressed how in depth that was / how much we learnt! 👍

PR #1332, opened 2024-05-07
Add basic CSP to manage-frontend

johnduffell:
At the moment we are trying to report on CSP violations, however there are huge numbers of vialoations being logged for manage.theguardian.com, which should definitely be allowed.  In the first 1000 results we mostly could only find manage.theguardian.com.

this PR adds 'self' as an allowed source which should hopefully reduce the noise.  Also sourcepoint.

After this we can do more PRs to add other domains based on the results from cloudwatch logs.


rBangay (approved):
👍

PR #1333, opened 2024-05-09
Supporter plus cancellation offer journey

rBangay:
## What does this change?

Offers supporter plus customers 2 months free as an incentive not to cancel their subscription.

#### New offer page
Screenshot 2024-06-19 at 09 45 25

#### New offer review page
Screenshot 2024-06-19 at 09 44 39

#### New offer confirmation page
Screenshot 2024-06-19 at 09 43 44

#### New cancellation confirm page (for supporter plus product)
Screenshot 2024-06-19 at 09 42 17


client/components/mma/cancel/CancellationReasonSelection.tsx
```diff
@@ -70,16 +72,27 @@ const ReasonPicker = ({
...
-					{ title: 'Confirmation' },
-				]}
-				additionalCSS={css`
-					margin: ${space[5]}px 0 ${space[12]}px;
-				`}
-			/>
+			{featureSwitches.supporterplusCancellationOffer &&
+			productType.productType === 'supporterplus' ? (
+				<ProgressStepper
+					steps={[{ isCurrentStep: true }, {}, {}, {}]}
```
tjmw, line 78: Just for my understanding - why don't any of these steps have a `title` property? (Like in the example on the other side of the ternary).

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
```diff
@@ -0,0 +1,302 @@
...
+	const mainPlan = getMainPlan(productDetail.subscription);
+
+	const offerPeriodWord = number2words(routerState.upToPeriods);
+	const offerPeriodType = routerState.upToPeriodsType.toLowerCase();
+	const nextNonDiscountedPaymentDate = parseDate(
+		routerState.nextNonDiscountedPaymentDate,
+		'yyyy-MM-dd',
+	).dateStr();
+
+	const standfirstCss = css`
```
tjmw, line 61: Do these emotion styles need to be defined in the function body of the component? I'm sure performance wise it's no problem, but would it make the component clearer if these were pulled outside (as they don't need to be defined on every render)? Obvs feel free to ignore, especially if this is the convention in this repo!

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
```diff
@@ -0,0 +1,302 @@
...
+					<div
+						css={css`
+							max-width: 290px;
+						`}
+					>
+						<BenefitsSection
+							small
+							benefits={[
+								{
+									name: '',
```
tjmw, line 244: I wonder if the `name` attribute should be made optional (instead of having to pass empty strings)?

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
```diff
@@ -0,0 +1,302 @@
...
+					cssOverrides={cancelButtonCss}
+					onClick={() => {
+						navigate('/');
+					}}
+				>
+					Return to your account
+				</Button>
+			</div>
+			<p css={termsCss}>
+				Lorem ipsum dolor sit amet, consec tetur adipiscing elit. Nam
```
tjmw, line 294: Don't forget to update this before merging!

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOfferReview.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+interface RouterSate extends DiscountPreviewResponse {
+	selectedReasonId: OptionalCancellationReasonId;
+	cancellationPolicy: string;
+	caseId: string;
+	holidayStops?: OutstandingHolidayStop[];
+	deliveryCredits?: DeliveryRecordDetail[];
+}
+
+type OfferApiCallStatus = 'NOT_READY' | 'PENDING' | 'FAILED' | 'SUCCESS';
```
tjmw, line 39: ❤️  this pattern 👍

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOfferReview.tsx
```diff
@@ -0,0 +1,251 @@
...
+					cssOverrides={ctaBtnCss}
+					onClick={() => {
+						navigate('..');
+					}}
+				>
+					Go back
+				</Button>
+			</div>
+			<p css={termsCss}>
+				Lorem ipsum dolor sit amet, consec tetur adipiscing elit. Nam
```
tjmw, line 243: Another TODO to update this before merging!

shared/numberUtils.ts
```diff
@@ -0,0 +1,27 @@
+const num =
+	'zero one two three four five six seven eight nine ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen'.split(
+		' ',
+	);
+const tens = 'twenty thirty forty fifty sixty seventy eighty ninety'.split(' ');
+
+export const number2words = (n: number): string => {
```
tjmw, line 7: Would it be worth throwing an error explicitly if a number < 0 is ever passed (and also a non-whole number) to make it clear that's not supported? (Sadly I don't think TS can help here!)

tjmw, line 7: Maybe premature at this point but this feels like it could belong in [`@guardian/libs`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/libs).

client/components/mma/cancel/CancellationReasonSelection.tsx
rBangay: The new designs for the cancellation flow don't have titles above each of the stages in the step indicator

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
rBangay: yeah, that's a good shout re re render. I'll move them out :+1:

shared/numberUtils.ts
rBangay: hmm, I'm not sure whether to throw an error here or not.

It does look like you can enforce non negative numbers in typescript using generics and template literals like this:

```
type NonNegative = number & { __nonNegative: true };
type AssertPositive<N extends number> = number extends N ? NonNegative : `${N}` extends `-${string}` ? never : N;

function sqrt<N extends number>(n: AssertPositive<N>): number {
    return Math.sqrt(n);
}
```

However, the compiler is not clever enough to do math on the function input .. eg 
`sqrt(5) // this is fine`
`sqrt(-5) // this will give a type error`
`sqrt(5 + 5) // uh ohh, the compiler doesn't know what to do with this and will give a type error`
this can be "fixed" by casting the input using as:
`sqrt(5 + 5 as NonNegative)`
but that's possibly not all that safe as you could just as easily cast a negative number and fool typescript

rBangay: I've settled for now on adding some documentation above the function declaration and adding some out of bounds function input to the unit/jest tests

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOfferReview.tsx
```diff
@@ -0,0 +1,267 @@
...
+					}),
+				},
+			);
+
+			if (response.ok) {
+				navigate('../offer-confirmed', {
+					state: routerState,
+				});
+			} else {
+				setPerformingDiscountStatus('FAILED');
```
tjmw, line 166: Maybe one for the future (and totally just my opinion) - if a non-okay response is always treated the same way as an error I think it can be a nice pattern to put that logic close to the `fetch` call - i.e. return a rejected promise if the response isn't okay. I think it can simplify the calling code, here we wouldn't need the condition in the try block, the `catch` block would handle both types of errors so the `setPerformingDiscountStatus('FAILED');` wouldn't need to be repeated.

shared/productResponse.ts
```diff
@@ -188,12 +188,14 @@ export interface Subscription {
...
 	plan?: PaidSubscriptionPlan; // this is used for memberships (remove when memberships no longer exist)
 	trialLength: number;
+	inDiscountPeriod?: boolean;
 	readerType: ReaderType;
 	deliveryAddress?: DeliveryAddress;
 	contactId?: string;
 	account?: {
 		accountName: string;
 	};
+	//cancellationOfferIsActive?: boolean;
```
tjmw, line 198: Should this be removed (instead of commented) if it's no longer needed?

client/components/mma/shared/benefits/BenefitsSection.tsx
```diff
@@ -8,32 +10,62 @@ import { benefitsCss, unavailableBenefitsCss } from './BenefitsStyles';
...
+				benefitsCss,
+				small &&
+					css`
+						${textSans15}
+					`,
+			]}
+		>
+			{benefits.map((benefit, benefitIndex) => (
+				<Benefit
+					key={`benefit-${benefitIndex}`}
```
tjmw, line 65: I guess it's okay to use the array index as the key here since we don't expect the order to change? (Otherwise I think that might confuse React?)

client/components/mma/shared/ProgressIndicator.tsx
```diff
@@ -9,7 +9,7 @@ interface Step {
 }
 
 interface ProgressIndicatorProps {
-	steps: [Step, Step, Step];
+	steps: [Step, Step, ...Step[]];
```
tjmw, line 12: This isn't `Step[]` because there are guaranteed to always be _at least_ 2 steps?

tjmw (approved):
Great!

I left a handful of (non-blocking) comments/questions. I don't know manage-frontend super well so might be worth getting another ✅  from someone who does!

client/components/mma/shared/ProgressIndicator.tsx
rBangay: yes exactly, I wanted it to be an array of steps with some guards around the number of elements

client/components/mma/shared/benefits/BenefitsSection.tsx
rBangay: I don't think the order will change nope

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOfferReview.tsx
rBangay: yeah I think I agree, although maybe this can be addressed in a subsequent pr

tjmw: yeah absolutely, not one for now 👍

PR #1335, opened 2024-05-10
CSP report header allow list additions

rBangay:
## What does this change?

Add more values to the csp report header in order to make the cloudwatch logs less noisy and narrow in on a final csp header


tjmw (approved):
Nice 👍

PR #1343, opened 2024-06-05
csp header css update

rBangay:
## What does this change?

Change invalid `default-style` to the correct `style-src` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

This will allow inline styles for `<style>` and `<link>` tags as well as inline `style` attribute.

This is not a welcome attribute but a necessary one in order to implement a csp header in manage. Hopefully at a later date we can look into using a hash or nonce to validate specific inline sources.

### DIFF
--- a/server/server.ts
+++ b/server/server.ts
@@ -53,7 +53,6 @@ if (featureSwitches.cspSecurityAudit) {
 		'report-to csp-endpoint',
 		`default-src ${cspDefaultSrcAllowList}`,
 		`style-src 'unsafe-inline'`, // this is unsafe but needed for now for emotion
-		`style-src-elem 'unsafe-inline'`, // this is unsafe but needed for now for emotion
 	];
 	server.use(function (_: Request, res: Response, next: NextFunction) {
 		res.set({
--- a/server/server.ts
+++ b/server/server.ts
@@ -52,7 +52,8 @@ if (featureSwitches.cspSecurityAudit) {
 		'report-uri /api/csp-audit-report-endpoint',
 		'report-to csp-endpoint',
 		`default-src ${cspDefaultSrcAllowList}`,
-		`default-style 'unsafe-inline'`, // this is unsafe but needed for now for emotion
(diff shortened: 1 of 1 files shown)


server/server.ts
```diff
@@ -52,7 +52,8 @@ if (featureSwitches.cspSecurityAudit) {
 		'report-uri /api/csp-audit-report-endpoint',
 		'report-to csp-endpoint',
 		`default-src ${cspDefaultSrcAllowList}`,
-		`default-style 'unsafe-inline'`, // this is unsafe but needed for now for emotion
+		`style-src 'unsafe-inline'`, // this is unsafe but needed for now for emotion
+		`style-src-elem 'unsafe-inline'`, // this is unsafe but needed for now for emotion
```
johnduffell, line 56: you don't need this line because
"If this directive is absent, the user agent will look for the style-src directive"

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src-elem#:~:text=If%20this%20directive%20is%20absent%2C%20the%20user%20agent%20will%20look%20for%20the%20style%2Dsrc%20directive

johnduffell:
```suggestion
```

server/server.ts
```diff
@@ -52,7 +52,8 @@ if (featureSwitches.cspSecurityAudit) {
 		'report-uri /api/csp-audit-report-endpoint',
 		'report-to csp-endpoint',
 		`default-src ${cspDefaultSrcAllowList}`,
-		`default-style 'unsafe-inline'`, // this is unsafe but needed for now for emotion
```
johnduffell, line 55: no idea where we got `default-style` from 😕

johnduffell (approved):
see comment for suggestion - this csp malarky is hard!

tjmw (approved):
LGTM too! (Pending John's comment about being able to remove `style-src-elem`).

server/server.ts
rBangay: my fault I think I read the csp report wrong 😖

PR #1345, opened 2024-06-19
Make sure that the production build uses the current lockfile

rupertbates:
## What does this change?

Up until now the final production build was being run with no lockfile. This meant that the versions of dependencies which end up being used in production can be different from those used during development. 

This PR fixes that issue.

We actually found this because of an issue in the okta library, where a newer version of a dependency broke the library, this PR has more information: https://github.com/guardian/gateway/pull/2775

### DIFF
--- a/script/ci.sh
+++ b/script/ci.sh
@@ -14,9 +14,10 @@ yarn test
 yarn bundle
 
 cp package.json dist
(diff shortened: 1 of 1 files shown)


tjmw (approved):
👍

PR #1346, opened 2024-06-24
remove subscriberId property

rBangay:
## What does this change?
Removes the 'subscriberId' property from teh product response interface and any references to it in fixture mocks ... This property hasn't been used in manage for a while now (we use `subscriptionId` instead).

This was removed from members-data-api here:  https://github.com/guardian/members-data-api/pull/1086


tjmw (approved):
👍

johnduffell (approved):
nice! I'm sure there are a few more properties we can look at improving as well if we want to do some health work on this response.

PR #1349, opened 2024-06-26
Revert "Add Value as CODEOWNERS"

johnduffell:
Remove CODEOWNERS because it requests reviews from everyone in the team automatically as soon as anyone raises a PR.  This is a problem because it's spammy and also it's no longer clear whether PRs are ready for review.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-* @guardian/value


ecguardian (approved):
👍

tjmw:
> also it's no longer clear whether PRs are ready for review.

I think if the PR is opened as draft then a review isn't requested, could that be a way to mitigate this (if we ever have to put CODEOWNERS back)?

johnduffell:
> > also it's no longer clear whether PRs are ready for review.
> 
> I think if the PR is opened as draft then a review isn't requested, could that be a way to mitigate this (if we ever have to put CODEOWNERS back)?

good point
image

and then this happens
image

But the builds still run

The problem would be convincing people to do that in reality!

PR #1350, opened 2024-06-28
fix conditional check

rBangay:
## What does this change?

Fix conditional check that would fail if the left hand side of the condition returned a falsey value (the right hand side of the condition would try to access a property of undefined)


tjmw (approved):
👍

PR #1352, opened 2024-07-02
copy update for update amount SP form (with old min amount)

rBangay:
## What does this change?
If you have a supporter plus at an old price that is below the new minumum, the update amount form should reflect that in the copy re the new minimum amount you can change it to

## Images
Before             |  After
  |

### DIFF
--- a/client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
+++ b/client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
@@ -312,7 +312,6 @@ export const SupporterPlusUpdateAmountForm = (
 							>
 								<TextInput
 									label={otherAmountLabel}
-									data-cy="supporter-plus-other-amount-input"
 									error={
 										(shouldShowOtherAmountErrorMessage &&
 											errorMessage) ||
--- a/client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
+++ b/client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
@@ -80,7 +80,13 @@ function validateChoice(
 	} else if (!chosenAmount || isNaN(chosenOptionNum)) {
 		return 'There is a problem with the amount you have selected, please make sure it is a valid amount';
 	} else if (!isNaN(chosenOptionNum) && chosenOptionNum < minAmount) {
-		return `${mainPlan.currency}${minAmount} per ${mainPlan.billingPeriod} is the minimum payment to receive this subscription. Please call our customer service team to lower your ${monthlyOrAnnual} amount below ${mainPlan.currency}${minAmount} via the Help Centre`;
+		return `${mainPlan.currency}${minAmount} per ${
+			mainPlan.billingPeriod
+		} is the ${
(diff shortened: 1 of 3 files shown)


client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
```diff
@@ -304,6 +312,7 @@ export const SupporterPlusUpdateAmountForm = (
 							>
 								<TextInput
 									label={otherAmountLabel}
+									data-cy="supporter-plus-other-amount-input"
```
tjmw, line 315: We can't target this using the label? (Feels a bit closer to how a human would interact with this).

client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
```diff
@@ -355,17 +364,21 @@ export const SupporterPlusUpdateAmountForm = (
 								size="medium"
 							/>
 							<p>
-								If you would like to lower your{' '}
-								{monthlyOrAnnual.toLowerCase()} amount below{' '}
-								{minPriceDisplay} please call us via the{' '}
+								If you would like to{' '}
+								{currentAmountIsBelowNewMin
```
tjmw, line 368: Would `change` work in both cases here? Only ask because it would simplify the code a bit!

cypress/tests/mocked/parallel-1/updateContributionAmount.cy.ts
```diff
@@ -98,6 +101,10 @@ describe('Update contribution amount', () => {
 
 		cy.findByText('Change amount').click();
 
+		cy.contains(
+			/£\d{2,3} per month is the minimum payment to receive this subscription./,
```
tjmw, line 105: Are we fuzzy matching the amount here to account for future config changes? (I.e. so this test doesn't have to be updated each time the config changes?)

tjmw (approved):
Looks good 👍  I left a few questions, just to help my understanding really!

cypress/tests/mocked/parallel-1/updateContributionAmount.cy.ts
rBangay: yeah I fuzzy matched here to reduce the number of places that you would have to update if/when we do another price rise ... there are still places where you have to do it manually though, so this doesn't solve that 'problem' everywhere, only here.

client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
rBangay: yeah possibly .. I took a steer from the stakeholders re the copy here:
Screenshot 2024-07-03 at 11 35 20

client/components/mma/accountoverview/updateAmount/SupporterPlusUpdateAmountForm.tsx
rBangay: yeah good shout, I aggree ... the new selector in the test for the input is a bit more involved but I guess the benefit is that we're not polluting the dom and we're closer to how a user would navigate the site

PR #1353, opened 2024-07-02
remove doNotSetUsername param as we no longer set username for test users

johnduffell:
This PR follows on when this one https://github.com/guardian/identity/pull/2528 is merged

It undoes some of the changes in https://github.com/guardian/manage-frontend/pull/1311
This is possible as we no longer set usernames for test users, so we don't need a flag to prevent that.

That flag is removed in the identity PR https://github.com/guardian/identity/pull/2528 as part of the changes to move the test user token from the username/firstname to the email address

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -97,7 +97,6 @@ type IDAPITestUserOptions = {
 	password?: string;
 	deleteAfterMinute?: boolean;
 	isGuestUser?: boolean;
-	doNotSetUsername?: boolean;
 };
 type IDAPITestUserResponse = [
 	{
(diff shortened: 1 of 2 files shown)


cypress/support/commands.ts
```diff
@@ -152,12 +150,11 @@ export const createTestUser = ({
 					password: finalPassword,
 					deleteAfterMinute,
 					isGuestUser,
-					doNotSetUsername,
 				} as IDAPITestUserOptions,
 			})
 			.then((res) => {
 				return cy.wrap({
-					emailAddress: finalEmail,
+					emailAddress: res.body.emailAddress as string,
```
johnduffell, line 157: this is similar to this PR https://github.com/guardian/gateway/pull/2791

we need to get the email address from the response, as the email address will be augmented with the test user token

rBangay (approved):
👍

PR #1354, opened 2024-07-03
ws peer dependency fix

rBangay:
## What does this change?
security fix: force `ws` peer dependency to use a patched/fixed version

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3900,26 +3900,6 @@
     "@jridgewell/resolve-uri" "^3.1.0"
     "@jridgewell/sourcemap-codec" "^1.4.14"
 
-"@jsonjoy.com/base64@^1.1.1":
-  version "1.1.2"
-  resolved "https://registry.yarnpkg.com/@jsonjoy.com/base64/-/base64-1.1.2.tgz#cf8ea9dcb849b81c95f14fc0aaa151c6b54d2578"
-  integrity sha512-q6XAnWQDIMA3+FTiOYajoYqySkO+JSat0ytXGSuRdq9uXE7o92gzuQwQM14xaCRlBLGq3v5miDGC4vkVTn54xA==
-
-"@jsonjoy.com/json-pack@^1.0.3":
-  version "1.0.4"
-  resolved "https://registry.yarnpkg.com/@jsonjoy.com/json-pack/-/json-pack-1.0.4.tgz#ab59c642a2e5368e8bcfd815d817143d4f3035d0"
-  integrity sha512-aOcSN4MeAtFROysrbqG137b7gaDDSmVrl5mpo6sT/w+kcXpWnzhMjmY/Fh/sDx26NBxyIE7MB1seqLeCAzy9Sg==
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -16592,27 +17125,10 @@ write-file-atomic@^4.0.2:
...
-  version "8.14.2"
-  resolved "https://registry.yarnpkg.com/ws/-/ws-8.14.2.tgz#6c249a806eb2db7a20d26d51e7709eab7b2e6c7f"
-  integrity sha512-wEBG1ftX4jcglPxgFCMJmZ2PLtSbJ2Peg6TmpJFTbe9GZYOQCDPdMYu/Tm0/bGZkw8paZnJY45J4K2PZrLYq8g==
-
-ws@^8.13.0, ws@^8.2.3:
-  version "8.13.0"
-  resolved "https://registry.yarnpkg.com/ws/-/ws-8.13.0.tgz#9a9fb92f93cf41512a0735c8f4dd09b8a1211cd0"
-  integrity sha512-x9vcZYTrFPC7aSIbj7sRCYo7L/Xb8Iy+pW0ng0wt2vCJv7M9HOMy0UoN3rr+IFC7hb7vXoqS+P9ktyLLLhO+LA==
+ws@>=7.5.10, ws@>=8.17.1, ws@^6.1.0, ws@^7.3.1, ws@^8.11.0, ws@^8.13.0, ws@^8.2.3:
+  version "8.17.1"
```
tjmw, line 17129: Maybe I shouldn't be worrying too much about yarn internals but I'm confused about how `8.17.1` satisfies all the versions listed above. E.g. `ws@^6.1.0` doesn't seem like it should be satisfied with `8.17.1`.

rBangay: It's a good point, I may have been too greedy with the globbing in the `resolutions` block in package.json .. I'll see if I can be more specific where multiple peer dependency versions are used

tjmw: Ah right sorry I hadn't quite appreciated that change in the `resolutions` block for some reason! Yeah could be worth being more specific (it looks like both v7 and v8, which are affected, have a patch available). So it might be just a case of making sure any v7 or v8 in the project are on the latest version of that major release. Having said all that, if it works then 👍

rBangay:
> Could be worth co-ordinating with @michaelbjacobson as he's working on a major storybook upgrade.

I think I've fixed the specific major versions the peer dependencies should use. It looks like you can be way more specific in the resolutions block, which hopefully will make Michaels life a lot easier

rBangay:
> > Could be worth co-ordinating with @michaelbjacobson as he's working on a major storybook upgrade.
> 
> I think I've fixed the specific major versions the peer dependencies should use. It looks like you can be way more specific in the resolutions block, which hopefully will make Michaels life a lot easier

Michaels pr went out first in the end, so I've rebased this branch and made a slight change (webpack-dev-server had been updated and no longer needs the version of ws it uses to be patched)

tjmw (approved):

PR #1355, opened 2024-07-08
ProgressIndicator and ProgressStepper seperate stories

rBangay:
## What does this change?
Seperates out the ProgressIndicator (older) and ProgressStepper (new) components into seperate stories. Quick tidy up in order to help with the eventual aim of replacing the ProgressIndicator component entirely.

## Images
Screenshot 2024-07-08 at 17 19 23

### DIFF
--- a/client/components/mma/shared/ProgressIndicator.stories.tsx
+++ b/client/components/mma/shared/ProgressIndicator.stories.tsx
@@ -10,37 +10,6 @@ export default {
 	},
 } as Meta<typeof ProgressStepper>;
 
-export const Default: StoryFn<typeof ProgressStepper> = () => {
-	return (
-		<>
-			<ProgressStepper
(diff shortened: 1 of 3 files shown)


client/components/mma/shared/ProgressIndicator.stories.tsx
```diff
@@ -10,37 +10,6 @@ export default {
 	},
 } as Meta<typeof ProgressStepper>;
```
tjmw, line 11: 👀  Should this `ProgressStepper` reference still be here?

tjmw (approved):

client/components/mma/shared/ProgressIndicator.stories.tsx
rBangay: ooo, no it shouldn't!

PR #1356, opened 2024-07-10
CSP script hashes report only

rBangay:
## What does this change?
Rework csp report header to work with inline script hashes and `strict-dynamic` to allow those inline scripts to load further scripts

#### More rational around constructing a csp header for an SPA can be found here:

https://auth0.com/blog/deploying-csp-in-spa/

https://web.dev/articles/strict-csp#adopting-a-strict-csp

### DIFF
--- a/server/server.ts
+++ b/server/server.ts
@@ -41,11 +41,13 @@ if (conf.DOMAIN === 'thegulocal.com') {
 server.use(helmet());
 
 export const createCsp = (hashes: string[]) => {
-	return `script-src ${hashes
-		.map((hash) => `'sha256-${hash}'`)
-		.join(
-			' ',
-		)} 'strict-dynamic'; style-src 'unsafe-inline'; object-src 'none'; base-uri 'none';`;
+	const prefixedHashes = hashes.map((hash) => `'sha256-${hash}'`);
+	const csp = [
+		`script-src ${prefixedHashes.join(' ')} 'strict-dynamic'`,
+		`style-src 'unsafe-inline'`,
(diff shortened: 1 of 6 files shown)


server/server.ts
```diff
@@ -40,29 +40,28 @@ if (conf.DOMAIN === 'thegulocal.com') {
...
-				'{ "group": "csp-endpoint", "endpoints": [ { "url": "/api/csp-audit-report-endpoint" } ] }',
-			'Content-Security-Policy-Report-Only': `${csp.join('; ')};`,
-		});
-		next();
+export const createCsp = (hashes: string[]) => {
+	return `script-src ${hashes
+		.map((hash) => `'sha256-${hash}'`)
+		.join(
+			' ',
+		)} 'strict-dynamic'; style-src 'unsafe-inline'; object-src 'none'; base-uri 'none';`;
```
johnduffell, line 48: I preferred the old way of building it up and then joining it
```suggestion
	const prefixedHashes = hashes.map((hash) => `'sha256-${hash}'`);
	const csp = [
		`script-src ${prefixedHashes.join(' ')} 'strict-dynamic'`,
		`style-src 'unsafe-inline'`,
		`object-src 'none'`,
	];
	return csp.join('; ');
```

johnduffell (approved):
looks good, I've made a minor comment but it doesn't really affect the functionality 👍

server/server.ts
rBangay: Yep, that's nicer :)

PR #1359, opened 2024-07-12
Create newspaper archive endpoint

andrewHEguardian:
An authentication endpoint to give users access to newspapers.com archive of the Guardian using their third party authentication endpoint.

Creates a new route at `/newspaperArchive/auth` that redirects to the URL returned by their API.

This relies on a new S3 secret for the auth key, which I've uploaded for DEV, CODE and PROD

## Why are we doing this?

We want to test this as a benefit to the Tier 3 product, initially the user experience will be a link within MMA only for those with this product. For the test we won't be checking the product in the endpoint but at the client side logic rendering the link.

### DIFF
--- a/server/routes/newspaperArchive.ts
+++ b/server/routes/newspaperArchive.ts
@@ -3,6 +3,16 @@ import type { Request, Response } from 'express';
 import fetch from 'node-fetch';
 import { withIdentity } from '../middleware/identityMiddleware';
 
+type NewspapersRequestBody = {
+	expires?: number; // defaults to 24 hours
+	'query-string'?: string;
+};
+
+// { url: "https://<subdomain>.newspapers.com/…?tpa=<token>" }
+type NewspapersResponseBody = {
+	url: string;
+};
+
 function base64(input: string) {
 	return Buffer.from(input).toString('base64');
 }
@@ -12,13 +22,9 @@ router.use(withIdentity(401));
 
 router.get('/auth', async (_req: Request, res: Response) => {
 	const subdomain = 'theguardian';
-	const authKey = process.env.newspaperArchive;
+	const authKey = '';
 	const authHeader = base64(`${subdomain}:${authKey}`);
-	const requestBody = {
-		expires: 86400, // optional, defaults to 24 hours,
-		'query-string':
-			'xid=1234&utm_campaign=awesome-campaign&utm_medium=referral&utm_source=editorial&utm_content=&utm_term=',
-	};
+	const requestBody: NewspapersRequestBody = {};
 
 	const response = await fetch(
 		'https://www.newspapers.com/api/userauth/public/get-tpa-token',
@@ -30,9 +36,9 @@ router.get('/auth', async (_req: Request, res: Response) => {
 			method: 'POST',
 			body: JSON.stringify(requestBody),
 		},
-	); // { url: "https://<subdomain>.newspapers.com/…?tpa=<token>" }
+	);
 
-	const responseJson = await response.json();
+	const responseJson = (await response.json()) as NewspapersResponseBody;
 	res.redirect(responseJson.url);
 });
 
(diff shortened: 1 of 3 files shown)


server/routes/newspaperArchive.ts
```diff
@@ -0,0 +1,69 @@
...
+> = s3ConfigPromise<NewspaperArchiveConfig>('authKey')('newspaper-archive');
+
+let authKey: string;
+
+const router = Router();
+
+router.use(withIdentity(401));
+
+router.get('/auth', async (_req: Request, res: Response) => {
+	const subdomain = 'theguardian';
```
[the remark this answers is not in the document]
johnduffell: it says subdomain but it's actually the user name that's used in the Authorisation header on line 57.
I think it's one of those things that's private rather than secret.  So it could give someone a good starting point for attacks, but they couldn't complete them using that information (unless e.g. the password was accidentally blank or something)
I assume authKey has been generated using a secure key generator so I wouldn't personally be concerned, aside from the noise in the PRs from the scanner.
An easy third option to save extra coding would be to store the authKey in S3 as `theguardian:passwordhere` and rename it to authString or similar.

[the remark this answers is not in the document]
andrewHEguardian: That third option seems like a sensible action 👍

server/routes/newspaperArchive.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	if (authKey === undefined) {
+		const config = await newspaperArchiveConfigPromise;
```
johnduffell, line 40: this isn't in a synchronised block so I assume it can cope with the expected number of requests all hitting this promise at the same time?  Or is the `newspaperArchiveConfigPromise` a singleton anyway? in which case we could just get rid of the guard completely?

johnduffell:
> An authentication endpoint to give users access to newspapers.com archive of the Guardian using their third party authentication endpoint.
> 
> Creates a new route at `/newspaperArchive/auth` that redirects to the URL returned by their API.
> 
> This relies on a new S3 secret for the auth key, which I've uploaded for DEV, CODE and PROD

the description explains the how, but not really the why.

So it sounds to me like we want anyone to hit manage.theguardian.com/auth (from an email link? or further work to put it into manage homepage?) while logged in to identity and it will let them in to view the newspaper via the third party?  Or do they have to have a particular product to be allowed in?

server/routes/newspaperArchive.ts
andrewHEguardian: Ah, you're right it's unnecessary. The S3 config gets requested when the server starts up. I'm not entirely sure how?

andrewHEguardian:
> > An authentication endpoint to give users access to newspapers.com archive of the Guardian using their third party authentication endpoint.
> > Creates a new route at `/newspaperArchive/auth` that redirects to the URL returned by their API.
> > This relies on a new S3 secret for the auth key, which I've uploaded for DEV, CODE and PROD
> 
> the description explains the how, but not really the why.
> 
> So it sounds to me like we want anyone to hit manage.theguardian.com/newspaperArchive/auth (from an email link? or further work to put it into manage homepage?) while logged in to identity and it will let them in to view the newspaper via the third party? Or do they have to have a particular product to be allowed in?

I've edited the description to answer the Why a bit

server/routes/newspaperArchive.ts
```diff
@@ -0,0 +1,61 @@
...
+const router = Router();
+
+router.use(withIdentity(401));
+
+router.get('/auth', async (_req: Request, res: Response) => {
+	const config = await newspaperArchiveConfigPromise;
+	const authString = config?.authString;
+	if (authString === undefined) {
+		log.error(`Missing newspaper archive auth key`);
+		res.status(500).send();
```
andrewHEguardian, line 39: Possibly not the user experience we'd actually want

johnduffell: see what you mean but the main thing for this MVP is that we get an alarm straight away if it happens

johnduffell (approved):
looks good, and nice idea for T3 (could even get people to take out a newspaper subscription on the back of it!)

PR #1363, opened 2024-07-17
restrict switch journey for supporter plus subs

rBangay:
### What does this PR change?
Stop people with supporter plus subscriptions from using the contribution to supporter plus switch journey

### Images
Screenshot 2024-07-17 at 17 53 35

This is just a little clientside nicety. It seems as though we were getting some traffic to the /switch route from users with a supporter plus subscription, which shouldn't be possible unless they have manually gone to this url, instead of letting them attempt a failed switch we can show the user an error message and send them off to the account overview.

### DIFF
--- a/client/components/mma/switch/options/SwitchOptions.tsx
+++ b/client/components/mma/switch/options/SwitchOptions.tsx
@@ -130,6 +130,32 @@ export const SwitchOptions = () => {
 
 	const navigate = useNavigate();
 
+	if (contributionToSwitch.tier === 'Supporter Plus') {
+		return (
+			<section css={sectionSpacing}>
+				<ErrorSummary
+					cssOverrides={errorSummaryOverrideCss}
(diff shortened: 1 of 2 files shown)


tjmw (approved):
Looks good 👍 

(Not sure if worth waiting for an approval from John as he's got more domain knowledge here than me).

client/components/mma/switch/options/SwitchOptions.tsx
```diff
@@ -130,6 +130,32 @@ export const SwitchOptions = () => {
 
 	const navigate = useNavigate();
 
+	if (contributionToSwitch.tier === 'Supporter Plus') {
```
johnduffell, line 133: why not say if it's not a contribution, if that's the only one we can switch?

johnduffell (approved):
great work, nice bit of belt and braces! 👍

PR #1366, opened 2024-07-29
Top tier copy changes

rBangay:
### What does this PR change?
Copy changed for the top tier (digital + print) product, which was modelled after the Guardian weekly product.

Minor refactor of the cancellation confirmation page to remove unnecessary product detail providing closure.


client/fixtures/productBuilder/baseProducts.ts
```diff
@@ -532,15 +532,15 @@ export function baseTierThree(): ProductDetail {
 				country: 'United Kingdom',
 			},
 			safeToUpdatePaymentMethod: true,
-			start: '2024-06-28',
-			end: '2025-06-13',
+			start: '2021-12-24',
```
michaelwmcnamara, line 535: Out of curiosity, why are these set to dates in the past?

tjmw (approved):
LGTM! 👍

client/fixtures/productBuilder/baseProducts.ts
rBangay: I moved them into the past to line up with the mocked data for the potential holiday stop api response. I could have setup some new holiday stop dates specifically for top tier, or updated the dates for all products everywhere, but I thought this was the simplest thing to do.

PR #1371, opened 2024-08-12
DRAFT - do not merge

AndreaDiotallevi:
DRAFT


rBangay:
Hey, I'm happy to approve, I just wanted to check, does the response from MDAPI specifically the `deliveryAddress.country` property always come through as a set list of options? eg will it always be "United States" and not "US" or "USA"?

AndreaDiotallevi:
> Hey, I'm happy to approve, I just wanted to check, does the response from MDAPI specifically the `deliveryAddress.country` property always come through as a set list of options? eg will it always be "United States" and not "US" or "USA"?

Good question!

The delivery address comes from the Salesforce contact record, so perhaps @graham228221 or @JoeMitchellGuardian can advise on this?

JoeMitchellGuardian:
> > Hey, I'm happy to approve, I just wanted to check, does the response from MDAPI specifically the `deliveryAddress.country` property always come through as a set list of options? eg will it always be "United States" and not "US" or "USA"?
> 
> Good question!
> 
> The delivery address comes from the Salesforce contact record, so perhaps @graham228221 or @JoeMitchellGuardian can advise on this?

This appears to be using the "Mailing Country" standard field on the Contact in Salesforce. This dataset is controlled by Salesforce and uses the full country name, rather than any abbreviations.

In the example provided by @rBangay, it would show "United States" as shown below 
Screenshot 2024-08-15 at 10 47 20

graham228221:
> > > Hey, I'm happy to approve, I just wanted to check, does the response from MDAPI specifically the `deliveryAddress.country` property always come through as a set list of options? eg will it always be "United States" and not "US" or "USA"?
> > 
> > 
> > Good question!
> > The delivery address comes from the Salesforce contact record, so perhaps @graham228221 or @JoeMitchellGuardian can advise on this?
> 
> This appears to be using the "Mailing Country" standard field on the Contact in Salesforce. This dataset is controlled by Salesforce and uses the full country name, rather than any abbreviations.
> 
> In the example provided by @rBangay, it would show "United States" as shown below Screenshot 2024-08-15 at 10 47 20

Actually some "legacy" subscribers might have abbreviations as they were never corrected. In Salesforce these are the contacts with active subscriptions:

MailingCountry | Contacts
-- | --
USA | 1673
US | 19230
UNITED STATES | 26720

I think for this piece of work the Billing Address (billToContact?) is more relevant:

BillingCountry | Contacts
 -- | --
USA | 510
US | 12046
UNITED STATES | 182249

client/__tests__/components/payment/stripe.test.ts
```diff
@@ -37,3 +42,14 @@ test('Uses Australian Stripe key for Australian delivery address', () => {
 		window.guardian.stripeKeyAustralia?.default,
 	);
 });
+
+test('Uses United States Stripe key for United States delivery address', () => {
+	const stripePublicKey = getStripeKey(
+		guardianWeeklySubscriptionUnitedStates.deliveryAddress?.country,
```
johnduffell, line 48: why create the whole `guardianWeeklySubscriptionUnitedStates` object only to pick out just the delivery country?

client/fixtures/subscription.ts
```diff
@@ -115,6 +115,60 @@ export const guardianWeeklySubscriptionAustralia: Subscription = {
...
+export const guardianWeeklySubscriptionUnitedStates: Subscription = {
+	paymentMethod: 'Card',
+	card: {
+		last4: '4242',
+		expiry: {
+			month: 3,
+			year: 2033,
+		},
+		type: 'Visa',
+		stripePublicKeyForUpdate: 'pk_test_123',
```
johnduffell, line 127: I haven't traced the code fully, but it seems like the public key is coming from MDAPI in some? cases and it would be based off the billing country. https://github.com/guardian/members-data-api/blob/7094e7fd2927c97101fa11c7b29050aa55afd31b/membership-attribute-service/app/services/AccountDetailsFromZuora.scala#L52

If it's all tested with GW and other products and works ok then it's probably ok, I think GW is the most tricky with the billing vs delivery country plus currencies. (we lock the delivery country and currency together IIRC so that people definitely pay the right price)

AndreaDiotallevi:
> > > > Hey, I'm happy to approve, I just wanted to check, does the response from MDAPI specifically the `deliveryAddress.country` property always come through as a set list of options? eg will it always be "United States" and not "US" or "USA"?
> > > 
> > > 
> > > Good question!
> > > The delivery address comes from the Salesforce contact record, so perhaps @graham228221 or @JoeMitchellGuardian can advise on this?
> > 
> > 
> > This appears to be using the "Mailing Country" standard field on the Contact in Salesforce. This dataset is controlled by Salesforce and uses the full country name, rather than any abbreviations.
> > In the example provided by @rBangay, it would show "United States" as shown below Screenshot 2024-08-15 at 10 47 20
> 
> Actually some "legacy" subscribers might have abbreviations as they were never corrected. In Salesforce these are the contacts with active subscriptions:
> 
> MailingCountry	Contacts
> USA	1673
> US	19230
> UNITED STATES	26720
> I think for this piece of work the Billing Address (billToContact?) is more relevant:
> 
> BillingCountry	Contacts
> USA	510
> US	12046
> UNITED STATES	182249

Thanks @graham228221 and @JoeMitchellGuardian for commenting.

Does this mean that we should check for all strings "United States", "UNITED STATES", "US" and "USA" when we pick the correct Stripe account?

client/__tests__/components/payment/stripe.test.ts
AndreaDiotallevi: Thanks John - I have refactored here: https://github.com/guardian/manage-frontend/pull/1371/commits/28d83658f3cf49cd1e7ec472e3547d04d18f3649

AndreaDiotallevi:
@rBangay @graham228221 @JoeMitchellGuardian 

FYI I will close this PR since ,following a call with Rupert and John, we are planning to update MDAPI so that manage-frontend won't need to work out the Stripe key for itself and always get it from MDAPI.

See related conversation: https://chat.google.com/room/AAAA6HaLTYQ/HD_8XQ2HBK4/FS2h9zgg6_s?cls=10

PR #1373, opened 2024-08-15
stop cancelled supporter plus users from being able to change their c…

rBangay:
### What does this PR change?
Stop cancelled supporter plus users from being able to change their contrubution amount.

Copy fix to show cancellation date prefix copy on manage product page before the cancellation date.

### Images
Before            |  After
  |


Joemgru (approved):
LGTM.

product / design Q - is there any reason we don't have a prompt for renewal in this scenario?

PR #1375, opened 2024-09-02
Cancellation offer journey Tablet breakpoint

rBangay:
### What does this PR change?

Add specific view for the tablet breakpoint for the cancellation offer journey. The tablet view changes from a stacked view to a column one.

### Images

Before             |  After
  |

### DIFF
--- a/.storybook/viewport.ts
+++ b/.storybook/viewport.ts
@@ -49,6 +49,19 @@ const viewportEntries = Object.entries(breakpoints).map(([name, width]) => {
 	];
 });
 
+const extraViewPorts = {
+	largeTablet: {
+		name: 'Tablet upper bounds',
+		styles: {
+			width: `${breakpoints.desktop - 1}px`,
+			height: `${breakpoints.desktop + 399}px`,
+		},
+	},
+};
+
 export const viewport = {
-	viewports: Object.fromEntries(viewportEntries),
+	viewports: {
+		...Object.fromEntries(viewportEntries),
+		...extraViewPorts,
+	},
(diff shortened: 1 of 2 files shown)


.storybook/viewport.ts
```diff
@@ -49,6 +49,19 @@ const viewportEntries = Object.entries(breakpoints).map(([name, width]) => {
 	];
 });
 
+const extraViewPorts = {
+	largeTablet: {
```
tjmw, line 53: Just to make sure I understand, we're adding our own new intermediate breakpoint (`largeTablet`) which doesn't exist in source, between tablet and desktop? Why's that needed in storybook? Looking at the changes below, I think we change the styling to do the column layout from tablet up (rather than desktop up as before)?

tjmw (approved):
Looks good! (With a question, which is probably me just missing something!)

.storybook/viewport.ts
rBangay: Hey, yeah so I added in this extra view specifically for story book. IT is supposed to represent the upperbounds (width) of the tablet breakpoint so that you can view a story at the smallest and largest tablet breakpoint bounds.

In this case it is useful given the image asset on the right hand side is displayed in its full width only at the upper bounds of the tablet breakpoint. Basically it makes it easier to review the component .

rBangay: Screenshot 2024-09-04 at 09 46 13

This probably explains it a lot better ... it allows me to add this extra option when selecting the window size for a story

PR #1384, opened 2024-09-06
Use discount api response for cancellation offer payment date

rBangay:
### What does this PR change?

Use the discount api response to return the next payment price (after the cancellation offer period ends). This should handle the case where the returning payment price will be different from the price paid before the offer was taken up, a price rise for example 👍 

Screenshot 2024-09-06 at 14 19 43

### DIFF
--- a/client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
+++ b/client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
@@ -19,6 +19,7 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import { Ribbon } from '@/client/components/shared/Ribbon';
 import { measure } from '@/client/styles/typography';
 import type { DiscountPreviewResponse } from '@/client/utilities/discountPreview';
+import { getMaxNonDiscountedPrice } from '@/client/utilities/discountPreview';
 import { DATE_FNS_LONG_OUTPUT_FORMAT, parseDate } from '@/shared/dates';
 import { number2words } from '@/shared/numberUtils';
 import { getMainPlan, isPaidSubscriptionPlan } from '@/shared/productResponse';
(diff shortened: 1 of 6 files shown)


client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
```diff
@@ -175,6 +175,14 @@ export const SupporterPlusOffer = () => {
 		'yyyy-MM-dd',
 	).dateStr(DATE_FNS_LONG_OUTPUT_FORMAT);
 
+	const strikethroughPrice = routerState.nonDiscountedPayments.reduce(
+		(prev, current) =>
+			prev && prev.amount > current.amount ? prev : current,
+	).amount;
```
tjmw, line 181: Not that I'm not a fan of `reduce` (I am, and think I probably overuse it) but what about this as an alternative? Maybe it's slightly clearer what the intention is?

```suggestion
	const strikethroughPrice = Math.max(...routerState.nonDiscountedPayments.map(p => p.amount));
```

Or splitting it into two:

```js
    const allAmounts = routerState.nonDiscountedPayments.map(p => p.amount);
    const strikeThroughPrice = Math.max(...allAmounts);
```

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOfferConfirmed.tsx
```diff
@@ -181,6 +186,16 @@ export const SupporterPlusOfferConfirmed = () => {
 		'yyyy-MM-dd',
 	).dateStr(DATE_FNS_LONG_OUTPUT_FORMAT);
 
+	const nextNonDiscountedPrice = routerState.nonDiscountedPayments.reduce(
+		(prev, current) =>
+			prev && prev.amount > current.amount ? prev : current,
+	).amount;
```
tjmw, line 192: Would it be worth extracting this to a method, since it's the same logic repeated in `SupporterPlusOffer.tsx`?

tjmw (approved):
Looks good! I left some nitpicky comments to take or leave, see what you think (no obligation to make any changes, I think they're all subjective).

client/components/mma/cancel/cancellationSaves/supporterplus/SupporterPlusOffer.tsx
rBangay: yeah I like this 👍

PR #1386, opened 2024-09-13
Add China and Qatar to non-servicable countries

david-pepper:
As part of the 2024 VAT Compliance annual review, we have identified that China and Qatar should be excluded from the countries that we sell subscriptions in. 

### What does this PR change?
Updates the list of NonServiceableCountries in MMA so that product switching cannot take place in China or Qatar.

### Testing
We updated the Cypress test to use China and Qatar as the source country for a supporter plus switch, and verified that the test passed for each country.

### Next steps/further info
Once complete, the next steps will be to inform customers in these countries that their subscriptions will be cancelled, and we will then proceed to cancel those subs.


rBangay (approved):
👍

PR #1395, opened 2024-09-25
use new react root method

rBangay:
### What does this PR change?

Update the react root method from `render` to `createRoot` in order to make proper use of react 18 instead of running in react 17 legacy mode.

Required changes to the holiday stop cypress tests to explicitly fire the mousover events on the date cells (this appeared to happen for free while using react 17 legacy mode)

More info here: https://github.com/reactwg/react-18/discussions/5

Also associated formik versino bump


tjmw (approved):
Nice!

PR #1398, opened 2024-09-26
Remove unnecessary web socket dependency resolutions

rBangay:
### What does this PR change?
Remove unnecessary web socket dependency resolutions (it looks as though specific resolution versioning like this doesn't work and the versions of ws that we use are all vulnerability free anyway)

link to resolution documentation: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/

### DIFF
--- a/package.json
+++ b/package.json
@@ -190,12 +190,6 @@
     "yup": "0.28.5",
     "zod": "3.22.4"
   },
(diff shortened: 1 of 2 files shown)


tjmw (approved):
Great! 🎉

PR #1401, opened 2024-10-07
fix(accessibility): Improve accessibility of "Email & Marketing" page.

AshCorr:
## Current situation/background

Current the "Email & Marketing" page is a bit of a headache to navigate using a screen reader, there are 2 major issues in particular:

 - Dropdowns are not marked as interactive, they appear as regular text to screen readers, and they also cannot be toggled using the expected keys "Space" or "Enter"
 - Checkboxes cannot be toggled when navigated to with a screen reader, more context on this below.

## What does this PR change?

 - Removes hidden `input` in consent checkboxes. I believe these were here to try and make the checkbox more accessible, but they seem to have the opposite effect at the moment as screen readers will focus them preferentially instead of the "real" checkbox, and these hidden checkboxes aren't hooked up properly to accept input.
 - Correctly handle onKeyDown events in the Dropdown and Checkboxes, as per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role these elements should be interactable with Space and sometimes Enter.
 - Switch dropdown button to an `a` instead of `div` and attach more accessibility attributes. Previously this dropdown gave no hint to screen readers that it was interactable and was quite difficult to interact with.
 - Add labels to all consent checkboxes.

## Next steps/further info

| Before | After |
|--------|--------|
| image | image |
| **Voiceover Prompts** |
| Dropdown |
| image | image |
| image |image| 

### Further improvements that could be made

 - Newsletter dropdowns are a H2, but they should/could probably be a H3 as the "Your newsletters" heading is already a H2.
 - The clock SVG by the newsletter frequency should be hidden or given a label

### DIFF
--- a/client/components/mma/identity/MarketingCheckbox.tsx
+++ b/client/components/mma/identity/MarketingCheckbox.tsx
@@ -5,7 +5,7 @@ import { standardSansText } from './sharedStyles';
 interface MarketingCheckboxProps {
 	id: string;
 	description?: string;
-	title?: string;
+	title: string;
 	selected?: boolean;
 	onClick: (id: string) => {};
 }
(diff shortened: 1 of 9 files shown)


client/components/mma/identity/MarketingCheckbox.tsx
```diff
@@ -38,18 +38,11 @@ const getDescription = (description: MarketingCheckboxProps['description']) => (
...
 			data-cy={id}
-			onClick={(e) => {
-				// Checkboxes inside labels will trigger click events twice.
-				// Ignore the input click event
-				if (
-					e.target instanceof Element &&
-					e.target.nodeName === 'INPUT'
-				) {
-					return;
-				}
```
AshCorr, line 52: This used to be needed in order to catch the onClick event of the hidden checkbox. Not needed now that the hidden checkbox is gone.

PR #1402, opened 2024-10-14
voucher subs holiday stop bug fix

rBangay:
### What does this PR change?

Fixes a bug where voucher subscription users were unable to check a checkbox confirming that they will destroy their vouchers cards for the holiday stops that they are trying to book. This bug was preventing the users from completing the holiday stop journey.

The problem seems to be with the mma checkbox component, which was used in the holiday stop journey and also in the marketing and consents page but only seems to negatively affect the holiday stop journey. The marketing and consents page already had some cypress test coverage where as creating a holiday stop with a voucher subscription was not covered and so now is as part of this pr.

The problem seems to coincide with the following pr to properly integrate react v18 into MMA (https://github.com/guardian/manage-frontend/pull/1395), although more work could be done to investigate why this might be.

There is a checkbox component in the source library, and so I have replaced/removed the mma internal checkbox component and replaced it with the source one and created a new cypress test to cover voucher subscription holiday stops.


client/components/mma/holiday/HolidayReview.tsx
```diff
@@ -212,17 +217,12 @@ export const HolidayReview = () => {
...
-								justifyContent: 'flex-end',
-								marginTop: '20px',
-								[until.mobileMedium]: {
-									flexDirection: 'column',
-									marginTop: 0,
-								},
+						css={{
+							marginTop: '20px',
+							[until.mobileMedium]: {
+								marginTop: 0,
```
rBangay, line 223: just fixing a very small layout issue where the inline error component was displaying inline instead of block

client/components/mma/holiday/HolidayReview.tsx
```diff
@@ -253,7 +253,11 @@ export const HolidayReview = () => {
 						{!!productType.holidayStops
 							.explicitConfirmationRequired &&
 							!isCheckboxConfirmed && (
-								<InlineError>
+								<InlineError
+									cssOverrides={css`
+										margin-top: ${space[3]}px;
+									`}
+								>
```
rBangay, line 260: again just fixing a very small layout issue where the inline error component was displaying inline instead of block

client/components/mma/holiday/HolidayReview.tsx
```diff
@@ -138,15 +141,17 @@ export const HolidayReview = () => {
...
-										setIsCheckboxConfirmed(newValue)
-									}
 									label={
 										productType.holidayStops
 											.explicitConfirmationRequired
 											.checkboxLabel
 									}
+									checked={isCheckboxConfirmed}
+									onChange={() => {
+										setIsCheckboxConfirmed(
```
tjmw, line 151: Is this a case where we should be [passing a function](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state) to `setIsCheckboxConfirmed`, since we're updating the state based on the current state? In the old pre-hooks world I think I would have done this, but not sure if it's required here. e.g. `setIsCheckBodyConfirmed(previousIsCheckboxConfirmed => !!previousIsCheckboxConfirmed)`

client/components/mma/identity/emailAndMarketing/NewsletterSection.tsx
```diff
@@ -82,6 +82,7 @@ export const NewsletterSection: FC<NewsletterSectionProps> = (props) => {
 	const { newsletters, clickHandler } = props;
 	return (
 		<PageSection
+			testSelector="your_newsletters"
```
tjmw, line 85: Is the `title` not enough for Cypress to find this? (I'm guessing not otherwise you wouldn't have added it - but interested in what the problem was!)

tjmw (approved):
Looks good! Nice to be using the source component instead of a home grown version 👍

client/components/mma/holiday/HolidayReview.tsx
rBangay: hey, I think we're ok here as we are only calling the setState method once and not multiple times in a row/tick.  Good shout though

client/components/mma/identity/emailAndMarketing/NewsletterSection.tsx
rBangay: I've replaced the data-cy attribute with a text and parent selector 👍

PR #1403, opened 2024-10-21
Newspaper Archive endpoint - add ncom-return-url support

andrewHEguardian:

Currently we have an endpoint that redirects a user to newspapers dot com with a token.

### What does this PR change?

This adds a mechanism to return the user to the image they were viewing. Newspapers.com will construct a url to send them to our endpoint. Eg going to -> https://manage.thegulocal.com/newspaperArchive/auth/?ncom-return-url=https%3A%2F%2Ftheguardian.newspapers.com%2Fimage%2F258351828%2F

redirects the user to this screen

image

The popup modal shouldn't be there but we do have "Free access" in the top right hand corner and you can view the page once you dismiss the modal.

### Next steps/further info

1. Find out why the popup is showing or if we're not constructing this URL correctly 
2. Add a check for supporter status to this endpoint

### DIFF
--- a/server/routes/newspaperArchive.ts
+++ b/server/routes/newspaperArchive.ts
@@ -55,10 +55,11 @@ router.get('/auth', async (req: Request, res: Response) => {
 	);
 
 	const responseJson = (await response.json()) as NewspapersResponseBody;
-	const tpaToken = new URL(responseJson.url).searchParams.get('tpa');
 
 	const archiveReturnUrlString = req.query['ncom-return-url'];
 	if (archiveReturnUrlString && typeof archiveReturnUrlString === 'string') {
+		const tpaToken = new URL(responseJson.url).searchParams.get('tpa');
+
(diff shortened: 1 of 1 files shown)


server/routes/newspaperArchive.ts
```diff
@@ -55,7 +55,16 @@ router.get('/auth', async (_req: Request, res: Response) => {
 	);
 
 	const responseJson = (await response.json()) as NewspapersResponseBody;
-	res.redirect(responseJson.url);
+	const tpaToken = new URL(responseJson.url).searchParams.get('tpa');
```
tjmw, line 58: Should this be moved inside the if block below? It looks like we only use `tpaToken` if the `ncom-return-url` query string arg is specified.

andrewHEguardian: Good shout

andrewHEguardian:
The modal issue I raised in the PR description has since been fixed on the n com side

PR #1404, opened 2024-10-22
Newspaper Archive auth endpoint - check supporter entitlement

andrewHEguardian:

We have an endpoint that authenticates a user, and sends them off to newspapers dot com. It doesn't check if they're actually entitled to go there.

### What does this PR change?

This PR adds a check of the user's supporter entitlements - by sending a request to members data api's `user-attributes/me` endpoint.

### Next steps/further info

I've left 3 todos: 

1.  Refactor to parse the responses
2. Send the user somewhere if they aren't entitled (this is a nice to have)
3. Update the list of entitlements MDAPI returns, currently none accurately reflect this benefit of Tier3

### DIFF
--- a/server/apiProxy.ts
+++ b/server/apiProxy.ts
@@ -88,30 +88,6 @@ export const proxyApiHandler =
 			outgoingURL,
 		};
 
-		const authorizationOrCookieHeader = async ({
-			req,
-			host,
-		}: {
-			req: Request;
-			host: string;
-		}): Promise<Headers> => {
-			// If Okta is disabled, always return the cookie header
-			const { useOkta } = await getOktaConfig();
-			if (!useOkta) {
-				return {
-					Cookie: getCookiesOrEmptyString(req),
-				};
-			}
-			switch (host) {
-				case 'members-data-api.' + conf.DOMAIN:
-					return {
-						Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
-					};
-				default:
-					return {};
-			}
-		};
-
 		fetch(outgoingURL, {
 			method: httpMethod,
 			body: requestBody,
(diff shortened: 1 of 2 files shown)


server/apiProxy.ts
```diff
@@ -193,6 +169,32 @@ export const proxyApiHandler =
...
+		case 'members-data-api.' + conf.DOMAIN:
+			return {
+				Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
+			};
+		default:
+			return {};
+	}
+};
+
 export const customMembersDataApiHandler = proxyApiHandler(
```
rupertbates, line 198: Could we use this function rather than having to export the authorization one?

andrewHEguardian: I had a go at this but it's quite hard to wrangle. Because it's a callback that doesn't return anything, we can give it our express Response object but we don't actually want to hijack what we return to the user

server/routes/newspaperArchive.ts
```diff
@@ -69,3 +80,28 @@ router.get('/auth', async (req: Request, res: Response) => {
...
+}
+
+async function getSupporterStatus(req: Request) {
+	const host = 'members-data-api.' + conf.DOMAIN;
+
+	return fetch(`https://${host}/user-attributes/me`, {
+		method: 'GET',
+		headers: {
+			...(await authorizationOrCookieHeader({ req, host })),
+			'Content-Type': 'application/json',
```
tjmw, line 102: Super minor point - but do we need to send a `Content-Type` header here? My understanding is that it's needed as a request header when sending a request with a particular content type in the body (which isn't the case here as it's a GET).

server/routes/newspaperArchive.ts
```diff
@@ -36,7 +39,14 @@ router.get('/auth', async (req: Request, res: Response) => {
 	const authString = config?.authString;
 	if (authString === undefined) {
 		log.error(`Missing newspaper archive auth key`);
-		res.status(500).send();
+		return res.sendStatus(500);
+	}
+
+	const hasCorrectEntitlement = await checkSupporterEntitlement(req);
```
tjmw, line 45: I wonder if this should be wrapped in a try/catch in case any of the underlying operations error (for example if the request to mdapi fails)? I guess in that case we'd want to set `hasCorrectEntitlement` to false to be on the safe side.

tjmw (approved):
I think this looks good. I left a couple of bits of minor feedback to take or leave, see what you think.

server/routes/newspaperArchive.ts
andrewHEguardian: I think you're right! Was copied from the other code that calls mdapi but doesn't seem necessary here. I've also copied the `X_GU_ID_FORWARDED_SCOPE` which I'm assuming is important. I'll try test without it...

server/routes/newspaperArchive.ts
andrewHEguardian: I've wrapped the whole thing in a try catch, seems more accurate than eg telling a user they don't have a product in the case the mdapi call failed

PR #1405, opened 2024-10-23
Newspaper archive endpoint: add parsing

andrewHEguardian:

In #1404 we added a check to supporter entitlements to the newspaper archive auth endpoint

### What does this PR change?

This PR adds some zod parsing to make it more typesafe. I'm using `parse` rather than `safeParse`. Any parsing errors should get picked up by the try catch at the top level of the route


tjmw (approved):
Great! 👍

PR #1406, opened 2024-10-23
Newspaper archive auth - show disclaimer message to user missing entitlements

andrewHEguardian:

If a user hits the newspaper archive authentication link without the requisite product, they are redirected to the account overview.

### What does this PR change?

This PR adds an info box to show users who have a product but not the correct one.

### Next steps/further info

Copy TBC. This is a nice to have rather than a MVP feature.

### Screenshots

User has other products (ignore the payment failure that's not relevant for access)

image

User has no products (the existing text seems to cover this scenario)

image


paul-daniel-dempsey (approved):
Looks good

PR #1410, opened 2024-10-29
All access digital Annual cancellation offer

rBangay:
### What does this PR change?

Currently All access digital monthly subscribers can be shown an offer as part of the cancelltation journey. This pr adds the ability to offer annual All access digital subscribers a similar offer.

The annual offer is a percentage off offer as opposed to a free period with the monthly subscribers.

## Images
Screenshot 2024-10-29 at 16 37 33


GHaberis (approved):
Not familiar with the context, but code looks good to me

PR #1413, opened 2024-10-31
Newspaper archive auth - add query params override, validate return URL

andrewHEguardian:

We have an authentication endpoint to redirect users to newspapers dot com. This allows us to deliver a new benefit to subscribers of the Digital + Print product. The endpoint includes a feature that allows users to go back to the same page they were looking at, included as a link on a n com sign in modal. At the moment this doesn't work because the query param gets stripped out by our authentication middleware. 

Modal with sign in link:
image

Example URL:
https://manage.theguardian.com/newspaperArchive/auth?ncom-return-url=https%3A%2F%2Ftheguardian.newspapers.com%2Fimage%2F259880906%2F%3Fmatch%3D1%26terms%3Dking%2527s%2520place

### What does this PR change?

This PR adds an override to the URL sanitisation in order to preserve the `ncom-return-url` query param, and a check that the hostname is indeed newspapers dot com.

### DIFF
--- a/server/middleware/identityMiddleware.ts
+++ b/server/middleware/identityMiddleware.ts
@@ -65,8 +65,11 @@ const signedOutAfterTokensIssued = ({
 	// where the middleware may continually keep running performAuthorizationCodeFlow().
 	guSoTimestamp <= Math.floor(Date.now() / 1000);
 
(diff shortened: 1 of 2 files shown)


coldlink (approved):
looking okay from an guardian/identity perspective!

PR #1414, opened 2024-11-06
fire custom quantum metric event at the start of the cancellation flow

rBangay:
### What does this PR change?

Fire a custom Quantum metric event at the start of the cancellation journey.

This is so that we can identify the billing period of the product eg monthly or annual.

The event string sent will look something like this: 

`cancellation start | all-access digital subscription | billing period: month`

### DIFF
--- a/client/components/mma/cancel/CancellationJourneyFunnel.tsx
+++ b/client/components/mma/cancel/CancellationJourneyFunnel.tsx
@@ -3,8 +3,9 @@ import { Navigate, useLocation } from 'react-router-dom';
 import { featureSwitches } from '@/shared/featureSwitches';
 import {
 	getSpecificProductTypeFromTier,
-	type ProductDetail,
+	isPaidSubscriptionPlan,
 } from '@/shared/productResponse';
+import type { ProductTypeKeys } from '@/shared/productTypes';
 import { CancellationContext } from './CancellationContainer';
 import type {
 	CancellationContextInterface,
@@ -12,14 +13,10 @@ import type {
 } from './CancellationContainer';
 import { CancellationReasonSelection } from './CancellationReasonSelection';
 
-function productHasEarlySaveJourney(productToCancel: ProductDetail): boolean {
-	const specificProductTypeKey = getSpecificProductTypeFromTier(
-		productToCancel.tier,
-	).productType;
-
+function productHasEarlySaveJourney(productTypeKey: ProductTypeKeys): boolean {
 	return (
-		specificProductTypeKey === 'membership' ||
-		(featureSwitches.digisubSave && specificProductTypeKey === 'digipack')
+		productTypeKey === 'membership' ||
+		(featureSwitches.digisubSave && productTypeKey === 'digipack')
 	);
 }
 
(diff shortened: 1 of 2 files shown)


client/components/mma/cancel/CancellationJourneyFunnel.tsx
```diff
@@ -36,9 +33,20 @@ export const CancellationJourneyFunnel = () => {
...
+	const productType = getSpecificProductTypeFromTier(productDetail.tier);
+	const productTypeKey = productType.productType;
+
+	const possiblePaidPlan = productDetail.subscription.currentPlans[0];
+	const qmEventString = `cancellation start | ${productType.friendlyName}${
+		isPaidSubscriptionPlan(possiblePaidPlan)
+			? ` | billing period: ${possiblePaidPlan.billingPeriod}`
+			: ''
+	}`;
+	window.QuantumMetricAPI?.sendEvent(184, 0, qmEventString);
```
johnduffell, line 45: what is 184 and 0? (and should they be named consts?)

johnduffell, line 45: for clarity prefer to keep the structure until we need to use it?
might need to tweak my example to make it more typescripty though
```suggestion
	const qmEvent = [
	  `cancellation start`,
	  productType.friendlyName,
	];
	if (isPaidSubscriptionPlan(possiblePaidPlan)
			qmEvent.push(`billing period: ${possiblePaidPlan.billingPeriod}`);
	}
	window.QuantumMetricAPI?.sendEvent(184, 0, qmEventString.join(" | ");
```

shared/globals.ts
```diff
@@ -27,9 +27,13 @@ export interface Globals extends CommonGlobals {
 	stripeKeyAustralia?: StripePublicKeySet;
 	stripeKeyDefaultCurrencies?: StripePublicKeySet;
 }
+interface QuantumMetricAPIPartial {
+	sendEvent: (param1: number, param2: number, event: string) => void;
```
johnduffell, line 31: is it this? 
https://github.com/quantummetric/analytics-module-qm/blob/3e927c8c670b2c1f13ae543a9e1863e8afa641bd/src/api/QuantumMetric.ts#L40-L44

client/components/mma/cancel/CancellationJourneyFunnel.tsx
rBangay: yeah, I've renamed them based on the interface you found 👍

rBangay: yep, I've refactored this in to something similar

shared/globals.ts
rBangay: That looks like the one yep ... it looks like support-frontend do something similar: https://github.com/guardian/support-frontend/blob/a5f4a1b3c8a3c32d14bbeade9601087f1e43c2ea/support-frontend/window.d.ts#L77

I have refactored the interface to make it clear what the functions arguments are.

client/components/mma/cancel/CancellationJourneyFunnel.tsx
```diff
@@ -36,9 +33,27 @@ export const CancellationJourneyFunnel = () => {
...
+
+	const possiblePaidPlan = productDetail.subscription.currentPlans[0];
+	const qmEventId = 184;
+	const qmIsConversionEvent = 0;
+	const qmEvent = [
+		'cancellation start',
+		productType.friendlyName,
+		isPaidSubscriptionPlan(possiblePaidPlan) &&
+			`billing period: ${possiblePaidPlan.billingPeriod}`,
+	].filter(Boolean);
```
johnduffell, line 47: so filter(Boolean) will filter out anything that is not a boolean?

johnduffell (approved):
great work 👍

client/components/mma/cancel/CancellationJourneyFunnel.tsx
rBangay: anything that is not truthy 👍

PR #1416, opened 2024-11-11
Update manage-frontend to use node v22

rBangay:
### What does this PR change?
ronseal


riff-raff.yaml
```diff
@@ -22,7 +22,7 @@ deployments:
         CODE: ManageFrontend-CODE.template.json
         PROD: ManageFrontend-PROD.template.json
       amiTags:
-        Recipe: jammy-membership-node18
+        Recipe: jammy-membership-node22
```
johnduffell, line 25: will we need to stop the node 18 one baking now (to save money), or is that still used by SDC?

johnduffell (approved):
fabulous

I assume it's tested on CODE?

rBangay:
> fabulous
> 
> I assume it's tested on CODE?

yep, I deployed it to code and tested it

PR #1415, opened 2024-11-11
Update cancellation reasons for SupporterPlus and Recurring contributions

rBangay:
### What does this PR change?
Updates the cancellation reasons for supporter plus and recurring contributions. Consolidates the list of reasons and add's 2 new options.

See the raw data in "Cancellation journeys" doc https://docs.google.com/spreadsheets/d/1rf0M-TYmXltJkLHUy7hIeVtezQMmfg4TOaJfMXD3ltk/edit?gid=1054852399#gid=1054852399

### images

Contribution Before             |  Contribution After
:-------------------------:|:-------------------------:
  |  

Supporter plus Before             |  Supporter plus After
:-------------------------:|:-------------------------:
  |

PR #1418, opened 2024-11-20
fairer array sorting algorithm

rBangay:
### What does this PR change?

Replace the existing shuffle array sort function with an algorithm that fairly randomises the contents. The new approach is based on the fisher-yates algothrithm (https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) a nice example of which you can see here: [Fisher–Yates Shuffle](https://bost.ocks.org/mike/shuffle/)

@johnduffell rightly pointed out in a previous pr (https://github.com/guardian/manage-frontend/pull/1415#discussion_r1846461396) that the randomisation of the output array wasn't all that random; 

results from his testing: 

```
Object.entries([...Array(1000000).keys().map(() => [1,2,3,4,5,6,7,8,9,10].sort( () => .5 - Math.random() )[0])].sort().reduce((acc, num) => {const e = {...acc};e[num]=acc[num] + 1; return e;}, {1:0,2:0,3:0,4:0,5:0,6:0,7:0,8:0,9:0,10:0})).map(([num, count]) => `${num}: ${count/10000}%`).map((a) => console.log(a))
1: 19.5461%
2: 7.2563%
3: 9.6343%
4: 12.8512%
5: 8.3207%
6: 8.9094%
7: 9.9178%
8: 11.1279%
9: 6.0317%
10: 6.4046%
```

It shows a weighting more specifically to the first element in the array (ideally these figures ought to be as close to 10% each in order to show a fair distribution)

and with the new Fisher–Yates Shuffle algorithm we can see the results much more evenly distributed:

```
1: 9.9887%
2: 10.0008%
3: 9.9885%
4: 10.0155%
5: 10.0069%
6: 9.9565%
7: 9.9425%
8: 9.9936%
9: 10.0209%
10: 10.0861%
```

### DIFF
--- a/client/utilities/utils.ts
+++ b/client/utilities/utils.ts
@@ -6,8 +6,27 @@ export function flattenEquivalent<T>(x: T): T {
 	return x;
 }
 
-export const shuffleArray = (array: unknown[]) =>
-	[...array].sort(() => 0.5 - Math.random());
+/*
+ * implementation of the Fisher–Yates Shuffle algorithm
+ * example here: https://bost.ocks.org/mike/shuffle/
+ */
+export const shuffleArray = (array: unknown[]) => {
+	let currentIndex = array.length;
+
+	// While there remain elements to shuffle...
+	while (currentIndex != 0) {
+		// Pick a remaining element...
+		const randomIndex = Math.floor(Math.random() * currentIndex);
+		currentIndex--;
+
+		// And swap it with the current element.
+		[array[currentIndex], array[randomIndex]] = [
+			array[randomIndex],
+			array[currentIndex],
+		];
+	}
+	return array;
+};
 
 export function formatAmount(amount: number) {
 	return Number.isInteger(amount) ? amount : amount.toFixed(2);
--- a/client/__tests__/utils.test.ts
+++ b/client/__tests__/utils.test.ts
@@ -1,0 +1,13 @@
+import { shuffleArray } from '../utilities/utils';
+
+describe('shuffleArray', () => {
+	it('randomizes the positions of the elements in the passed array', () => {
+		const inputArrString = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].toString();
+		let a;
+		for (a = 0; a < 100; a++) {
+			expect(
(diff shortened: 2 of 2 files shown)


johnduffell:
nice one! good bit of algorithms.  why not add a unit test to it to round it off?

client/utilities/utils.ts
```diff
@@ -6,8 +6,27 @@ export function flattenEquivalent<T>(x: T): T {
...
-	[...array].sort(() => 0.5 - Math.random());
+/*
+ * implementation of the Fisher–Yates Shuffle algorithm
+ * example here: https://bost.ocks.org/mike/shuffle/
+ */
+export const shuffleArray = (array: unknown[]) => {
+	let currentIndex = array.length;
+
+	// While there remain elements to shuffle...
+	while (currentIndex != 0) {
```
johnduffell, line 17: as the last step, this will always swap the first element with itself.  Which is not in itself a problem as it's a no-op, but it was interesting.

johnduffell (approved):
nice work, big improvement 👏

rBangay:
> nice one! good bit of algorithms. why not add a unit test to it to round it off?

I have added a unit test, but it was tricky to know how to properly test this. I didn't want to add a test that ran too many cycles to find whether the spread was even or not so in the end I just wrote a test that calls the function 100 times and checks that the output array order is not the same as the input array order .... the input array size was 10, I guess in theory the test could fail if by chance the shuffle resulted in an array with the elements in the same order but I fugured that was incredibly unlikely enough to make this test valid.

johnduffell:
> > nice one! good bit of algorithms. why not add a unit test to it to round it off?
> 
> I have added a unit test, but it was tricky to know how to properly test this. I didn't want to add a test that ran too many cycles to find whether the spread was even or not so in the end I just wrote a test that calls the function 100 times and checks that the output array order is not the same as the input array order .... the input array size was 10, I guess in theory the test could fail if by chance the shuffle resulted in an array with the elements in the same order but I fugured that was incredibly unlikely enough to make this test valid.

probably not worth going around the houses too much but if we're running it 100 times we should expect each number should be in each position at least a few times

johnduffell (approved):
nice work 👏

PR #1419, opened 2024-11-22
Add cloudwatch metric for redirecting from ncom

andrewHEguardian:

We have an endpoint to redirect users from newspapers.com back again with a third party authentication token

### What does this PR change?
This PR adds a cloudwatch metric to track this happening

image

### Next steps/further info
Possibly add an alarm if this hasn't happened in a while?


rupertbates (approved):
👍 Nice

PR #1420, opened 2024-11-26
account overview offer period restriction

rBangay:
### What does this PR change?
Restrict the account overview 'Your offer is active' message to monthly customers (until we have a more detailed response from mdapi about the offer status of a product to allow us to distinguish between offer types). 

This should stop annual users from seeing a monthly offer message on the account overview.

A later pr will add the annual offer message in once more work has been done in mdapi to return the info needed to display the message.

### Images
Screenshot 2024-11-26 at 15 03 32


MaelGNM (approved):
Looks good to me!

PR #1425, opened 2024-12-04
Use user-benefits API to work out newspaper archive entitlement

rupertbates:
### What does this PR change?
Up until now we have been checking members-data-api to see whether a user is eligible for newspaper archive access. This was problematic because the only way we could deduce whether a user was entitled or not was if they had both the `supporterPlus` and `guardianWeeklySubscriber` flag set, however this is not actually the same as being subscribed to tier three.

This PR replaces the call to mdapi with a call to the new user-benefits API which has been built to centralise the logic around which users are entitled to which benefits.

[Depends on this PR](https://github.com/guardian/support-service-lambdas/pull/2572)

### DIFF
--- a/server/apiProxy.ts
+++ b/server/apiProxy.ts
@@ -188,6 +188,10 @@ export const authorizationOrCookieHeader = async ({
 			return {
 				Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
 			};
+		case 'user-benefits.' + conf.API_DOMAIN:
+			return {
+				Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
+			};
 		default:
 			return {};
 	}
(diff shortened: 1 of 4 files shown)


server/apiProxy.ts
```diff
@@ -188,6 +188,10 @@ export const authorizationOrCookieHeader = async ({
 			return {
 				Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
 			};
+		case 'user-benefits.' + conf.API_DOMAIN:
+			return {
+				Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
+			};
```
rBangay, line 194: these two cases look like they return the same thing, so you could probably combine them like this:

```
case 'members-data-api.' + conf.DOMAIN:
case 'user-benefits.' + conf.API_DOMAIN:
  return {
    Authorization: `Bearer ${req.signedCookies[OAuthAccessTokenCookieName]}`,
  };
```

rBangay (approved):
👍  ... approved with one pedantic comment

PR #1426, opened 2024-12-10
Cross spawn vuln fix

rBangay:
### What does this PR change?

Update dependencies (and remove uneeded ones) to ensure that we are consuming. a new or patched version of cross spawn.

Fixes a regexp DOS vulnerablility reported by dependabot and snyk:

https://github.com/guardian/manage-frontend/security/dependabot/164

### More detail on dependencies:

- copy-node-modules ... I cannot see why we would need use this seeing as we copy the package.json file over to the dist directory already and run yarn install. I wonder whether this was a quirk of when we used to be on teamcity (perhaps we weren't able to run the yarn install task on out team city instances or something like that) anyhoo, we're in github action land now. It looks like a brave soul tried to do the same all the way back in 2018 to no avail: https://github.com/guardian/manage-frontend/pull/63
- husky ... I've updated husky to a much more recent major version which now doesn't seem to use cross-spawn at all 👍 . It required changes to the way we setup and use our husky hooks though.
-  stylis-pxtorem ... I couldn't see we were using this anywhere anyway
- guardian/commercial ... updated a few major versions, but there doesn't seem to be any breaking changes for us. It bumped the version of cross-spawn it used from 5.1.0 - 7.0.3
- I bumped the minor version of cypress in the hope it would consume a fixed/patched version of cross-spawn but nope ... still its good to keep up to date 🤓 

After that then all the remaining dependencies that consume cross-spawn were now on version 7.0.3, so I added a line to the resolutions block in package.json to force a patched version of at least 7.0.5.


MaelGNM (approved):
It looks good to me, thanks for the explanations in the description!

PR #1430, opened 2024-12-16
path-to-regexp dependency vuln fix

rBangay:
### What does this PR change?
Bump relevant peer dependencies in order to bump the version of path-to-regexp that the project uses to a patched version.

Fixes the vulnerabilities reported by snyk and dependabot:
https://app.snyk.io/org/guardian-value/project/8acda083-6b55-431d-a2e0-a985b2349e78#issue-SNYK-JS-PATHTOREGEXP-8482416


MaelGNM (approved):
👍

package.json
```diff
@@ -145,7 +145,7 @@
     "@emotion/react": "11.11.1",
     "@guardian/ab-core": "2.0.0",
     "@guardian/ab-react": "2.0.1",
-    "@guardian/commercial": "23.7.4",
+    "@guardian/commercial": "^23.7.5",
```
johnduffell, line 148: what does the extra `^` thing do?

johnduffell: according to google it lets it automatically update the minor and patch version.  But isn't that what package.json is supposed to prevent?

rBangay: if you don't use the caret then yes. If you have confidence that minor version bumps won't break anything then you can add the caret and let npm/yarn install the latest in the line of minor versions

PR #1431, opened 2024-12-19
hide the support cta's from the help centre footer

rBangay:
### What does this PR change?
Removes the support CTA from the footer in the help centre for the sake of app stores.

### Images
Before             |  After
  |  

#### example of mobile breakpoint without CTA:
Screenshot 2024-12-20 at 11 48 26

### DIFF
--- a/client/components/shared/footer/Footer.tsx
+++ b/client/components/shared/footer/Footer.tsx
@@ -171,7 +171,7 @@ const fillEmailSignup = (_: SyntheticEvent<HTMLIFrameElement>) => {
 	return;
 };
 
-export const Footer = () => {
+export const Footer = ({ hideSupport }: { hideSupport?: boolean }) => {
 	const TODAY = new Date(Date.now());
 
 	const [isInUSA, setIsInUSA] = useState<boolean>(false);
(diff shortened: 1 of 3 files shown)


client/components/shared/footer/Footer.tsx
```diff
@@ -255,19 +255,21 @@ export const Footer = () => {
 									</ul>
 								))}
 
-								<div css={supportStyles}>
-									<div css={supportTitleStyles}>
-										Support the&nbsp;Guardian
+								{!hideSupport && (
```
alinaboghiu, line 258: minor but could use a `showSupport` var instead and avoid a couple of negations

rBangay: Hey, I could do yeah, although the variable would require negation in its declaration. 

I went for `hideSupport` as the prop name so that you can use the footer throughout MMA without having to pass through an additional prop, so hiding the CTA's is the exception. For context this footer component is shared between the help centre and manage my account.

michaelbjacobson (approved):
Looks good to me!

client/components/shared/footer/Footer.tsx
alinaboghiu: > use the footer throughout MMA without having to pass through an additional prop
fair 👍

PR #1432, opened 2024-12-19
Remove Canada post strike message from account overview

rBangay:
### What does this PR change?
Removes the temporary Canada post strike message from account overview for users with a Teir three or Guardian weekly sub and a Canadian delivery address

### Images
Before   |  After
  |


MaelGNM (approved):
👍

PR #1435, opened 2024-12-24
Move cancel btn, US only

rBangay:
### What does this PR change?

Moves the cancellation cta from a link in the manage product page to the specific product card in the account overview. It is now a button instead of a link.

NOTE: This is currently only for products with a billing address in the US.

### Images
Account overview before             |  Account overview after
  |


ecguardian (approved):
LGTM 👍

PR #1436, opened 2024-12-27
CDK cross-span vuln fix (WIP)

rBangay:
### What does this PR change?

Fix security vulnerability in `cross-spawn` in CDK land.

As part of this pr I have also updated the version of [`CDK`](https://github.com/guardian/cdk) we use by quite a few major versions.

### DIFF
--- a/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
+++ b/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
@@ -21,6 +21,7 @@ Object {
       "GuEc2App",
       "GuCertificate",
       "GuInstanceRole",
+      "GuSsmSshPolicy",
       "GuDescribeEC2Policy",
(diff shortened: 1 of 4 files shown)


cdk/lib/__snapshots__/manage-frontend.test.ts.snap
```diff
@@ -2359,6 +2399,9 @@ Object {
             "Ref": "AMIManagefrontend",
           },
           "InstanceType": "t4g.small",
+          "MetadataOptions": Object {
+            "InstanceMetadataTags": "enabled",
+          },
```
johnduffell, line 2404: this is to allow access to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/work-with-tags-in-IMDS.html
This is probably useful for kibana logs or wazuh or something but shouldn't break anything.

johnduffell (approved):

PR #1438, opened 2025-01-06
supporterplus and tier three feast app benefit

rBangay:
### What does this PR change?

Add's the feast app to the list of benefits for All access digital (supporter plus) and Tier three customers.

The benefits lists show up in the product card in the account overview and manage product pages as well as the product switch journey from contribution to all access digital.

### Images
Before             |  After
  |  

Before             |  After
  |


johnduffell (approved):
nice looks good!

PR #1441, opened 2025-01-16
ESLint update

rBangay:
### What does this PR change?

Update eslint from version 8 to 9.

Update to use [guardian eslint rules](https://www.npmjs.com/package/@guardian/eslint-config?activeTab=readme)

Still to do for a later pr(s):
- I have added two eslint ignore line rules for the `react-hooks/exhaustive-deps` rule in order to stop re-render loops. A more detailed look into the root cause would be a good idea.
- the eslint.config.mjs file still maintains a list of globally ignored rules that we need to chip away at


client/components/helpCentre/contactUs/contactUsForm.tsx
```diff
@@ -180,6 +152,34 @@ export const ContactUsForm = (props: ContactUsFormProps) => {
...
+			window.v2ReCaptchaOnLoadCallback = renderReCaptcha;
+			document.head.appendChild(script);
+		}
+	}, []);
+
+	useEffect(() => {
+		if (!formValidationState.fileAttachment.isValid) {
+			validateForm();
+		}
+	});
```
MaelGNM, line 174: Why have you removed the dependencies here?

rBangay: Hey, that's because adding the `validateForm` function to the dependency array causes a re-render loop. One solution would be to wrap the `validateForm` function in a useCallback hook ... or like here I have removed the dependency array entirely and let the `validateForm` function be called on every page update, seeing as it's not an expensive function to run and the page doesn't re-render that often this seemed like the easiest solution

MaelGNM: That makes sense, thanks!

MaelGNM (approved):
Great work!

PR #1443, opened 2025-01-28
Upgrade to v4 of upload-artifact task in cypress workflow

MaelGNM:
The cypress e2e GitHub action workflow uses the v3 of the `upload-artifact` task which is being [deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).

This upgrades to v4.


MaelGNM:
The cypress E2E workflow failed, but that is unrelated to the changes. Also the changed step only happens on a fail and worked as expected here.

rBangay (approved):
👍

PR #1447, opened 2025-01-30
Fix Sign In Tests

coldlink:

In Gateway we've now changed the way that sign in works, making passcodes the default option in https://github.com/guardian/gateway/pull/3035. So the sign in screen/flow has changed.

Instead to sign in with a password we have to navigate to a separate page to do so, which is what this PR does.


rBangay (approved):
👍

PR #1450, opened 2025-02-05
allow test users to work with product-switch-api

johnduffell:
I noticed that I couldn't switch product as a test user in PROD, this was because it was trying to hit the PROD endpoint.  It looked like the extra header to flag whether it's a test user subscription wasn't being sent to manage-frontend backend.

This PR adds the extra header for product switches.

There may be other ones that are missing (TODO add logging in another PR)


rBangay (approved):
👍

PR #1451, opened 2025-02-05
log if the test user header isn't set on the server side

johnduffell:
We noticed the product-switch-api didn't work with test users because the header wasn't being set by the client side.

This change makes the server side log when it notices the missing header, so that we can fix them more easily.

Hopefully the messages will go to sentry so we have a chance of spotting them?  Otherwise we can look in cloudwatch.

### DIFF
--- a/server/apiGatewayDiscovery.ts
+++ b/server/apiGatewayDiscovery.ts
@@ -150,7 +150,10 @@ const getApiGateway = (
 			) =>
 			async (req: express.Request, res: express.Response) => {
 				const testUserHeader = req.header(MDA_TEST_USER_HEADER);
-				if (!['true', 'false'].includes(testUserHeader ?? '')) {
+				if (
+					testUserHeader === undefined ||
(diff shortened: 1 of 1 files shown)


rBangay (approved):
👍

server/apiGatewayDiscovery.ts
```diff
@@ -149,7 +149,13 @@ const getApiGateway = (
 				shouldNotLogBody?: boolean,
 			) =>
 			async (req: express.Request, res: express.Response) => {
-				const isTestUser = req.header(MDA_TEST_USER_HEADER) === 'true';
+				const testUserHeader = req.header(MDA_TEST_USER_HEADER);
+				if (!['true', 'false'].includes(testUserHeader ?? '')) {
```
rBangay, line 153: could we change this to something like the following ... I wonder if it's a bit more readable:

```
if (testUserHeader === undefined || !['true', 'false'].includes(testUserHeader)) {
```

PR #1456, opened 2025-02-07
Fix Cypress tests

rBangay:
### What does this PR change?

Expicitly returns after an MPAPI endpoint sends a 500 response (if it doesn't have access to idenity details). This stops the subsequent lines of code from running which are dependant on those idientity details being present.

This affected cypress integration tests that were asserting a redirect at the end of the test back to the account overview but not mocking the subsequent MDAPI and MPAPI api calls that would then be made.


MaelGNM (approved):
Good spot!

MaelGNM:
Closing this pull request as Snyk workflow is deprecated

PR #1458, opened 2025-02-07
chromatic random content render fix

rBangay:
### What does this PR change?
The cancellation reasons are rendered in a random order on the client to avoid people from clicking on the first option and throwing off any analysis we might be able to gather from the results. This however resulted in a visual regression change in chromatic every time a pull request was made to manage regardless of whether you had updated the cancellation reasons. 

This pr fixes the ordering of the cancellation reasons in storybook/chromatic so that they don't appear to change order.


MaelGNM (approved):
Looks good to me!

PR #1459, opened 2025-02-10
fix(cypress): fix sign in okta test

coldlink:
### What does this PR change?

Identity have changed the way that password sign in works in https://github.com/guardian/gateway/pull/3058, so update the  test behaviour to reflect that.


rBangay (approved):
👍

PR #1460, opened 2025-02-13
fixes a conditional logic bug that resulted in newspaper delivery and…

rBangay:
### What does this PR change?
Fixes a conditional logic bug that resulted in newspaper delivery and guardian patron subscribers seeing the wrong cancellation screen.

There was a conditional check that was looking for the presence of a property on a possibly non existant property and using that check to decide which branching path/comnponent to render at the start of the cancellation journey. Some of our products do not have a cancellation property on the productDetail object but didn't want to get thrown into the same bucket as other products that did have the cancellation property but not a reasons property on that object 😮‍💨 

I've added in a few more cypress integration tests that should help us not get into this situation again.

### NO THANK YOU 👎 
Screenshot 2025-02-13 at 17 08 49

### YES PLEASE 👍 
Screenshot 2025-02-13 at 17 10 46

Screenshot 2025-02-13 at 17 11 29


client/components/mma/cancel/CancellationJourneyFunnel.tsx
```diff
@@ -58,7 +58,10 @@ export const CancellationJourneyFunnel = () => {
 		return <Navigate to="./landing" state={{ ...routerState }} />;
 	}
 
-	if (!productType.cancellation?.reasons) {
+	if (
+		Object.hasOwn(productType, 'cancellation') &&
```
rBangay, line 62: This is the line of code that fixes the main bug

client/components/mma/cancel/CancellationSummary.tsx
```diff
@@ -157,55 +157,57 @@ const actuallyCancelled = (
...
-										productType.urlPart +
-										'_cancellation_summary'
-									}
-									theme="brand"
-									size="small"
-								/>
-							</div>
-						</>
-						)
+						{(!productType.cancellation ||
```
rBangay, line 160: There was a rendering bug here where we had some missing braces

johnduffell (approved):
I don't quite understand why the fix fixes it, but if it does, it does! 👍

client/components/mma/cancel/CancellationJourneyFunnel.tsx
rBangay: the property is on the object, not it's prototype.

Previously we were checking against 2 scenarios:
- with cancellation reasons
- without cancellation reasons

What we need to do is check for 3 scenarios:
- with cancellation reasons
- without cancellation reasons
- without a cancellation object at all

PR #1468, opened 2025-02-17
conditional logic refactor (cencellation journey)

rBangay:
### What does this PR change?
I've moved the conditional logic to check to see if we should render the 'contact us to cancel' component out of the 'cancellation reason selection' component as we don't render the reasons to these users/products anyway.

It's been moved over to the 'CancellationJourneyFunnel' component which works as a conduit for various products cancellation journey and which path they should take.

Hopefully this change makes it more obvious that there are products that we need to show the 'contact us to cancel' component to and that this has nothing to do with the reason selection component.


ecguardian (approved):
LGTM

PR #1470, opened 2025-02-19
Show singular cancelled subscription

rBangay:
### Context
You have only 1 product and it is cancelled and therefore returned in the `/cancelled` api response on the account overview.

Screenshot 2025-02-19 at 10 20 07

### What does this PR change?
If you have only 1 product and it it cancelled then it should still show up in the account overview if it is returned in the '/cancelled' api response. Before if you didn't also have an active pruduct then it would render the empty account overview component

### Images
Before             |  After
  |


MaelGNM (approved):
👍

PR #1472, opened 2025-03-04
subscribe/support again cta

rBangay:
### What does this PR change?

Display a subscribe/support again CTA in the cancelled product card on the account overview if the customer has no other active products.

### Images

with other active products:
Screenshot 2025-03-04 at 17 58 22

with no other active products:
Screenshot 2025-03-04 at 18 01 47

### DIFF
--- a/client/components/mma/accountoverview/CancelledProductCard.tsx
+++ b/client/components/mma/accountoverview/CancelledProductCard.tsx
@@ -18,8 +18,10 @@ import {
 
 export const CancelledProductCard = ({
 	productDetail,
+	hasOtherActiveSubs,
 }: {
 	productDetail: CancelledProductDetail;
+	hasOtherActiveSubs: boolean;
 }) => {
 	const specificProductType = getSpecificProductTypeFromTier(
 		productDetail.tier,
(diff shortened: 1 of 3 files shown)


andresilva-guardian (approved):
Nice work @rBangay 💪 
For what I could understand:
* The PR correctly updates the tracking events to differentiate between "support_again" and "subscribe_again"
* The checkoutUrlPart addition to product types is consistent across all relevant products
* The filtering logic for active products not pending cancellation is clean

My suggestions for improvement:
* Consider better naming for activeProductsNotPendingCancellation, maybe something like "fullyActiveProducts" instead
* Make the checkout URL construction more robust by creating a function to build it and hold the base url inside it
* I'm not aware yet of our unit test coverage, but an unit test for this would be nice.

Thank you for your work 👏

client/components/mma/accountoverview/CancelledProductCard.tsx
```diff
@@ -100,7 +99,7 @@ export const CancelledProductCard = ({
 						<div css={wideButtonLayoutCss}>
 							{showSubscribeAgainButton && (
 								<LinkButton
-									href="https://support.theguardian.com/uk/subscribe"
+									href={`https://support.theguardian.com/${specificProductType.checkoutUrlPart}`}
```
johnduffell, line 102: this ends up with a double slash as we have the slash both here and in the checkoutUrlPart
https://support.theguardian.com//support
Then that causes a 404

PR #1474, opened 2025-03-10
Make ELB to use a restricted cypher list in order to pass Nessus

paulbrown1982:
## What and why are you doing in this PR?
This PR changes the manage-frontend ELB to use a restricted cypher list in order to pass our Tenable/nessus PCI testing. 

Currently it is complaining with this issue: https://www.tenable.com/plugins/nessus/159543

This [AWS doc](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html) explains the options available, and the only list that is a subset of the Tenable list is: `ELBSecurityPolicy-TLS13-1-2-Res-2021-06`
 

[**Trello Card**](https://trello.com)
[none]

## How to test

- [ ] 1. Run Tenable against manage-frontend CODE ELB, see failure
- [ ] 2. Deploy PR to CODE (not late in Friday!), check site works OK in browser via ELB and Fastly.
- [ ] 3. Run Tenable against support-frontend CODE ELB, see pass
- [ ] 4. Merge to Prod
- [ ] 5. Run Tenable against support-frontend PROD ELB, see pass

Tenable scan passes.

Fastly might not work with reduced cypher suite.

### DIFF
--- a/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
+++ b/cdk/lib/__snapshots__/manage-frontend.test.ts.snap
@@ -1913,7 +1913,7 @@ Object {
         },
         "Port": 443,
         "Protocol": "HTTPS",
-        "SslPolicy": "ELBSecurityPolicy-TLS13-1-2-2021-06",
+        "SslPolicy": "ELBSecurityPolicy-TLS13-1-2-Res-2021-06",
       },
       "Type": "AWS::ElasticLoadBalancingV2::Listener",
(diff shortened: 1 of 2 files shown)


michaelbjacobson (approved):
All looks good to me! 

I've done a little digging into the likelihood of problems with Fastly, and I'm fairly confident we shouldn't encounter any issues. But more checking on that is probably worthwhile.

PR #1477, opened 2025-03-13
add Richard to google site verification

rBangay:
### What does this PR change?
Add Richard to the google site verification list and refactor the express routes to take an array of names and files instead of declaring a whole new route in code each time someone is added.

### DIFF
--- a/server/routes/core.ts
+++ b/server/routes/core.ts
@@ -88,18 +88,27 @@ router.get('/sitemap.txt', async (_, res: Response) => {
  * The current owners are listed below.
  * For each owner, please add github name and verification file.
  */
+const verificationFiles: Array<{ owner: string; filename: string }> = [
+	{
+		owner: 'kelvin-chappell',
+		filename: 'google6e3510e8603d6b4c.html',
+	},
+	{
+		owner: 'john-duffell',
+		filename: 'google499559a5c868dea3.html',
+	},
+	{
+		owner: 'richard-bangay',
+		filename: 'google7e53090f90b876c1.html',
+	},
+];
 
-// kelvin-chappell
-router.get('/google6e3510e8603d6b4c.html', (_, res: Response) => {
-	res.contentType('text/html').send(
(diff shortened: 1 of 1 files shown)


MaelGNM (approved):

server/routes/core.ts
```diff
@@ -88,18 +88,27 @@ router.get('/sitemap.txt', async (_, res: Response) => {
  * The current owners are listed below.
  * For each owner, please add github name and verification file.
  */
+const verificationFiles: Array<{ owner: string; filename: string }> = [
+	{
+		owner: 'kelvin-chappell',
+		filename: 'google6e3510e8603d6b4c.html',
```
johnduffell, line 94: to be fair we should probably remove kelvin now!

server/routes/core.ts
```diff
@@ -88,18 +88,27 @@ router.get('/sitemap.txt', async (_, res: Response) => {
...
+	},
+	{
+		owner: 'john-duffell',
+		filename: 'google499559a5c868dea3.html',
```
johnduffell, line 98: we don't need the `.html` bit although it doesn't hurt.

server/routes/core.ts
```diff
@@ -88,18 +88,27 @@ router.get('/sitemap.txt', async (_, res: Response) => {
  * The current owners are listed below.
  * For each owner, please add github name and verification file.
  */
+const verificationFiles: Array<{ owner: string; filename: string }> = [
+	{
+		owner: 'kelvin-chappell',
+		filename: 'google6e3510e8603d6b4c.html',
+	},
+	{
+		owner: 'john-duffell',
```
johnduffell, line 97: what's the reason to put the owner as a string instead of a comment?  it might be a lot more concise to just have a string array with a comment after each line

rBangay: hmm, I'm easy either way, I guess this way it forces you to add your name in there or typescript will moan at you

PR #1480, opened 2025-03-20
Add user-benefits information to the diagnostic information page

rupertbates:
The diagnostics information page at `/help-centre/diagnostic-information` allows us to retrieve debug information for users who are experiencing issues. Up until now this page has not had any information from the user-benefits API.

### What does this PR change?
This PR adds information from the user-benefits API
### Screenshot
Screenshot 2025-03-20 at 16 12 51


johnduffell (approved):
nice I've always used the /cookies route on support-frontend + MDAPI direct, but this one is more comprehensive and has most of that plus more.

PR #1484, opened 2025-03-25
Add Banner to Help Centre due to Telephony issues

JoeMitchellGuardian:
### Current situation/background
There is currently a telephony issue in our contact centre. This PR adds a banner to direct our customers to email us or use live chat to contact us with their queries to offset this issue.

### What does this PR change?
Add a new Banner to help centre - image to follow.

PR #1485, opened 2025-03-25
Remove UK Phoneline banner

JoeMitchellGuardian:
The UK Phone line issue with our contact centre is now resolved, so we can remove the banner added this morning.
### What does this PR change?
Reverts #1484, thus removing the banner.


ecguardian (approved):
👍

PR #1487, opened 2025-04-02
Add the holiday stop notice period for newspaper print delivery products

graham228221:

When a customer adds holiday stops, we only show the notice period for Paper Voucher and Subscription Card products:

Screenshot 2025-04-02 at 12 17 45

This is not shown for the other delivery products 😞 
image

### What does this PR change?
Adds in the notice period for the Home Delivery and National Delivery products.

Screenshot 2025-04-02 at 13 36 16

### Next steps/further info
Check if there is a good way to word the Guardian Weekly product which I think is something like "earliest holiday stop date is a week on Saturday"? It's defined [here](https://github.com/guardian/support-service-lambdas/blob/main/handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/GuardianWeekly.scala#L41).

### DIFF
--- a/shared/productTypes.ts
+++ b/shared/productTypes.ts
@@ -73,7 +73,7 @@ type SfCaseProduct =
 	| 'Guardian Weekly'
 	| 'Digital Pack Subscriptions'
 	| 'Supporter Plus'
-	| 'Tier '
+	| 'Tier Three'
 	| 'Guardian Ad-Lite'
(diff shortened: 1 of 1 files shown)


JoeMitchellGuardian (approved):

ecguardian (approved):
👍

PR #1493, opened 2025-04-14
Fix Payment Method Update Journey with Multiple Subscriptions of the Same Product

andresilva-guardian:
Currently, our payment method update journey has an issue when users have multiple subscriptions of the same product type. When a user with multiple subscriptions attempts to update their payment method:

1. The system cannot determine which specific subscription to update if the user navigates directly to the payment update URL (e.g., https://manage.thegulocal.com/payment/homedelivery/)
2. If there are multiple subscriptions of the same product type, the system redirects to the overview page
3. This creates problems when users are redirected back from external payment platforms (like Stripe Checkout), as the context of which subscription is being updated is lost

This issue was discovered during testing of the Stripe Checkout flow for updating payment methods for Observer subscriptions (Tortoise Media).

### What does this PR change?
This PR addresses the issue by:

1. Adding support for a subscriptionId query parameter in the payment update flow
2. Passing the subscription ID to the Stripe Checkout session creation endpoint
3. Including the subscription ID in the success URL when returning from Stripe
4. Modifying the PaymentDetailUpdateContainer to filter product details based on the subscription ID from query parameters
5. Ensuring the product details are preserved in the navigation state throughout the payment update journey

These changes allow the system to identify which specific subscription is being updated, even when users are redirected back from external payment platforms.

### Next steps/further info

1. We should consider adding similar subscription ID handling for other payment method update flows
2. We should monitor the implementation to ensure it works correctly in production for all use cases, especially when redirecting from external payment platforms

### Trello Card
[Link](https://trello.com/c/NPQxc7DJ/767-render-correct-subscription-of-multiple-products)

### DIFF
--- a/client/components/mma/paymentUpdate/PaymentDetailUpdateContainer.tsx
+++ b/client/components/mma/paymentUpdate/PaymentDetailUpdateContainer.tsx
@@ -115,7 +115,7 @@ export const PaymentDetailUpdateContainer = (
 					)}
 					render={renderContextAndOutletContainer(
 						isFromApp,
-						queryParameters.get('subscriptionId') ?? undefined,
+						queryParameters.get('subscriptionId') ?? undefined, // for internal links we use the context instead of query params
 					)}
 					loadingMessage={`Retrieving current payment details for your ${props.productType.friendlyName}...`}
 				/>
--- a/client/components/mma/paymentUpdate/PaymentDetailUpdateContainer.tsx
+++ b/client/components/mma/paymentUpdate/PaymentDetailUpdateContainer.tsx
@@ -1,6 +1,11 @@
 import type { Context } from 'react';
 import { createContext } from 'react';
-import { Navigate, Outlet, useLocation } from 'react-router-dom';
+import {
+	Navigate,
+	Outlet,
+	useLocation,
+	useSearchParams,
+} from 'react-router-dom';
 import type {
 	MembersDataApiResponse,
 	ProductDetail,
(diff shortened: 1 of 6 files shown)


client/components/mma/paymentUpdate/PaymentDetailUpdateContainer.tsx
```diff
@@ -86,7 +113,10 @@ export const PaymentDetailUpdateContainer = (
 					fetch={createProductDetailFetcher(
 						props.productType.allProductsProductTypeFilterString,
 					)}
-					render={renderContextAndOutletContainer(isFromApp)}
+					render={renderContextAndOutletContainer(
+						isFromApp,
+						queryParameters.get('subscriptionId') ?? undefined,
```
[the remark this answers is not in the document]
andresilva-guardian: No @johnduffell, this mechanism is only to be used in case of redirections from other websites back to MMA. Otherwise, we expect that MMA is able to find the proper context by itself within navigation. Since Stripe Checkout needs a redirection, it will use this mechanism.

client/components/mma/paymentUpdate/card/StripeCheckoutSessionButton.tsx
```diff
@@ -57,6 +58,7 @@ export const StripeCheckoutSessionButton = (
 			body: JSON.stringify({
 				paymentMethodType: props.paymentMethodType,
 				productTypeUrlPart: props.productTypeUrlPart,
+				subscriptionId: props.subscriptionId,
```
johnduffell, line 61: so this JSON has to match the type of `clientRequestBody` in server/stripeCreateCheckoutSessionHandler.ts I think?

Would it be possible to make them both have a common type, so that if they get out of step it would stop compiling?  Also it would make it easier to navigate the code.

johnduffell (approved):
I don't claim to understand this well but it makes sense to me!  Maybe @rBangay will also weigh in since he is the manage-frontend top committer.

client/components/mma/paymentUpdate/card/StripeCheckoutSessionButton.tsx
andresilva-guardian: I made the change @johnduffell thanks for the suggestion 💪

PR #1494, opened 2025-04-15
observer product name change amends

rBangay:
### What does this PR change?

- Variable and function name changes to distinguish between Newspaper Delivery and Ditgital voucher Sunday products and Observer specific products.
- Added a new story so that you can see only Observer products on the account overview
- Refactored the logic for the personalised header on the account overview so that the heading standfirst now does not render if you have soley Observer products, regardless of how many you may have. Previously the standfirst would not be rendered if you had a singular product that was an Observer product.


cypress/tests/mocked/parallel-2/cancelObserverDelivery.cy.ts
```diff
@@ -2,22 +2,22 @@ import { toMembersDataApiResponse } from '../../../../client/fixtures/mdapiRespo
 import { observerDelivery } from '../../../../client/fixtures/productBuilder/testProducts';
 import { signInAndAcceptCookies } from '../../../lib/signInAndAcceptCookies';
 
-describe('Cancel home delivery - Sunday', () => {
+describe('Cancel Observer delivery', () => {
```
johnduffell, line 5: it's still  "home" delivery even if it's observer (as opposed to national delivery)

client/utilities/sundayTheObserverSubscription.ts
```diff
@@ -8,3 +8,12 @@ export const isSundayTheObserverSubscription = (
...
 };
+
+export const isNotSundayTheObserverSubscription = (
+	productDetail: ProductDetail,
+): boolean => {
+	return (
+		productDetail.tier !== 'Newspaper Delivery - Observer' &&
+		productDetail.tier !== 'Newspaper Digital Voucher - Observer'
+	);
+};
```
johnduffell, line 19: ~please can we not have negative logic because I can't understand it~ 
please can we have positive logic because I can understand it better!
```suggestion
export const isSundayTheObserverSubscription = (
	productDetail: ProductDetail,
): boolean => {
	return (
		productDetail.tier === 'Newspaper Delivery - Observer' ||
		productDetail.tier === 'Newspaper Digital Voucher - Observer'
	);
};
```

client/components/mma/accountoverview/PersonalisedHeader.tsx
```diff
@@ -55,8 +55,7 @@ export const PersonalisedHeader = ({
 
 	const isSoleyObserverProductOwner =
```
johnduffell, line 56: should have been "solely" but I would have said "onlyHasObserverProducts"

cypress/tests/mocked/parallel-2/cancelObserverDelivery.cy.ts
rBangay: but the Observer is only inside M25 peeps right .. do we need to make that distinction?

client/utilities/sundayTheObserverSubscription.ts
rBangay: I moved these functions out into productResponse and renamed the non Observer one to `isGuardianProduct`

cypress/tests/mocked/parallel-2/cancelObserverDelivery.cy.ts
johnduffell: given the level of context we have, then yes, "observer only delivery" is not available outside M25 at present, but for the test I think it makes sense to be clear.
Although having said that "home delivery" is quite a legacy name

shared/productResponse.ts
```diff
@@ -119,6 +119,20 @@ export function isProduct(
 	return productTiers.includes((data as ProductDetail)?.tier);
 }
 
+export const isGuardianProduct = (productDetail: ProductDetail): boolean => {
+	return (
+		productDetail.tier !== 'Newspaper Delivery - Observer' &&
+		productDetail.tier !== 'Newspaper Digital Voucher - Observer'
+	);
+};
```
johnduffell, line 127: can we just reuse rather than reimplementing?
```suggestion
export const isGuardianProduct = (productDetail: ProductDetail): boolean => !isObserverProduct(productDetail);
```

client/components/mma/accountoverview/PersonalisedHeader.tsx
```diff
@@ -53,9 +53,9 @@ export const PersonalisedHeader = ({
 
 	const supportStartYear = dateString(oldestDate, 'MMMM yyyy');
 
-	const isSoleyObserverProductOwner =
+	const onlyHasObserverProducts =
 		mpapiResponse.subscriptions.length === 0 &&
-		!productDetails.some(isNotSundayTheObserverSubscription);
+		!productDetails.some(isGuardianProduct);
```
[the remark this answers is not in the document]
johnduffell: in which case I think isGuardianProduct will be unused and can be deleted

shared/productResponse.ts
rBangay: Your suggestion to use the array.every method negates the need for the `isGuardianProduct` entirely so I've removed it 👍

PR #1496, opened 2025-04-17
update setup-node action version

rBangay:
### What does this PR change?

We noticed that our chromatic action was failing with a 422 error code:

Screenshot 2025-04-17 at 12 06 03

`setup-node@v2` relies on a deprecated version of `@actions/cache`. Upgrading to v4 will hopefully fix this issue.

#### Reference
https://github.com/actions/setup-node/issues/1275#issuecomment-2788204086


MaelGNM (approved):
👍

PR #1505, opened 2025-05-09
Display future product tier

andresilva-guardian:

Currently, when a user has scheduled a future change to their subscription (such as downgrading from a higher tier membership to Contributor), this information is not displayed in the Account Overview page. Additionally, the system is incorrectly allowing users to navigate to the Membership Switch flow even when they already have a future plan change scheduled.

### What does this PR change?

This PR makes several improvements to how we handle and display future subscription plans:

1. Added a `futureProductTitle` variable in ProductCard that determines if there's a future plan change scheduled
2. Modified the product details section to display the "Next product" information when a future plan change exists
3. Updated the MembershipSwitch component to redirect users to the Overview page if they already have a future plan change to Contributor tier
4. Fixed an issue in NextPaymentDetails where we were displaying `paymentValue` instead of the current price
5. Added the tier property to the SubscriptionPlan interface to properly support future plan changes

#### Visual Changes
| Before | After |
|--------|--------|
| Screenshot 2025-05-09 at 15 19 56 | Screenshot 2025-05-09 at 15 20 28 |
* NOTE: "Next product" changed to "Switching to" by indication of @Alexguild 

### Next steps/further info

- We may want to add more comprehensive information about planned subscription changes, including when the change will take effect
- Consider adding visual indicators or badges to make future plan changes more noticeable
- We should test this thoroughly across different subscription types to ensure all edge cases are handled correctly
- Documentation should be updated to reflect these changes for the support team

### DIFF
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -442,7 +442,7 @@ export const ProductCard = ({
 									)}
 								{futureProductTitle && (
 									<div>
-										<dt>Next product</dt>
+										<dt>Switching to</dt>
 										<dd>{futureProductTitle}</dd>
 									</div>
 								)}
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -204,13 +204,13 @@ export const ProductCard = ({
 		productDetail.subscription.nextPaymentDate !==
 			productDetail.subscription.potentialCancellationDate;
 
-	const futureProductType =
+	const futureProductTitle =
 		productDetail.subscription.futurePlans.length > 0 &&
 		productDetail.subscription.futurePlans[0].tier &&
-		getSpecificProductType(productDetail.subscription.futurePlans[0].tier);
-	const futureProductTitle =
-		futureProductType && futureProductType !== specificProductType
-			? futureProductType.productTitle(mainPlan)
+		productDetail.tier
+			? getSpecificProductType(
+					productDetail.subscription.futurePlans[0].tier,
+			  ).productTitle(mainPlan)
 			: null;
 
 	return (
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -204,6 +204,15 @@ export const ProductCard = ({
 		productDetail.subscription.nextPaymentDate !==
 			productDetail.subscription.potentialCancellationDate;
 
+	const futureProductType =
+		productDetail.subscription.futurePlans.length > 0 &&
+		productDetail.subscription.futurePlans[0].tier &&
+		getSpecificProductType(productDetail.subscription.futurePlans[0].tier);
+	const futureProductTitle =
+		futureProductType && futureProductType !== specificProductType
+			? futureProductType.productTitle(mainPlan)
+			: null;
+
 	return (
 		<Stack space={4}>
 			{hasCancellationPending && productDetail.subscription.end && (
@@ -303,7 +312,7 @@ export const ProductCard = ({
 					<Card.Section backgroundColor="#edf5fA">
 						<p css={benefitsTextCss}>
 							You’re supporting the Guardian with{' '}
-							{nextPaymentDetails.paymentValue} per{' '}
+							{nextPaymentDetails.currentPriceValue} per{' '}
 							{nextPaymentDetails.paymentInterval}, and have
 							access to exclusive extras.
 						</p>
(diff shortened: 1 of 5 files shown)


client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -431,6 +440,12 @@ export const ProductCard = ({
 											</dd>
 										</div>
 									)}
+								{futureProductTitle && (
+									<div>
+										<dt>Switching to</dt>
+										<dd>{futureProductTitle}</dd>
```
johnduffell, line 446: just to understand better - if they are switching from 95 to 49 EUR is the 49 EUR (only) going to show in the the next payment amount field?
I'm not sure if there's any situation where they would have a forward dated switch to a discounted plan, so it would be ok at this point, but it does seem like the charges can be complex.

Another option (not for this PR I'm sure) would be to show a billing preview of the next few payments.

client/components/mma/shared/NextPaymentDetails.tsx
```diff
@@ -89,6 +90,9 @@ export const getNextPaymentDetails = (
 			isNewPaymentValue,
 			nextPaymentDateKey: 'Next payment date',
 			nextPaymentDateValue,
+			currentPriceValue: `${mainPlan.currency}${(
+				(subscription.plan?.price ?? 0) / 100
```
johnduffell, line 94: I saw a comment in a PR saying that subscription.plan is deprecated and we should be using current/futurePlans fields .  Not sure if they have the right information to switch over (or how aspirational that comment was)

client/components/mma/accountoverview/ProductCard.tsx
andresilva-guardian: Hey @johnduffell , the thing is, currently you can read "you're supporting the Guardian with €49 (...)" - BUT THIS IS NOT TRUE - you paid the 95 euros until that moment to have all the "exclusive extras". I believe is "false advertising" allowing the user to understand that it could get the "exclusive extras" por 49 euros when in fact, for 49 euros there will be no "exclusive extras" anymore (next year). Makes sense?

client/components/mma/cancel/cancellationSaves/membership/MembershipSwitch.tsx
```diff
@@ -262,6 +261,17 @@ export const MembershipSwitch = () => {
...
 		}
 	};
+
+	if (
+		membership.subscription.futurePlans.length > 0 &&
+		membership.subscription.futurePlans[0].tier &&
+		membership.subscription.futurePlans[0].tier === 'Contributor'
+	) {
+		// Switch already planned, so navigate to Overview
+		navigate('/');
```
johnduffell, line 271: is this just a fall back if they navigate directly to the page?  (Even if not, I'm sure it's better than letting them switch when they shouldn't.)

client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -204,6 +204,15 @@ export const ProductCard = ({
 		productDetail.subscription.nextPaymentDate !==
 			productDetail.subscription.potentialCancellationDate;
 
+	const futureProductTitle =
+		productDetail.subscription.futurePlans.length > 0 &&
+		productDetail.subscription.futurePlans[0].tier &&
```
johnduffell, line 209: nit pick-just wondering if these two lines could be combined with some kind of question mark, I'm not sure the exact syntax but this kind of thing
```suggestion
		productDetail.subscription.futurePlans[0]?.tier &&
```

client/components/mma/accountoverview/ProductCard.tsx
johnduffell: yeah got it, so when we say "you're supporting..with..." then it has to be the current charge, and actually it has to be clear when they pay the next amount it's for something else.
And actually I rechecked your screenshots and it looks pretty clear with the change of product and amount represented together. 👍

johnduffell (approved):
looks good, only a few questions/nit picks.
I noticed you said the documentation should be updated, but I didn't know which documentation that is.  I also wanted to mention that if you feel like it could cause customer confusion we can also contact the CSR and userhelp to give them a heads up.

client/components/mma/shared/NextPaymentDetails.tsx
johnduffell: this PR is the one that says plan is deprecated https://github.com/guardian/members-data-api/pull/371

client/components/mma/cancel/cancellationSaves/membership/MembershipSwitch.tsx
andresilva-guardian: @johnduffell Just to let you know, I have added a change to remove the button that redirects to the membership switch page if the future plan is to be a Contributor.

client/components/mma/shared/NextPaymentDetails.tsx
andresilva-guardian: @johnduffell I will add this to the code
Screenshot 2025-05-12 at 12 14 26

So IDEs can show this variable formatted with Strikethrough
Screenshot 2025-05-12 at 12 16 40

I will of course fix it right now also 👍

andresilva-guardian: @johnduffell Code fixed ✅

PR #1508, opened 2025-05-15
update cancel switch discount background image

rBangay:
### What does this PR change?
Update of the background image for the cancellation switch to supporter plus with a discount page

### Images

| Before  | After |
|   |   |


ecguardian (approved):
👍

PR #1509, opened 2025-05-16
make Sunday only voucher subs appear in overview

rBangay:
### What does this PR change?
Previously the voucher product (physical delivery of voucher booklets) for all days of the week would have come through MDAPI as a Guardian product, the Sunday versino of this product is now returned as a specific Observer product. This pr adds the voucher Sunday into the known product types so that it will show up in MMA.

### Images
Screenshot 2025-05-16 at 15 21 45

### DIFF
--- a/shared/productResponse.ts
+++ b/shared/productResponse.ts
@@ -123,7 +123,8 @@ export function isProduct(
 export const isObserverProduct = (productDetail: ProductDetail): boolean => {
 	return (
 		productDetail.tier === 'Newspaper Delivery - Observer' ||
-		productDetail.tier === 'Newspaper Digital Voucher - Observer'
+		productDetail.tier === 'Newspaper Digital Voucher - Observer' ||
+		productDetail.tier === 'Newspaper Voucher - Observer'
 	);
 };
 


ecguardian (approved):
LGTM 👍

client/components/mma/identity/emailAndMarketing/EmailAndMarketing.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react';
```
johnduffell, line 1: is this file just formatting changes?

johnduffell: if there are so many formatting issues it might be worth running a format on the whole codebase and putting it in as a single PR (and you can ask github to ignore it in the "blame" somehow)

shared/productTypes.ts
```diff
@@ -579,6 +581,54 @@ export const PRODUCT_TYPES: Record<ProductTypeKeys, ProductType> = {
 			swapFeedbackAndContactUs: true,
 		},
 	},
+	voucherobserver: {
+		productTitle: () => 'Newspaper Voucher - Observer',
```
johnduffell, line 585: if this is a key value that relates to MDAPI responses, we should have it once in a const, and then reference it in all the places it's used.  Or maybe it's a mixture of MDAPI and display names is it?  e.g. I just wouldn't know which to change if it changed in MDAPI for whatever reason.

shared/productTypes.ts
```diff
@@ -579,6 +581,54 @@ export const PRODUCT_TYPES: Record<ProductTypeKeys, ProductType> = {
...
+		groupedProductType: 'subscriptions',
+		allProductsProductTypeFilterString: 'Voucher',
+		urlPart: 'voucher',
+		checkoutUrlPart: '/subscribe', // https://support.theguardian.com/uk/subscribe
+		getOphanProductType: () => 'PRINT_SUBSCRIPTION',
+		productPageNewsletterIDs: [FRONT_PAGE_NEWSLETTER_ID],
+		softOptInIDs: [
+			SoftOptInIDs.SupportOnboarding,
+			SoftOptInIDs.SubscriberPreview,
+			SoftOptInIDs.SupporterNewsletter,
```
johnduffell, line 598: is this right, I thought soft opt ins were completely off for observer only?

johnduffell (approved):
nice work, could do with an improvement to the PR title, and see a few inline comments 👍

shared/productTypes.ts
johnduffell: https://github.com/guardian/support-service-lambdas/blob/14df86afac680264a811409ce93f854835bc6b94/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala#L64

rBangay: I'm going to remove these options from all of the Observer products in a subsequent pr 👍

PR #1510, opened 2025-05-19
Only UK billing country contributions should be able to switch with a discount to supporter plus (via cancellation journey)

rBangay:
### What does this PR change?
Temporarily disable contribution cancellation switch discount offer for people with a billing country outside of the UK ... until we are ready to launch worldwide.

This is a soft validation block on the client and just serves to reduce/stop calls out to the switch api for customers outside of the UK ... the switch api isn't currently setup to handle billing countries outside of the UK yet so the switch discount wouldn't have worked for them anyway.

### DIFF
--- a/client/components/mma/cancel/CancellationReasonReview.tsx
+++ b/client/components/mma/cancel/CancellationReasonReview.tsx
@@ -288,7 +288,6 @@ const ConfirmCancellationAndReturnRow = (
 		isPaidSubscriptionPlan(mainPlan) && mainPlan.billingPeriod === 'month';
 
 	const allowCountrySwitchDiscount =
-		isPaidSubscriptionPlan(mainPlan) &&
 		productDetail.billingCountry === 'United Kingdom';
 
 	const isAnnualContributionAndDiscountIsActive =
--- a/client/components/mma/cancel/CancellationReasonReview.tsx
+++ b/client/components/mma/cancel/CancellationReasonReview.tsx
@@ -287,8 +287,13 @@ const ConfirmCancellationAndReturnRow = (
 	const isMonthlyBilling =
 		isPaidSubscriptionPlan(mainPlan) && mainPlan.billingPeriod === 'month';
 
+	const allowCountrySwitchDiscount =
+		isPaidSubscriptionPlan(mainPlan) &&
+		productDetail.billingCountry === 'United Kingdom';
+
 	const isAnnualContributionAndDiscountIsActive =
 		productType.productType === 'contributions' &&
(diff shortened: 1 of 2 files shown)


client/components/mma/cancel/CancellationReasonReview.tsx
```diff
@@ -287,8 +287,13 @@ const ConfirmCancellationAndReturnRow = (
...
 
+	const allowCountrySwitchDiscount =
+		isPaidSubscriptionPlan(mainPlan) &&
+		productDetail.billingCountry === 'United Kingdom';
+
 	const isAnnualContributionAndDiscountIsActive =
 		productType.productType === 'contributions' &&
+		allowCountrySwitchDiscount &&
 		isAnnualBilling &&
 		reasonIsEligibleForSwitch(routerState.selectedReasonId);
```
johnduffell, line 298: should be 

- Annual Recurring Contributor - YES
- UK Only (for now) - YES (now)
- isActive: 'subscription status is active', - YES
- zeroAccountBalance: 'account balance is zero' - YES
- noNegativePreviewItems: next invoice has no negative items - YES
- RC must be paying under or equal to the discounted Supporter Plus price - YES
- any time after acquisition - YES (by default)
- Can be redeemed only once per subscription - YES (because you can't switch back again)
- If they select: “I’m taking a break from news”, “I’m not fully satisfied with the product features or benefits”, “I can no longer afford to support you”, “I don’t have enough time to read regularly”, “I want to support in a different way”, “I don’t feel that the Guardian values my support” - YES (reasonIsEligibleForSwitch)

client/components/mma/cancel/CancellationReasonReview.tsx
```diff
@@ -287,8 +287,13 @@ const ConfirmCancellationAndReturnRow = (
 	const isMonthlyBilling =
 		isPaidSubscriptionPlan(mainPlan) && mainPlan.billingPeriod === 'month';
 
+	const allowCountrySwitchDiscount =
+		isPaidSubscriptionPlan(mainPlan) &&
```
johnduffell, line 291: why do we need to check isPaidSubscriptionPlan everywhere?  is it just ruling out the old Friend/Staff?

johnduffell (approved):
nice one, get rid of those errors :clap:

client/components/mma/cancel/CancellationReasonReview.tsx
rBangay: we don't! good spot, that was there when I thought that the billingCountry was a property on the main plan object which is related to paid subscription plans

PR #1511, opened 2025-05-19
remove Observer product softOptInIDs

rBangay:
### What does this PR change?
Removes any elements from the `softOptInIDs` array in the productTypes file for all Observer products. The consequence of doing this is to remove the supporter email opt ins from the emails and marketting page for people that only have Observer products ... these emails are Guardian related.

### Images
Before             |  After
  |


johnduffell (approved):
nice, matches with our rule here https://github.com/guardian/support-service-lambdas/blob/237352e53c6f80c08eba19cf141646edb0803010/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala#L64

PR #1512, opened 2025-05-20
chromatic regression fix for cancellation reasons

rBangay:
### What does this PR change?
Supply a sorted cancellation reasons array to storybook stories so that we don't trip the chromatic visual regression tests everytime we make any change in manage.

The reason chromatic was reporting a visual regression/diff is because on the site the cancellation reasons order in randomized so as not to skew the options people select.

An alternative approach so sorting the reasons in storybook might have been to add the `data-chromatic=ignore` attribute to the components that render the reasons. I didn't choose this option as I thought it was better not to pollute the dom with test attributes.

### Images
Screenshot 2025-05-20 at 15 48 01


andresilva-guardian (approved):
Hey amazing work @rBangay ! Thanks for the effort! 💪 🎉 
Your PR achieves what you're targeting and makes perfect sense for addressing the Chromatic visual regression issue. The changes are well-focused and should eliminate the false-positive regression reports without compromising your testing or component design. Keep going strong! 😎 
Thank you for improving our lives! 😉

PR #1523, opened 2025-07-02
Moves code to the use of new property mmaProductKey

j-ruda-guardian:

PR makes the shift from the previously used `tier` property to the new `mmaProductTier`, as part of the effort to release newspaper+digital benefits.

### What does this PR change?

Changes to use the new property, naming of some functions and everything needed to support that change.

### DIFF
--- a/client/components/helpCentre/KnownIssues.tsx
+++ b/client/components/helpCentre/KnownIssues.tsx
@@ -29,7 +29,7 @@ export interface KnownIssueObj {
 	date: string; // "29 Aug 1997 02:40"
 	message: string;
 	link?: string; // optional href that is rendered after the message
-	affectedProducts?: string[]; // maps to productDetail.tier property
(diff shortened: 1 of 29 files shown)


rBangay (approved):
Nice, hopefully this will make the relationship between productType objects in MMA and the response from MDAPI a bit clearer 👍

johnduffell:
can you update the PR description to include a link to the mdapi PR and more background info please?  This will really help future people investigating changes made.

Something like this would be helpful:

---

Currently we use the historically named `tier` property to decide what product functions to offer the user.  This aligns with the Zuora product, apart from a special case for Observer only subscriptions.  We would like to release newspaper+digital benefits, and in future, Tier 3 digital-only guardian weekly access.

As we decouple manage functionality from the zuora product, it makes sense to name the property according to its function, to aid future maintainers of the code.

### What does this PR change?

Changes to use the new `mmaProductTier` property introduced in https://github.com/guardian/members-data-api/pull/1137

PR #1524, opened 2025-07-02
Revert "Moves code to the use of new property mmaProductKey"

johnduffell:
Reverts guardian/manage-frontend#1523

Reverting because the above PR doesn't include the new `+ Digital` in the getSpecificProductTypeFromProductKey function, so anyone on paper+ no longer can see their product.


ecguardian (approved):
👍

PR #1530, opened 2025-07-09
Upgrading Typescript and some @types versions. Includes cookie-parser upgrade

j-ruda-guardian:

We want to upgrade Typescript to the latest version, which includes a major upgrade. That went easy, but as part of this work, I noticed that the nvmrc of this project mentions v22, but @types of node was on 18. Upgrading that library required cookie-parser to solve some typing issues.

It's still pending to upgrade helmet. I'll do it on another PR 

### What does this PR change?

Libraries upgrades and making sure that everything works fine without warnings going off

### Next steps/further info

Upgrade helmet to 8.1 in a separate PR.

I've tested that everything works both locally and by deploying to CODE and having a pass through the app


server/idapiProxy.ts
```diff
@@ -95,7 +98,10 @@ const idapiOrOAuthHeaders = ({
 			'X-GU-ID-Client-Access-Token': `Bearer ${config.accessToken}`,
 			...securityCookieToHeader(cookies),
 			// Avatar API expects a Cookie header with the SC_GU_U cookie.
-			Cookie: subdomain === 'avatar' ? `SC_GU_U=${cookies.SC_GU_U};` : '',
+			Cookie:
+				subdomain === 'avatar' && cookies.SC_GU_U
+					? `SC_GU_U=${cookies.SC_GU_U};`
+					: '',
```
johnduffell, line 104: so we actually need to omit the cookie completely if the value is falsley, whereas before we would have  (incorrectly?) set it with a potentially problematic value? (but that wasn't noticed for some reason? because people only hit the endpoint with cookie set?)

server/idapiProxy.ts
```diff
@@ -28,8 +28,8 @@ interface IdapiFetchOptions {
 }
 
 interface CookiesWithToken {
-	SC_GU_U: string;
-	[key: string]: string;
+	SC_GU_U?: string;
+	[key: string]: string | undefined;
```
johnduffell, line 32: it doesn't feel like having a cookie with no value makes sense, is that really a thing?

server/idapiProxy.ts
j-ruda-guardian: Exactly. But the strictier typing doesn't allow for what we were doing before. The check will set the cookie safely, when it's a valid value

server/idapiProxy.ts
johnduffell: having checked, it seems like you must have a value (the `=` is mandatory) but it may have an empty string as the value (there may be 0 or more characters after the `=`)

https://datatracker.ietf.org/doc/html/rfc6265#:~:text=cookie%2Dpair%20%20%20%20%20%20%20%3D%20cookie%2Dname%20%22%3D%22%20cookie%2Dvalue

So it might be that the upstream thing needs tweaking rather than allowing undefined?

server/server.ts
```diff
@@ -38,7 +38,7 @@ if (conf.DOMAIN === 'thegulocal.com') {
 	process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 }
 
-server.use(helmet());
+server.use(helmet() as unknown as RequestHandler);
```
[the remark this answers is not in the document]
johnduffell: got it ok it's a compatibility issue and this will be not needed very soon

server/idapiProxy.ts
j-ruda-guardian: Actually, there's a better approach to this. Thanks for the questions, it made me realize something. Im updating the PR soon

server/idapiProxy.ts
johnduffell: ah I see that you made another change while I was trying to understand!  But I'd actually worked out that it would be undefined if the underlying cookie value was not a recognised format for the cookie-parser library.
This means that the "cookie" value returned by this library could be
- not present at all if it wasn't set at all, or
- undefined if it was set to something not starting with `"j:"`

johnduffell (approved):
looks good and surprisingly few changes for a major upgrade like this!
I don't fully understand the `CookiesWithToken` changes so you may want to get a review off someone more typescript knowledgable (I'm a TS learner at the moment hence lots of questions and not so many answers!)

server/idapiProxy.ts
```diff
@@ -27,10 +27,7 @@ interface IdapiFetchOptions {
 	};
 }
 
-interface CookiesWithToken {
-	SC_GU_U: string;
-	[key: string]: string;
-}
+type CookiesWithToken = Record<string, string>;
```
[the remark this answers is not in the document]
j-ruda-guardian: true. That depends where you want the compromise to be, because node 22 won't allow for such flexible interfaces. If we want CookiesWithToken to imply the precense of SC_GU_U (which makes sense) we need to cast req.cookies as CookiesWithToken, so TS doesn't complain. It's a little strange?? But also could be ok.

```const options = setOptions({
			sendAuthHeader,
			useOkta,
			path: url,
			subdomain: 'idapi',
			method,
			cookies: req.cookies as CookiesWithToken,
			signedCookies: req.signedCookies as CookiesWithToken,
			config,
		});```
		
what do you think?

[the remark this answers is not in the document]
rBangay: I see what you mean. Cookie properties much like anything loaded asynchronously have inherently unknowable properties.

I'm happy not having the 'SC_GU_U' property typed explicitly so long as we have considered what will happen if that property doesn't exist 👍

rBangay (approved):
Thank you for taking this on

server/idapiProxy.ts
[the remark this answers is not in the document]
johnduffell: perhaps we should be rejecting the incoming request with a 4xx error if it's not set? rather than trying to call downstream (identity) and waiting for it to complain?

PR #1547, opened 2025-09-08
[DNM] Fixes issue with cancellation endpoints not passing testUser headers

j-ruda-guardian:

I noticed this issue while testing the new Student offer's cancellation journey. This is preventing me from testing in CODE at the moment.

### What does this PR change?

Fixes the issue with test user header not being passed properly for the cancellation and preview-discount endpoints.


johnduffell:
@j-ruda-guardian that message is supposed to be a warning and should only block using test users, not using CODE directly (using test users only makes sense in PROD, as it points the app at the CODE lambdas even in PROD)

(Also this PR is marked as [DNM] which I assume is "do not merge", if it is supposed to be merged it's worth removing that prefix)

PR #1549, opened 2025-09-10
Move the diagnostic-information route behind sign in

rBangay:
Currently you can access the diagnostic-information route (https://manage.theguardian.com/help-centre/diagnostic-information)  publicly without having to sign in

### What does this PR change?
Puts the diagnostic-information route (https://manage.theguardian.com/help-centre/diagnostic-information) behind the sign in gate

## WHY!?
This is a request from user help. If you visit the diagnostic-information route while you are signed in it provides you with more useful information.

This is a fairly typical user help customer interaction:

They need to ask users to check if a subscription is coming through.

- So they send users to this url to capture their cookies
- and the users goes there and sends us the debug information
- but they haven’t signed in (which is neccessary for the sub info)
- so they have to tell them to sign in and go again

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,5 +184,6 @@
   "license": "UNLICENSED",
   "msw": {
     "workerDirectory": ".storybook/static"
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
 }
--- a/shared/requiresSignin.ts
+++ b/shared/requiresSignin.ts
@@ -1,7 +1,9 @@
 import { normalize } from 'path'; // webpack polyfills this for the browser via node-polyfill-webpack-plugin
 
+type PublicPath = string | { path: string; exclude: string };
+
 // To avoid security vulnerabilities do not add public paths that do not end in a slash
-const publicPaths = [
+const publicPaths: PublicPath[] = [
 	'/api/contact-us/',
 	'/api/known-issues/',
 	'/api/reminders/cancel/',
@@ -10,7 +12,10 @@ const publicPaths = [
 	'/api/csp-audit-report-endpoint/',
 	'/cancel-reminders/',
 	'/create-reminder/',
-	'/help-centre/',
+	{
+		path: '/help-centre/',
+		exclude: '/help-centre/diagnostic-information/signed-in/',
+	},
 	'/maintenance/',
 	'/sign-in-error/',
 ];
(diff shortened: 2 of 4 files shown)


j-ruda-guardian (approved):

shared/requiresSignin.ts
```diff
@@ -1,7 +1,9 @@
 import { normalize } from 'path'; // webpack polyfills this for the browser via node-polyfill-webpack-plugin
 
+type PublicPath = string | { path: string; exclude: string };
```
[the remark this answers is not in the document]
rBangay: Hey 👋  ... if I change the shape of the type to exclude a string option then I'd have to change the publicPaths from 

```
const publicPaths: PublicPath[] = [
	'/api/contact-us/',
	'/api/known-issues/',
	{path: '/yo/', exclude: 'yo/no/'}
```

to something like this where you have to define an object for each entry:
```
```
const publicPaths: PublicPath[] = [
	{path:'/api/contact-us/'},
	{path: '/api/known-issues/'},
	{path: '/yo/', exclude: 'yo/no/'}
```

I can do that if that's what people want, I just wanted to make the minimum changes to the structure in order to facilitate signed in routes within other public routes.

File: package.json [hunk 2]
```diff
@@ -184,5 +184,6 @@
   "license": "UNLICENSED",
   "msw": {
     "workerDirectory": ".storybook/static"
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
```
johnduffell, line 188: what's this?

shared/requiresSignin.ts
```diff
@@ -10,15 +12,25 @@ const publicPaths = [
 	'/api/csp-audit-report-endpoint/',
 	'/cancel-reminders/',
 	'/create-reminder/',
-	'/help-centre/',
+	{
+		path: '/help-centre/',
+		exclude: '/help-centre/diagnostic-information/',
```
johnduffell, line 17: just wondering did you consider moving diagnostic information out of the help centre to avoid this?

package.json
rBangay: I got that for free when I ran `yarn` ... by the looks of this issue (https://github.com/nodejs/node/issues/51888) it might be on the way out anyway and isn't necessary, so I'll bin it off ...

johnduffell (approved):
👍

PR #1554, opened 2025-09-17
Change Cancellation Message for one-off payment and renewal status in ProductCard

andresilva-guardian:

After an e2e testing for the student page, a reference to recurring payment and cancellation on MMA was spotted. After some research we found out that the "Cancellation Subscription" feature only is displayed for subscriptions with United States as billing country, making it the reason why we missed this to begin with.

### What does this PR change?

Changes the copywriting of this feature when a subscription as no auto-renew and as no next Payment Date.

### Next steps/further info

The original conversation could be read [here](https://chat.google.com/room/AAQAK5a5stM/aE9hJ-FO4dw/aE9hJ-FO4dw?cls=10).

### New appearance
| Before    | After |
| -------- | ------- |
| Screenshot 2025-09-17 at 15 04 35 | Screenshot 2025-09-17 at 16 03 13 |

### DIFF
--- a/client/components/mma/accountoverview/ProductCard.tsx
+++ b/client/components/mma/accountoverview/ProductCard.tsx
@@ -566,17 +566,24 @@ export const ProductCard = ({
 											max-width: 350px;
 										`}
 									>
-										{!productDetail.subscription.autoRenew &&
+										{!productDetail.subscription
+											.autoRenew &&
 										!productDetail.subscription
-											.nextPaymentDate ? (<>
-												This is a one-off payment and will
-												not renew. You’ll continue to enjoy
-												your benefits until the end of the
-												current billing period.
-											</>) : (<>
-												Stop your recurring payment, at the
-												end of current billing period.
-											</>)}
+											.nextPaymentDate ? (
+											<>
+												This is a one-off payment and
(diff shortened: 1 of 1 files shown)


client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -561,14 +561,29 @@ export const ProductCard = ({
...
+									) : (
+										<p
+											css={css`
+												max-width: 350px;
+											`}
+										>
+											Stop your recurring payment, at the
+											end of current billing period.
+										</p>
+									)}
```
johnduffell, line 586: I think the styling is the same each way
```suggestion
									<p
										css={css`
											max-width: 350px;
										`}
									>
										{!productDetail.subscription.autoRenew &&
										!productDetail.subscription
											.nextPaymentDate ? (<>
												This is a one-off payment and will
												not renew. You’ll continue to enjoy
												your benefits until the end of the
												current billing period.
											</>) : (<>
												Stop your recurring payment, at the
												end of current billing period.
											</>)}
									</p>
```

johnduffell (changes requested):
wasn't sure whether I should do request changes actually! it's kind of a conditional approval

PR #1629, opened 2026-03-30
Print Products cancellation Journey implementation

j-ruda-guardian:

The cancellation journey for print products has been redesigned, but this rollout is now intentionally scoped to **Guardian Weekly only**. The branch introduces a dedicated print cancellation journey architecture while keeping existing non-print journeys on their router-state path.

This PR completes that Guardian Weekly print journey end-to-end across selection, review, confirm, and execute, and aligns the flow with a print-specific state model.

### What this PR changes

This branch introduces a dedicated Guardian Weekly print cancellation path that uses:
- a print loader (`usePrintCancellationLoader`)
- a print store (`PrintCancellationStore`)
- print-specific UI paths in reason selection, review, confirm, and execute stages

Cancellation behavior remains config-driven in `shared/productTypes.ts`, with Guardian Weekly mapped to the new printed reasons set and existing print flow wrapper configuration. Entry to the new journey is gated via `isPrintProduct(...)`, currently dialed to `guardianweekly` only.

The branch also consolidates Guardian Weekly print reasons into `PrintedProductsCancellationReasons.tsx` (including `mma_duplicate_subscription`) and removes the now-redundant GW-specific reasons source.

### State/routing model and regressions prevention

For Guardian Weekly print, cross-stage journey data (selected reason, case id, credits, etc.) is owned by `PrintCancellationStore`, with route guards and loader hydration supporting direct entry and refresh cases.

For legacy/non-print products, existing router-state flow remains in place. Recent fixes in this branch specifically ensure:
- print feedback captured in step 1 is now persisted to the case before continuing
- print journey state is reset per new journey entry to avoid stale state reuse
- minimal-footer treatment is scoped to print only, avoiding unintended non-print UI regressions

### UI and flow behavior

The Guardian Weekly print journey now provides:
- print-specific reason picker and copy
- print-specific review step with pause/advisor content
- print-specific confirm experience
- print-specific execute/success experience with branded confirmation treatment

`MMAPageSkeleton` matching was also updated for nested cancellation routes to prevent skeleton transition issues.

### Scope / impact

Primary impact: Guardian Weekly cancellation journey.  
Secondary impact: shared cancellation container/loader wiring and product-type configuration.  
Non-print behavior: preserved on router-state flow, with regression safeguards added where shared code paths were touched.


client/utilities/productUtils.ts
```diff
@@ -117,6 +117,21 @@ export const hasCancellationFlow = (
...
+		'homedeliveryplusdigital',
+		'nationaldelivery',
+		'nationaldeliveryplusdigital',
+		'voucher',
+		'voucherplusdigital',
+		'digitalvoucher',
+		'digitalvoucherplusdigital',
+		'tierthree',
+	].includes(productType.productType);
+};
```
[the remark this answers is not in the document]
j-ruda-guardian: I can literally see `voucherplusdigital` in the list

client/components/mma/MMAPageSkeleton.tsx
```diff
@@ -133,12 +133,14 @@ const MMALocationObjectArr: LocationObject[] = [
 
 export const MMAPageSkeleton = () => {
 	const location = useLocation();
```
j-ruda-guardian, line 135: this change is needed to make the skeleton page work on nested routes, not just exact path matches.

shared/productTypes.ts
```diff
@@ -751,7 +751,7 @@ export const PRODUCT_TYPES: Record<ProductTypeKeys, ProductType> = {
 			},
 		},
 		cancellation: {
-			reasons: shuffledGWCancellationReasons,
+			reasons: shuffledPrintedProductsCancellationReasons,
```
johnduffell: tiny nitpick - normally we call them print products
```suggestion
			reasons: shuffledPrintProductsCancellationReasons,
```

client/components/mma/cancel/PrintProductsCancellationReasons.tsx
```diff
@@ -0,0 +1,61 @@
+import { shuffleArray } from '@/client/utilities/utils';
+import { standardAlternateFeedbackIntro } from './cancellationConstants';
+import type { CancellationReason } from './cancellationReason';
+
+const printedProductsCancellationReasons: CancellationReason[] = [
+	{
+		reasonId: 'mma_dont_read_enough',
+		linkLabel: "I don't use the subscription enough to justify the cost",
+		alternateFeedbackIntro: standardAlternateFeedbackIntro,
```
johnduffell, line 9: nitpick: it looks like elsewhere "alternative" is used e.g. `PrintCancellationAlternatives` but in this file "alternate" is used, I prefer to keep to "alternative"

Having said that maybe the naming needs to be more specific? As I'm not sure what "standard alternative" means - it sounds like a contradiction

client/utilities/productUtils.ts
```diff
@@ -117,6 +117,14 @@ export const hasCancellationFlow = (
 export const hasDeliveryFlow = (productType: ProductType) =>
 	productType.delivery?.showAddress;
 
+// Release gate for the new Print Cancellation flow. Returns true only for
+// products that should enter the print-specific cancellation journey
+// (dedicated reason picker, alternatives, confirm and execute pages plus
+// the PrintCancellationStore). Any product type applicable to the new
+// flow should be added to the list.
+export const usesPrintCancellationFlow = (productType: ProductType): boolean =>
+	['guardianweekly'].includes(productType.productType);
```
johnduffell: ai assisted: I was discussing architectural improvements with copilot and it came up with this which I thought is probably fine but still worth considering, as I never know the PRODUCT_TYPES vs normal if statements tradeoff:

---
The print-flow gate is procedural, not declarative This repo already has a strong PRODUCT_TYPES configuration model, but the flow split is controlled by a hardcoded helper:

For this system, a cleaner convention would usually be one of:
- a flag on ProductType.cancellation, e.g. journeyModel: 'router-state' | 'store'
- a grouped print capability flag
- or a print-specific flow wrapper/config object in productTypes.ts

The current approach is okay for a rollout gate, but architecturally it’s a bit ad hoc.

client/components/mma/cancel/stages/ExecuteCancellation.tsx
```diff
@@ -213,8 +222,8 @@ export const ExecuteCancellation = () => {
 
 	return (
 		<>
-			{(alternativeIsOffer && !routerState.eligibleForFreePeriodOffer) ||
-				(alternativeIsPause && !routerState.eligibleForPause && (
+			{(alternativeIsOffer && !eligibleForFreePeriodOffer) ||
```
johnduffell, line 222: ai assisted: I realise it's existing, but I think it's a good point about complexity of expressions - if you think it is worth changing it's possibly one for another PR

---
Using `||` to select JSX is legal, but it’s not idiomatic and it’s easy to get wrong because a bare boolean can short-circuit rendering in surprising ways. Conventionally this should be shaped as:

- a boolean computed first, then condition && <UI />
- or a named helper variable

client/components/mma/cancel/cancellationPrint/printReasonPicker.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+	${from.tablet} {
+		display: flex;
+	}
+`;
+
+export const PrintReasonPicker = ({ productType }: PrintReasonPickerProps) => {
+	const cancellationContext = useContext(
+		CancellationContext,
+	) as CancellationContextInterface;
```
johnduffell: this could be `object` (which I guess is `{}`), what will happen if it fails to cast?

johnduffell: there are a few more `as` in the PR, perhaps worth a once over to see if any can be closed up easily?

client/components/mma/cancel/CancellationReasonReview.tsx
```diff
@@ -14,10 +14,14 @@ import {
 import type { ChangeEvent, FC } from 'react';
```
johnduffell, line 14: ai assisted observation: I think this is probably one for the next refactor, but I think it's a good spot as this has grown to 900 lines.  I can't vouch for the actual suggestion of how to split it, or the urgency of doing so!

---
CancellationReasonReview.tsx now carries:

- routing concerns
- product-type branching
- async data loading
- case persistence
- page-title effects
- and render branching between print/non-print UI

That’s survivable, but it’s pushing beyond ideal React component cohesion. Conventionally, once a component starts owning both orchestration and substantial UI divergence, it’s a sign to split earlier.

client/components/mma/MMAPageSkeleton.tsx
```diff
@@ -133,12 +133,14 @@ const MMALocationObjectArr: LocationObject[] = [
...
+	const isPathMatch = (pathname: string, basePath: string) =>
+		pathname === basePath ||
+		pathname === `${basePath}/` ||
```
johnduffell: can we get rid of this one, as `${basePath}/` definitely starts with `${basePath}/`?
or is it needed to handle the `/` case? in which case we might be able to be more explicit.

it may be possible to strip a single trailing slash before starting work to avoid the first check too?

andresilva-guardian (approved):

client/components/mma/cancel/PrintProductsCancellationReasons.tsx
j-ruda-guardian: I did name one step of the flow as PrintCancellationAlternatives, so it was introduced by me as a description of what the step does. But the app uses alternateFeedbackIntro and standardAlternateFeedbackIntro as a legacy properties. THAT BEING SAID, the new flow doesn't use it :) Im removing these

client/utilities/productUtils.ts
j-ruda-guardian: This is actually a great suggestion that I oversaw. Let me see if I can introduce the right pattern quickly

j-ruda-guardian: Ok it was straight forward. I will push it in the next commit

client/components/mma/cancel/stages/ExecuteCancellation.tsx
j-ruda-guardian: I think this is a bit of a nitpick from the AI :) But we can take a look later for a different PR

client/components/mma/cancel/cancellationPrint/printReasonPicker.tsx
j-ruda-guardian: Yeah this is generally accepted but considered a smell. Let me give it a fix

client/components/mma/cancel/CancellationReasonReview.tsx
j-ruda-guardian: Yeah this file was actually one of the big challenges. Let's leave this for the future, This is something I really didn't want to touch in this PR

client/components/mma/MMAPageSkeleton.tsx
j-ruda-guardian: This is a nice suggestion :) I will apply it for the next commit

client/components/mma/cancel/stages/ExecuteCancellation.tsx
johnduffell: yeah in that case ignore!

johnduffell (approved):
looks great, I think the review I was doing was more for my learning benefit than yours, and Andre has kindly looked over it, so I think go ahead when you're ready 👍 

I think it's great that you've had time to go through and bring the code up to modern standards as part of the wider work 👏

PR #1631, opened 2026-04-08
fixes a duplicate preview request in the cancellation flow

j-ruda-guardian:

This PR fixes a duplicate preview request in the cancellation flow for annual contribution switch-offer reasons.

Currently, both ContributionsCancellationFlowFinancialSaveAttempt and ConfirmCancellationAndReturnRow can mount on the same review screen and trigger /api/product-move/recurring-contribution-to-supporter-plus/:subscriptionName preview calls in quick succession, which can cause Zuora lock-competition errors. This change makes ContributionsCancellationFlowFinancialSaveAttempt the single owner for that preview path by skipping ConfirmCancellationAndReturnRow when that save body is active, preserving existing UI while removing the duplicate API call.

Tested by running the affected annual contribution cancellation journeys (including eligible switch-offer reasons), then checking browser network calls and backend logs. I verified only one preview=true request is sent to /api/product-move/recurring-contribution-to-supporter-plus/:subscriptionName per page load and confirmed the flow still behaves correctly.


client/components/mma/cancel/CancellationReasonReview.tsx
```diff
@@ -100,6 +101,10 @@ const ContactUs = (reason: CancellationReason) =>
 		</p>
 	);
 
+const isFinancialSaveAttemptSaveBody = (
+	saveBody?: CancellationReason['saveBody'],
+): boolean => saveBody === ContributionsCancellationFlowFinancialSaveAttempt;
```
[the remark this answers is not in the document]
j-ruda-guardian: Not needed for this PR

client/components/mma/cancel/CancellationReasonReview.tsx
```diff
@@ -167,14 +172,16 @@ const FeedbackFormAndContactUs = (props: FeedbackFormProps) => {
...
-			/>
+			{!isFinancialSaveAttemptSaveBody(props.reason.saveBody) && (
+				<ConfirmCancellationAndReturnRow
+					hide={!!props.reason.hideSaveActions}
+					reasonId={props.reason.reasonId}
+					productType={props.productType}
+					caseId={props.caseId}
+					holidayStops={props.holidayStops}
+					deliveryCredits={props.deliveryCredits}
+				/>
```
johnduffell, line 183: so we were putting the component there before, which was causing the second call. But now we are only putting it there if it's not a financial save?
I would expect to see "if financial save then display X else display Y" or is this something that's omitted completely?

Might be useful to have some screenshots of what each of these e.g. `ConfirmCancellationAndReturnRow` relate to, to help review things.

johnduffell (approved):
thanks for fixing! I don't really understand the fix, but since you've tested it, I don't think that blocks it.  See comment though for what I was confused by.

client/components/mma/cancel/CancellationReasonReview.tsx
j-ruda-guardian: So ConfirmCancellationAndReturnRow, it was visually hidden (hideSaveActions, renders nothing in this case) but it still mounted, and its useEffect still ran, causing a second preview call. ContributionsCancellationFlowFinancialSaveAttempt (the financial-save component) already does the preview call. That's why we were getting the repeated API call. 

Now we don’t mount ConfirmCancellationAndReturnRow at all for financial-save reasons. No mount -> no second useEffect -> no duplicate preview call.

PR #1636, opened 2026-04-27
Setups Users on Sentry's init

j-ruda-guardian:

Sentry alerts in manage-frontend were reporting 0 users affected because we initialized Sentry but never attached user identity to events. That meant errors had no user context in client paths.

### What does this PR change?

- Added client-side analytics user binding in client/utilities/analytics.ts
- Wired this helper into both client entrypoints right after Sentry.init(...)
- Added/updated tests


client/utilities/analytics.ts
```diff
@@ -60,3 +63,22 @@ export const trackEvent = ({
 };
 
 export const trackEventInOphanOnly = (event: Event) => trackEvent(event);
+
+export const setAnalyticsUserFromIdentity = (
```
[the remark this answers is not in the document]
j-ruda-guardian: Let's keep the name generic for this function, as this file handles analytics independent of the provider

johnduffell (approved):
great idea, hope we can get a better idea if it's one user repeating or many users hitting the same issue

client/utilities/analytics.ts
```diff
@@ -60,3 +62,15 @@ export const trackEvent = ({
 };
 
 export const trackEventInOphanOnly = (event: Event) => trackEvent(event);
+
+export const setAnalyticsUserFromBrowserId = () => {
```
graham228221: Might slightly rename this to be clearer:

```suggestion
export const setAnalyticsUserFromConsentCookie = () => {
```

client/utilities/analytics.ts
```diff
@@ -60,3 +62,15 @@ export const trackEvent = ({
...
+	const consentDate = getCookie('consentDate');
```
graham228221, line 67: For our future reference:

Why are we using `consentDate` as a proxy for a user identifier?

We do not have Data Privacy sign off for storing PII data in Sentry, so instead we are using `consentDate` to represent the user session - this is the datestamp from the users last consent banner choice. 

While not strictly unique, it is probably unique enough for the purposes of error monitoring. And as it's not PII, we don't have any Data Privacy concerns.

graham228221 (approved):
Thanks for this and sorry again for the back-and-forth! 

Just a suggestion to rename the function slightly to better reflect where the identifier is coming from, but I'd leave that up to you.

PR #1638, opened 2026-04-28
Sentry crashes caused by brittle date/state handling in MMA

j-ruda-guardian:

This PR fixes two Sentry crashes caused by brittle date/state handling in MMA flows.

Background: we saw production errors from account/cancellation journeys when APIs returned partial data (or route state was missing after hash/deep-link style navigation), which led to runtime exceptions instead of graceful fallbacks.

https://the-guardian.sentry.io/issues/7257798041/?environment=theguardian.com&project=1208603&query=is%3Aunresolved&referrer=issue-stream&sort=freq
https://the-guardian.sentry.io/issues/6542307010/?environment=theguardian.com&project=1208603&query=is%3Aunresolved&referrer=issue-stream&sort=freq

### What does this PR change?

- Hardened account overview date handling, guarded PersonalisedHeader against empty/invalid date inputs, guarded InAppPurchaseCard cancellation-date formatting
- Hardened cancellation flow routing, added null-safe guards in ConfirmCancellation and ExecuteCancellation for missing location.state, redirecting safely instead of crashing
- Added regression tests for invalid/missing dates in account overview components, missing route state in cancellation confirm/execute stages

### DIFF
--- a/client/components/mma/accountoverview/InAppPurchaseCard.tsx
+++ b/client/components/mma/accountoverview/InAppPurchaseCard.tsx
@@ -2,6 +2,7 @@ import { css } from '@emotion/react';
 import { space, textSans17 } from '@guardian/source/foundations';
 import { Button, Stack } from '@guardian/source/react-components';
 import { InfoSummary } from '@guardian/source-development-kitchen/react-components';
+import { isValid } from 'date-fns';
 import { capitalize } from 'lodash';
 import { useNavigate } from 'react-router';
 import { dateString } from '../../../../shared/dates';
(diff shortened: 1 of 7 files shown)


client/components/mma/accountoverview/InAppPurchaseCard.tsx
```diff
@@ -57,12 +58,17 @@ export const InAppPurchaseCard = ({
...
+	const hasValidCancellationDate =
+		!!cancellationDate && isValid(cancellationDate);
 	return (
 		<Stack space={4}>
-			{subscription.cancellationTimestamp && (
+			{hasValidCancellationDate && (
 				<InfoSummary
 					message={`Your app subscription was cancelled in ${dateString(
-						new Date(subscription.cancellationTimestamp),
+						cancellationDate,
```
johnduffell, line 73: I'm having trouble getting my head around this - so if `subscription.cancellationTimestamp` exists then we pass it in to the `Date` constructor, which may successfully return a date object that fails `isValid`

I guess we don't use zod or similar?  It would be good to filter out invalid dates as early as possible so we don't need to worry later in our business logic.

johnduffell (approved):
looks reasonable, I can't claim to fully understand but if it's tested ok then go for it.

It would be good to add specific detail on what each fix is meant to handle - perhaps as inline PR self-comments.  Apart from helping reviewers like myself, this will also help if someone is refactoring or fixing edge cases in future and wonders why the check was added.

client/components/mma/cancel/stages/ConfirmCancellation.tsx
```diff
@@ -114,19 +114,23 @@ export const ConfirmCancellation = () => {
 
 	const isInTrialPeriod = subscription.trialLength > 0;
 
+	useEffect(() => {
+		pageTitleContext.setPageTitle(
+			`Cancel ${groupedProductType.friendlyName}`,
+		);
+	}, [groupedProductType.friendlyName, pageTitleContext]);
+
+	if (!routerState) {
```
j-ruda-guardian, line 123: Missing routerState can happen when people navigate directly to the flow without clicking the button, hence missing the routerState being populated on click

PR #1643, opened 2026-05-15
I-100 | Wrong benefits showing for GW Gift subs in MMA

enesihsa:
Guardian Weekly Gift benefits shown to the gifter currently include the weekly print delivery too, which is incorrect. Changes to the product description and the benefits in the product card were needed. 

### What does this PR change?
ProductCard now displays a special text and benefits configuration if the user has the `readerType: 'Gift'` and `specificProductType.productType === 'guardianweekly'`. In addition a redundant case (`guardianWeeklyGiftRecipient`) in the storybook was removed, since the giftee cannot be linked to a Guardian account in any way. 

#### Screenshots from CODE
image

### Next steps/further info
My previous questions were answered by Graham. 

No cancellation or address change flows were needed since it is not possible to change the address of the giftee or cancel it as it is a pre-paid gift subscription. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214505396536938

### DIFF
--- a/client/fixtures/productBuilder/testProducts.ts
+++ b/client/fixtures/productBuilder/testProducts.ts
@@ -7,7 +7,6 @@ import {
 	baseDigitalVoucherPlus,
 	baseGuardianAdLite,
 	baseGuardianWeekly,
-	baseGuardianWeeklyGift,
 	baseHomeDelivery,
 	baseHomeDeliverySaturdayPlus,
 	baseMembership,
@@ -42,7 +41,7 @@ export function guardianWeeklyPaidByCard() {
 }
 
 export function guardianWeeklyGiftPurchase() {
-	return new ProductBuilder(baseGuardianWeeklyGift())
+	return new ProductBuilder(baseGuardianWeekly())
 		.payByCard()
 		.gift(true)
 		.getProductDetailObject();
--- a/client/fixtures/productBuilder/testProducts.ts
+++ b/client/fixtures/productBuilder/testProducts.ts
@@ -36,7 +36,7 @@ export function voucherPlusPaidByCard() {
 }
 
 export function guardianWeeklyPaidByCard() {
-	return new ProductBuilder(baseGuardianWeeklyGift())
+	return new ProductBuilder(baseGuardianWeekly())
 		.payByCard()
 		.getProductDetailObject();
 }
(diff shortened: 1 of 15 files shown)


client/fixtures/productBuilder/testProducts.ts
```diff
@@ -35,19 +36,20 @@ export function voucherPlusPaidByCard() {
 }
 
 export function guardianWeeklyPaidByCard() {
-	return new ProductBuilder(baseGuardianWeekly())
+	return new ProductBuilder(baseGuardianWeeklyGift())
```
enesihsa: I'll return this back to how it was since it is for paid subscriptions and not gift subscriptions. Looking into the Cypress errors at the moment.

shared/productResponse.ts
```diff
@@ -82,6 +82,7 @@ const productKeys = [
 	'Newspaper Digital Voucher',
 	'Newspaper Digital Voucher + Digital',
 	'Contributor',
+	'Guardian Weekly - Gift',
```
johnduffell: the list/enum is all the possible values for the mma key returned from MDAPI.

At the moment, this is the product name in zuora (e.g. `Guardian Weekly - Domestic`) although with a couple of special cases e.g. observer where a migration was impractical.
https://github.com/guardian/members-data-api/blob/39a7b714627fe0c245b761f61d49d3a3399c9997/membership-attribute-service/app/models/AccountDetails.scala#L224-L230

I think we should try to keep the special cases to a minimum.


We do store the reader type in zuora, e.g. `Direct` or `Gift` and return that via MDAPI.


I think that manage should look at the reader type field to decide what copy to display.  It must already be looking at that somewhere in order to display the "Gift" flash at the top right of the card, so can we extend that condition to the copy as well?


Happy to discuss it when we chat tomorrow.

johnduffell:
small typo in the PR description - should be gifter or gift purchaser.

> Guardian Weekly Gift benefits shown to the giftee currently include the weekly print delivery too, which is incorrect.

should be

> Guardian Weekly Gift benefits shown to the gift purchaser currently include the weekly print delivery too, which is incorrect.

johnduffell (approved):
nice one thanks so much for persevering on this one, the end result looks great 💯  and will avoid any further confusion from readers 👍

PR #1664, opened 2026-06-05
Add Extra accounts page and invitation management UI

j-ruda-guardian:
### Summary

Adds an Extra accounts MMA page for Digital Plus subscribers (gated by ?TEST_EXTRA_ACCOUNTS_FLAG=true for now)
Adds nav entry, account overview banner, and useExtraAccounts hook + Zustand store
Implements send, cancel, and remove-access flows with confirmation modals and success/error feedback
Moves Modal to client/components/shared with containerCssOverrides and hideCloseButton

### Notes
API calls are mocked (2s delay); list refetch uses hardcoded data until backend is ready

###Test plan
- Enable with ?TEST_EXTRA_ACCOUNTS_FLAG=true on a Digital Plus test account
- Send invitation from an empty slot (form validation, loading, success message)
- Cancel a pending invitation via modal
- Remove access from an active account via modal
- Confirm holiday modals still work after the shared Modal move


client/components/mma/extraAccounts/ExtraAccountRow.tsx
```diff
@@ -0,0 +1,220 @@
...
+						<Button
+							priority="subdued"
+							size="small"
+							cssOverrides={sendInvitationCss}
+						>
+							Cancel invitation
+						</Button>
+					}
+				>
+					Lorem Ipsum
```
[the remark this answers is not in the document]
j-ruda-guardian: Irrelevant in this phase

client/components/mma/extraAccounts/ExtraAccountRow.tsx
```diff
@@ -0,0 +1,220 @@
...
+						<Button
+							priority="subdued"
+							size="small"
+							cssOverrides={sendInvitationCss}
+						>
+							Remove access
+						</Button>
+					}
+				>
+					Lorem Ipsum
```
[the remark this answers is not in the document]
j-ruda-guardian: Irrelevant in this phase

client/utilities/hooks/useExtraAccounts.ts
```diff
@@ -0,0 +1,246 @@
...
+		if (isAccountLoading || !mdapiResponse) {
+			return;
+		}
+
+		hasStartedLoading.current = true;
+
+		if (!hasDigitalPlus(mdapiResponse)) {
+			setShouldRedirect(true);
+			return;
+		}
```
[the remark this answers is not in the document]
j-ruda-guardian: Won't work since the query param is not present is the extra account page. Also, we don't need to guard this page, it's ok since the links won't be visible anyways and there are no security concerns

client/utilities/hooks/useExtraAccounts.ts
```diff
@@ -0,0 +1,246 @@
+import * as Sentry from '@sentry/browser';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import type { ExtraAccount } from '../../stores/ExtraAccountsStore';
+import {
+	ExtraAccountsLoadingState,
+	useExtraAccountsStore,
+} from '../../stores/ExtraAccountsStore';
+import { trackEvent } from '../analytics';
+import { hasDigitalPlus } from '../extraAccounts';
+import { useAccountDataLoader } from './useAccountDataLoader';
```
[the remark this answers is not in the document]
j-ruda-guardian: We are not checking for the flag in the direct route. The flag works only for hiding the buttons and navigation options in the main Account Overview. We don't need it inside the hook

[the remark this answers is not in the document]
j-ruda-guardian: At least for now, since there is no APIs being called inside of this new page

client/components/shared/nav/NavConfig.tsx
```diff
@@ -51,6 +53,12 @@ export const NAV_LINKS: NavLinks = {
 		local: true,
 		icon: AccountOverviewIcon,
 	},
+	extraAccounts: {
+		title: 'Extra accounts',
+		link: '/extra-accounts',
+		local: true,
```
enesihsa, line 59: What exactly does local mean here?

client/components/shared/Faqs.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+];
+
+export default {
+	title: 'Components/Faqs',
+	component: Faqs,
+	decorators: [ReactRouterDecorator],
+	parameters: {
+		layout: 'padded',
+	},
+} as Meta<typeof Faqs>;
```
enesihsa, line 39: What does this Meta export do?

client/stores/ToastStore.ts
```diff
@@ -0,0 +1,76 @@
+import type { ReactNode } from 'react';
+import { create } from 'zustand';
+import { devtools } from 'zustand/middleware';
+
+export type ToastSeverity = 'success' | 'error' | 'info';
+
+export interface Toast {
+	message: ReactNode | string;
+	severity: ToastSeverity;
```
enesihsa, line 9: Is this commonly called 'severity'? it made me think of errors initially only realised the different types after seeing the type above.

client/components/mma/extraAccounts/ExtraAccountInviteForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+				</CheckboxGroup>
+			</div>
+
+			<div css={actionsCss}>
+				<Button
+					priority="primary"
+					size="small"
+					isLoading={isSubmitting}
+					disabled={isSubmitting}
+					onClick={() => void handleSend()}
```
enesihsa: Is the `() => void` before `handleSend()` necessary? Can we not just have `handleSend()`?

client/components/shared/nav/NavConfig.tsx
j-ruda-guardian: local means this is an inapp MMA route, so the nav uses a React Router Link instead of a normal <a> that leaves MMA. Same as Account overview, Billing, etc.

client/components/shared/Faqs.stories.tsx
j-ruda-guardian: That's Storybook's default export, the story metadata like title, component, decorators. as Meta<typeof Faqs> is just the type. Same pattern as our other stories.

client/stores/ToastStore.ts
j-ruda-guardian: yeah, it's a common name in toast or alert libs like MUI. Happy to rename to tone or variant if you prefer, not precious about it.

client/utilities/hooks/useExtraAccounts.ts
```diff
@@ -0,0 +1,390 @@
...
+	ExtraAccountsLoadingState,
+	useExtraAccountsStore,
+} from '../../stores/ExtraAccountsStore';
+import { useToastStore } from '../../stores/ToastStore';
+import { trackEvent } from '../analytics';
+import { getDigitalPlusProduct } from '../extraAccounts';
+import { fetchWithDefaultParameters } from '../fetch';
+import { useAccountDataLoader } from './useAccountDataLoader';
+
+const MAX_ACCOUNTS = 3;
```
enesihsa: Could this not be a shared const? it's also defined in ExtraAccounts.ts

client/components/mma/extraAccounts/ExtraAccountInviteForm.tsx
j-ruda-guardian: Not strictly necessary here. void is just to explicitly ignore the Promise from the async handler. ESLint isn't enforcing floating promises in this repo, so I'll switch it.

enesihsa (approved):
Looks good to me!

client/stores/ToastStore.ts
enesihsa: Don't have a preference too, haven't worked much with toasts before so wasn't sure if it was a common pattern or something leftover from a previous iteration or so.

PR #1671, opened 2026-06-10
Created a template for the secondary user product card in account overview

enesihsa:
This PR creates the necessary product card sections and benefits copies for the secondary user account benefits as part of the multiple accounts initiative. 

### What does this PR change?
- Creates a new product card section for secondary user subscriptions. 
  - I assumed primary users will get their product data from MDAPI since their subscription is registered in Zuora. 
- `SecondaryAccountCard.tsx` to render the shared subscription product card on account overview without depending on a ProductDetail response since the endpoint for secondary accounts will not return a detailed subscription data. 
- Storybook card for visual testing. 
- Connection setup to '/api/me/secondary-user' endpoint and guarded DEV and CODE test for a hard-coded mock response from the API. This is still TBC. 
- New benefits text configuration to be used by the new card. 

### Next steps/further info
1. See https://github.com/guardian/manage-frontend/pull/1700 for the leave subscription button. 
2. For the product card we need the primary user's name, and once the secondary user clicks 'Leave Subscription' we need name and email address. The current implementation assumes a single API call which returns both. 
3. Endpoints and data format of the response still to be completed by the backend. 
  a. There is a guard for testing in CODE and DEV with a url parameter. 
4. Primary user benefits will change for the digital plus (addition of account sharing), this is to be removed from secondary accounts (benefitsOverride can be used for this).


client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -332,212 +342,302 @@ export const ProductCard = ({
...
+										</Button>
 									</div>
-								)}
-								{((isGifted && !userIsGifter) ||
-									!productDetail.subscription.autoRenew) && (
+								</div>
+							</Card.Section>
+						</>
+					)}
+				{!isSecondary && (
```
enesihsa: This also causes many indentation changes.

client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -305,9 +313,11 @@ export const ProductCard = ({
...
 						<p css={benefitsTextCss}>
-							{cardConfig.getBenefitsSectionCopy(
-								nextPaymentDetails,
-							)}
+							{isSecondary && primarySubscriber // TODO should probably find a better way to do this too.
+								? `You're enjoying ${mainPlan.name} as part of a shared subscription.`
+								: cardConfig.getBenefitsSectionCopy(
+										nextPaymentDetails,
+								  )}
 						</p>
```
j-ruda-guardian: It's possible that nextPaymentDetails will not exist, so 
`cardConfig.getBenefitsSectionCopy && nextPaymentDetails`
Will always test false. You could do something like:
`cardConfig.getBenefitsSectionCopy && (nextPaymentDetails || isSecondary) && (`
And then 
`isSecondary ? "you are enjoying copy" : cardConfig.getBenefitsSectionCopy()`

client/components/mma/accountoverview/ProductCard.tsx
```diff
@@ -332,212 +342,302 @@ export const ProductCard = ({
...
-									<dd data-qm-masking="blocklist">
-										{
-											productDetail.subscription
-												.subscriptionId
-										}
-									</dd>
-								</div>
-								{groupedProductType.tierLabel && (
+				{isSecondary &&
+					primarySubscriber && ( // TODO Separate more properly.
```
j-ruda-guardian: out of curiosity, why is the primarySubscriber needed?

client/components/mma/accountoverview/ProductCard.tsx
j-ruda-guardian: !isSecondary just renders what we had before right?

client/components/mma/accountoverview/ProductCard.tsx
enesihsa: The product card will show the details (full name and email) of the associated primary user according to the figma models. In practice they preferably should come together (ie if the user is secondary then I assume backend should always return the primary user's details too?) 

I should combine the primarySubscriber into determining isSecondary value probably unless if there's a scenario where we may have one but not the other.

j-ruda-guardian (commented):
I think this looks good. May I ask you for a change that will make reviewing this files a lot easier:
Please, take the new secondary user card and extract it to a new file (This ProductCard file is already more than crowded enough) and then import it and render it conditionally like you are doing. That might solve this indentation nightmare and will also make the whole thing easier to review! 

nice work, thank you!

client/components/mma/accountoverview/ProductCard.tsx
j-ruda-guardian: Ah ok, primarySubscriber is the owner of the original plan. That makes sense 👍 Oka, yeah you can put them together somewhere, as you are doing this check a couple of times across the file

client/components/mma/accountoverview/ProductCard.tsx
enesihsa: Ah thanks, missed this one. Another question I had was would it be possible to have the secondary user details in the getBenefitsSectionCopy too (or a more generic way of fetching them), currently it has one for each Zuora subscription type I think, but there are exceptions now for secondary users and GW Gift subscriptions.

client/components/mma/accountoverview/ProductCard.tsx
enesihsa: Yes, I think by default we show billing information and payment method since all other plans require some form of payment, but this is the only one that doesn't show these.

client/fixtures/productBuilder/baseProducts.ts
```diff
@@ -1260,3 +1276,14 @@ export function baseTierThree(): ProductDetail {
 		isTestUser: false,
 	};
 }
+
+export function baseSecondaryUser(): MultipleAccountsApiResponse {
```
enesihsa: We also don't have the data atm (update 21.07.2026 received necessary endpoint specification)

client/components/mma/accountoverview/ProductCardSections.tsx
```diff
@@ -93,21 +97,25 @@ export const BenefitsCopyAndToggle = ({
...
 	cardConfig.getBenefitsSectionCopy &&
-	nextPaymentDetails && (
+	(nextPaymentDetails || overrideBenefitsText) && (
 		<Card.Section
 			backgroundColor={benefitsSectionBackgroundColour}
 			removeBorders
 		>
 			<p css={benefitsTextCss}>
-				{cardConfig.getBenefitsSectionCopy(nextPaymentDetails)}
+				{nextPaymentDetails
```
enesihsa, line 115: This has to be done because secondary account benefits copy text cannot take in a nextPaymentDetails since it is not used for anything in the text. All other subscription product cards show some information about pricing and next payment dates.

client/components/mma/accountoverview/ProductCardConfiguration.ts
```diff
@@ -46,6 +46,9 @@ const guardianWeeklyGiftBenefitsCopy = (npd: NextPaymentDetails) =>
 const guardianWeeklyBenefitsCopy = (npd: NextPaymentDetails) =>
 	`You're subscribed to The Guardian Weekly for ${npd.currentPriceValue} per ${npd.paymentInterval} and receive a curated news magazine featuring our best global journalism in print, as well as unlimited access to our full suite of digital benefits.`;
 
+export const secondaryUserBenefitsCopy = () =>
+	`You're enjoying Digital plus as part of a shared subscription.`;
```
enesihsa, line 50: Is it ok for this to stay as hard-coded Digital plus at the moment? Currently there is not way to pass the product type to the product card config.

client/components/mma/accountoverview/ProductCardSections.tsx
```diff
@@ -135,6 +143,42 @@ export const GuardianAdLiteCopy = ({
...
+						are private and not shared with the subscription owner.
+					</p>
+				</div>
+				<div css={centeredActionCss}>
+					<Button
+						aria-label={`${subscriptionName} : Leave shared subscription`}
+						size="small"
+						priority="tertiary"
+						cssOverrides={sharedMembershipLeaveButtonCss}
+						onClick={() => undefined}
```
enesihsa: The secondary user leave journey has its own PR: https://github.com/guardian/manage-frontend/pull/1700

client/components/mma/accountoverview/SecondaryAccountCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+	SecondaryUserSubscriptionDetails,
+} from './ProductCardSections';
+
+export const SecondaryAccountProductCard = ({
+	primaryUser,
+}: {
+	primaryUser: MultipleAccountPrimaryUser;
+}) => {
+	const specificProductType =
+		getSpecificProductTypeFromProductKey('Digital Pack');
```
enesihsa, line 20: ~~Currently the MultipleAccountAPI does not return any subscription/product data so I hard-coded Digital Pack as the default product.~~ This has changed and subscription ID is not returned, as far as I know it cannot be used to infer the specific product without an extra API key. This makes sense currently as we only have digital pack shared subscription.

server/apiGatewayDiscovery.ts
```diff
@@ -256,6 +258,38 @@ const invoicingAPIGateway = getApiGateway(
 );
 export const invoicingAPI = invoicingAPIGateway.authorisedExpressCallback;
 
+const multipleAccountAPIGateway = getApiGateway(
+	'support',
+	'multiple-account-api',
+);
+export const multipleAccountAPI =
+	// multipleAccountAPIGateway.authorisedExpressCallback;
```
enesihsa, line 266: Ideally this should be used instead of the authorisation below.

server/routes/api.ts
```diff
@@ -104,6 +105,23 @@
...
+// 				lastName: 'Piri',
+// 				email: 'pepe.piri@chicken.com',
+// 			},
+// 		],
+// 	},
+// };
+
+router.get(
+	'/secondary-user/me',
+	multipleAccountAPI('secondary-user/me', 'MULTIPLE_ACCOUNTS', []),
```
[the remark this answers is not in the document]
j-ruda-guardian: Ok I was looking into this. Seems like there's been other routes with the same warning and we don't have an approach for the whole project. We can dismiss this, and even if we wanted to fix it, it's definitely out of the scope of this PR @enesihsa

client/utilities/hooks/useAccountDataLoader.ts
```diff
@@ -55,6 +58,14 @@ async function fetchAllAccountData() {
...
 			)
 			.catch((): null => null),
+		secondaryAccountFetcher()
+			.then(
+				(r) =>
+					JsonResponseHandler(
+						r,
+					) as Promise<MultipleAccountApiResponse>,
+			)
+			.catch((): null => null),
```
j-ruda-guardian, line 68: So if this call fails, the expression on the AccountOverview 
`secondaryAccountDetails?.subscriptions.length === 0`
Will be false and users will not see EmptyAccountOverview

j-ruda-guardian: I believe this one is important to fix

client/components/mma/accountoverview/AccountOverview.tsx
```diff
@@ -445,6 +451,19 @@ const AccountOverviewPage = ({ isFromApp }: IsFromAppProps) => {
 				/>
 			)}
 			{possiblyAffectedByCanadaPostStrike && <CanadaStrike />}
+			{secondaryAccountDetails?.subscriptions &&
+				secondaryAccountDetails.subscriptions.length > 0 && (
+					<Fragment>
```
j-ruda-guardian: You are missing a key here

client/components/mma/accountoverview/SecondaryAccountCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+	const mainPlan = {
+		name: specificProductType.friendlyName,
+		shouldBeVisible: true,
+	};
+	if (!mainPlan) {
+		throw new Error('mainPlan does not exist in SecondaryAccountCard');
+	}
+	const productTitle =
```
j-ruda-guardian: I think this is deadcode isn't it?

server/apiGatewayDiscovery.ts
j-ruda-guardian: What do you think about this? Do you want to explore this in this PR? WE still have time to figure this out

client/components/mma/accountoverview/ProductCardSections.tsx
j-ruda-guardian: This is ok. Maybe disable this button in this PR for the sake of completion, so it doesn't get merged in a "broken" state

j-ruda-guardian (commented):
This PR looks very good so far! I left a couple of suggestions and one comment I think is important

server/apiGatewayDiscovery.ts
enesihsa: Initially I thought it could pose a security if users could pass a different/custom token, but it actually doesn't seem that problematic, we can address this in a future PR too if it is safe to merge as it is now.

client/utilities/hooks/useAccountDataLoader.ts
enesihsa: Good catch, thanks, addressed it by changing the guard to `(!secondaryAccountDetails || secondaryAccountDetails.subscriptions.length === 0)` which should return true if the secondary accounts endpoint fails to return anything or there are no secondary subscriptions.

client/components/mma/accountoverview/SecondaryAccountCard.tsx
enesihsa: Yes, I think this is from when I was trying to fetch the mainPlan with a passed parameter. Removed it now.

client/components/mma/accountoverview/AccountOverview.tsx
```diff
@@ -445,6 +452,19 @@ const AccountOverviewPage = ({ isFromApp }: IsFromAppProps) => {
...
+						<h2 css={subHeadingCss}>Shared with you</h2>
+						{secondaryAccountDetails.subscriptions.map(
+							(subscription) => (
+								<SecondaryAccountProductCard
+									subscription={subscription}
+								/>
+							),
+						)}
+					</Fragment>
+				)}
```
j-ruda-guardian, line 468: So I think you moved the map inside of the fragment. This fragment might not be needed anymore. But the immediate element after the map needs a key. Something like
```
<SecondaryAccountProductCard
    subscription={subscription}
    key={subscription.subscriptionName}
/>
```

j-ruda-guardian (approved):
LGTM! Nice job :)

PR #1692, opened 2026-06-24
resolve ws vulnerability

shtukas:
Here we resolve a vulnerability with ws, reported here: https://github.com/guardian/manage-frontend/security/dependabot/354

Note that we also needed to
- Ensure slint-visitor-keys: ^4.0.0
- Correct a pre commit linting check in server/apiProxy.ts

### DIFF
--- a/server/apiProxy.ts
+++ b/server/apiProxy.ts
@@ -57,7 +57,7 @@ export const proxyApiHandler =
 				(evolvingPath: string, urlParamName: string) =>
 					evolvingPath.replace(
 						':' + urlParamName,
-						req.params[urlParamName] || '',
+						String(req.params[urlParamName] || ''),
 					),
 				path,
 			)
--- a/package.json
+++ b/package.json
@@ -188,7 +188,9 @@
     "flatted": "^3.4.2",
     "follow-redirects": "^1.16.0",
     "serialize-javascript": "^7.0.3",
-    "tmp": "^0.2.7"
+    "tmp": "^0.2.7",
+    "ws": "^8.21.0",
+    "eslint-visitor-keys": "^4.0.0"
   },
   "license": "UNLICENSED",
   "msw": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,19 +2,14 @@
 # yarn lockfile v1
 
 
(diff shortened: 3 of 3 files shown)


enesihsa (approved):

server/apiProxy.ts
```diff
@@ -57,7 +57,7 @@ export const proxyApiHandler =
 				(evolvingPath: string, urlParamName: string) =>
 					evolvingPath.replace(
 						':' + urlParamName,
-						req.params[urlParamName] || '',
+						String(req.params[urlParamName] || ''),
```
enesihsa, line 60: Was this due to some tsc linting errors? I think I got this a few times too.

shtukas: Yep! It was

```
server/apiProxy.ts:60:7 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string | string[]' is not assignable to parameter of type '(substring: string, ...args: any[]) => string'.
      Type 'string' is not assignable to type '(substring: string, ...args: any[]) => string'.

60       req.params[urlParamName] || '',
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/typescript/lib/lib.es5.d.ts:470:5
    470     replace(searchValue: string | RegExp, replacer: (substring: string, ...args: any[]) => string): string;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.


Found 1 error in server/apiProxy.ts:60
```

johnduffell:
looking at the lock file, this PR seems to have switched from npmjs.org to yarnpkg.org, what does that mean?

Also, there seems to be a new sentry issue that I don't fully understand, a handful of responses to /me/mma are coming back falsy, but it's not clear why only a small portion and how this PR could have affected it
https://the-guardian.sentry.io/issues/7574364801/?environment=theguardian.com&project=1208603&query=is%3Aunresolved&referrer=issue-stream&sort=freq

PR #1694, opened 2026-06-25
ensure aws-cdk-lib: 2.246.0

shtukas:
ensure aws-cdk-lib: 2.246.0 to address a vulnerability presented here: https://github.com/guardian/manage-frontend/security/dependabot/352

Note that other dependencies, as well as the snapshot, needed to be updated to prevent build issues.

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -19,7 +19,7 @@
     "@types/node": "18.7.14",
     "aws-cdk": "2.178.2",
     "aws-cdk-lib": "2.246.0",
-    "constructs": "10.4.2",
+    "constructs": "10.6.0",
     "eslint": "9.15.0",
     "jest": "^27.5.1",
     "prettier": "^2.7.1",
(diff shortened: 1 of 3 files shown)


cdk/package.json
```diff
@@ -15,25 +15,29 @@
...
+    "ts-node": "^10.9.2",
+    "typescript": "~5.3.3"
   },
   "prettier": "@guardian/prettier",
   "resolutions": {
-    "picomatch": "2.3.2",
     "flatted": "^3.4.2",
     "lodash": "^4.18.0",
-    "minimatch": "^9.0.7"
+    "minimatch": "^5.1.8",
```
enesihsa: Would this not revert the previous minimatch fix?

enesihsa (approved):

cdk/package.json
shtukas: No. The two PRs are independent. Merging from main will take care of it

PR #1697, opened 2026-06-29
Revert "resolve ws vulnerability"

shtukas:
Reverts guardian/manage-frontend#1692 (Looks like the ws update is causing a race condition. reverting to see if we recover the standard behaviour)


johnduffell (approved):
thanks let's see if that stops the errors

PR #1699, opened 2026-07-01
resolve ws vulnerability

shtukas:
This is another attempt at resolving a ws vulnerability: https://github.com/guardian/manage-frontend/security/dependabot/354 (first attempt was here: https://github.com/guardian/manage-frontend/pull/1692)

Note that the first attempt wasn't well done become overriding a version on resolutions the way it was done, in the package.json overrides all versions everywhere and may by accident upgrade major versions. 

In this attempt we used an idea from @j-ruda-guardian by which we simply remove from the yarn.lock like the ws 7.x.x and 8.x.x entries and let `yarn install` upgrade from `7.5.10` to `7.5.11` and from `8.18.0` to `8.21.0` which are both patched. This ensures that the version update to a patched version is done, but without overriding the major version.


j-ruda-guardian (approved):

johnduffell (commented):
much clearer and more focused, hope that works better this time 👍

PR #1702, opened 2026-07-06
Add discounted Annual S+->D+ switch + Fix upgrade product payment copy (post #1642)

j-ruda-guardian:
## Summary
Follow-up to #1642 that centralises upgrade payment/discount copy, fixes several post-merge bugs, and adds annual discount support in the preview API.
### Shared payment copy utility
- Adds `client/utilities/upgradeProductPaymentCopy.ts` with shared helpers for discount detection, currency formatting, date labels, and copy for Information / Confirmation / Thank You
- Refactors `UpgradeProductInformation`, `UpgradeProductConfirmation`, and `UpgradeProductThankYou` to use the shared builders instead of inline string assembly
### Discount detection & API
- Fixes `isDiscountedOffer` to use `discountedPrice < targetCatalogPrice` (avoids false positives when discounted price equals catalog price, e.g. 100% off edge cases)
- Enables discount preview/switch for **annual** subscriptions (removes monthly-only `discountSwitchEnabled` gating in upgrade hooks; defaults `discountSwitchEnabled` to `true` in `productUtils`)
### Copy & formatting fixes
- Uses `formatAmount` consistently for currency display (e.g. `£7.50` not `£7.5`)
- Uses `preview.nextPaymentDate` for payment copy instead of `subscription.nextPaymentDate`
- Correct promo duration pluralisation (`1 month` not `1 months`); missing `upToPeriods` falls back to `"unknown"` not `0`
- Payment paragraphs use `upToPeriods - 1` for remaining promo periods (first discounted period counted separately)
### Annual-specific copy
- **Non-discounted annual:** unchanged wording — e.g. `After this, from January 15th, your yearly payment will be £149`
- **Discounted annual:** new wording — e.g. `After this, from 06 July 2027, your payment will be £200 every year. Your next payment date will be the 6th of July.`
  - Promo duration is **not** repeated in the payment paragraph (shown in the card header instead)
  - Thank You uses matching `every year` copy and `the 6th of July` date label
- **Monthly discount** copy unchanged (promo duration still in payment paragraph)
### Safety guards
- Confirmation/Information require `previewResponse` before render
- Confirmation returns early if `nextPaymentDetails` is missing (avoids undefined payment interval)
### Storybook & tests
- Adds annual discount stories: Information / Confirmation / Thank You
- Adds 20 unit tests in `client/__tests__/upgradeProductPaymentCopy.test.ts`

## Test plan
- [x] `yarn jest client/__tests__/upgradeProductPaymentCopy.test.ts --verbose`
- [x] `yarn type-check`
- [X] Spot-check Storybook: `Information - Monthly Discount`, `Confirmation - Monthly Discount`, `Thank You - Monthly Discount` (expect `£7.50`, "3 months" in header, "2 months" in payment copy)
- [X] Manual smoke test discounted upgrade flow on CODE

## Notes
- Addresses unresolved review threads from #1642 (banner WIP threads intentionally skipped)

### DIFF
--- a/client/components/mma/upgradeProduct/UpgradeProductConfirmation.tsx
+++ b/client/components/mma/upgradeProduct/UpgradeProductConfirmation.tsx
@@ -140,7 +140,11 @@ export const UpgradeProductConfirmation = () => {
 		false,
 	);
 
-	const paymentInterval = nextPaymentDetails?.paymentInterval ?? 'month';
+	if (!nextPaymentDetails) {
+		return null;
+	}
+
+	const { paymentInterval } = nextPaymentDetails;
 
 	const discountConditionsText =
 		isDiscountedOffer && isDiscountedPreview(previewResponse)
@@ -239,9 +243,7 @@ export const UpgradeProductConfirmation = () => {
 								{discountConditionsText}
 							</span>
 						) : (
-							<span>
-								{'/' + nextPaymentDetails?.paymentInterval}
-							</span>
+							<span>{'/' + paymentInterval}</span>
 						)}
 					</p>
 				</Card.Section>
@@ -297,16 +299,16 @@ export const UpgradeProductConfirmation = () => {
 			<div css={termsAndConditionsContainerCss}>
 				<p css={termsAndConditionsTextCss}>
 					This subscription auto-renews. You'll be charged the
-					applicable {nextPaymentDetails?.paymentInterval}ly amount at
-					each renewal unless you cancel. You can cancel your
-					subscription at any time before your next renewal date. To
-					cancel go to <a href="/">Manage My Account</a>. Cancellation
-					will take effect at the end of your current{' '}
-					{nextPaymentDetails?.paymentInterval}ly payment period.
-					There is also a cooling off period of 14 days from sign-up.
-					You can cancel your subscription within 14 days of sign-up
-					by <a href="/help-centre#call-us">contacting us</a> and
-					receive a full refund.
+					applicable {paymentInterval}ly amount at each renewal unless
+					you cancel. You can cancel your subscription at any time
+					before your next renewal date. To cancel go to{' '}
+					<a href="/">Manage My Account</a>. Cancellation will take
+					effect at the end of your current {paymentInterval}ly
+					payment period. There is also a cooling off period of 14
+					days from sign-up. You can cancel your subscription within
+					14 days of sign-up by{' '}
+					<a href="/help-centre#call-us">contacting us</a> and receive
+					a full refund.
 				</p>
 			</div>
 
--- a/client/components/mma/upgradeProduct/UpgradeProductConfirmation.tsx
+++ b/client/components/mma/upgradeProduct/UpgradeProductConfirmation.tsx
@@ -35,7 +35,6 @@ import {
 	isDiscountedPreview,
 } from '@/client/utilities/upgradeProductPaymentCopy';
 import { formatAmount } from '@/client/utilities/utils';
-import { dateString } from '@/shared/dates';
 import { PRODUCT_TYPES } from '@/shared/productTypes';
 import { Pill } from '../../shared/Pill';
 import { productCardConfiguration } from '../accountoverview/ProductCardConfiguration';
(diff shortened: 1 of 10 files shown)


client/components/mma/upgradeProduct/UpgradeProductConfirmation.tsx
```diff
@@ -130,44 +140,23 @@ export const UpgradeProductConfirmation = () => {
...
-				: 0
-		} ${previewResponse?.discount?.upToPeriodsType.toLowerCase()} and then ${
-			mainPlan.currency
-		}${previewResponse?.targetCatalogPrice} per ${
-			nextPaymentDetails?.paymentInterval
-		}. The ${nextPaymentDateDayDiscounted} will be your next payment date.`;
-	} else {
-		paymentConditionsText += `After this, from ${nextPaymentDate}, your ${nextPaymentDetails?.paymentInterval}ly payment will be ${mainPlan.currency}${previewResponse?.targetCatalogPrice}`;
-	}
+	const paymentInterval = nextPaymentDetails?.paymentInterval ?? 'month';
```
johnduffell: why wouldn't there be a payment interval and why does that mean it's monthly?
it looks before we'd have inserted "undefined" `£12/undefined` or `your undefined payment will be`

client/utilities/productUtils.ts
```diff
@@ -255,7 +255,7 @@ export const changePlanFetch = ({
 	targetProduct,
 	preview = false,
 	newAmount,
-	discountSwitchEnabled,
+	discountSwitchEnabled = true,
```
johnduffell, line 258: this PR is to fix the copy and refactor, but it also seems to be to switch it on for Annual as well?

j-ruda-guardian: Yeah I know :) I will update the description and the title. Jim asked me to deploy the yearly and I figured it would be a good opportunity to check that these changes are looking good as well

client/utilities/upgradeProductPaymentCopy.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export function formatCurrency(currency: string, amount?: number) {
+	return `${currency}${formatAmount(amount)}`;
+}
+
+export function formatDiscountPeriodLabel(
+	upToPeriods: number | undefined,
+	upToPeriodsType: SwitchDiscountResponse['upToPeriodsType'],
+): string {
+	if (upToPeriods == null || upToPeriods <= 0) {
```
johnduffell, line 88: upToPeriods == null but 
upToPeriods: number | undefined
is that right? would a === have picked that put?

client/utilities/upgradeProductPaymentCopy.ts
```diff
@@ -0,0 +1,172 @@
...
+		preview.discount != null &&
+		preview.discount.discountedPrice < preview.targetCatalogPrice
+	);
+}
+
+/** Payment-paragraph periods: API total minus the first discounted period already in flight. */
+export function getRemainingDiscountPeriods(
+	upToPeriods?: number,
+): number | 'unknown' {
+	if (upToPeriods == null || upToPeriods <= 0) {
```
johnduffell, line 69: same as below, upToPeriods is optional but we're checking for null

client/utilities/upgradeProductPaymentCopy.ts
```diff
@@ -0,0 +1,172 @@
...
+		return 'unknown';
+	}
+	return `${upToPeriods} ${appendCorrectPluralisation(
+		upToPeriodsType,
+		upToPeriods,
+	).toLowerCase()}`;
+}
+
+function formatRemainingDiscountPeriodLabel(
```
johnduffell, line 103: might be useful to have a javadoc style comment above some of these just with an example "5 months" (or even work it into the function name if that works `formatRemainingDiscountPeriodXMonths`)
not sure how that would work for them all, especially it would be duplicating the tests.

client/utilities/upgradeProductPaymentCopy.ts
j-ruda-guardian: fair TypeScript-style question :) == null is a common idiom meaning “nullish”. This is deliberate.

j-ruda-guardian: Happy to switch to upToPeriods === undefined if we want the check to match the type exactly and we don’t care about runtime null from the API. What do you think @johnduffell

johnduffell: oh that's a new learning for me about TS types!  So if the server can return null we should add that to the type.  Although I thought from memory either the whole discount object is there, or no discount at all.  I can check if necessary.

client/components/mma/upgradeProduct/UpgradeProductThankYou.tsx
```diff
@@ -121,7 +118,9 @@ export const UpgradeProductThankYou = () => {
 					>
 						<b css={whatHappensNowItemInformationBoldTextCss}>
 							{isDiscountedOffer
-								? `Your new payment date will be the ${nextPaymentDateDay}.`
+								? mainPlan.billingPeriod === 'year'
+									? `Your new payment date will be ${nextPaymentDateDayLabel}.`
+									: `Your new payment date will be the ${nextPaymentDateDayLabel}.`
```
johnduffell, line 123: nit/simplification: put the `the` in the return value of `formatUpgradeNextPaymentDayLabel` to avoid rechecking the billing period?
(the logic is the same but we don't use `isYearlyBilling` function here)

johnduffell (approved):
nice one thanks for getting that all ready! 👏

PR #1703, opened 2026-07-07
read extraTaxApplies not taxExclusive so the tax notice shows

andrade0:
The Canada tax-exclusive notice wasn't showing. manage-frontend reads `productDetail.taxExclusive`, but members-data-api emits the field as `extraTaxApplies` (it was renamed in guardian/members-data-api#1251, off the back of the naming feedback on that PR). So `taxExclusive` is always undefined and the notice condition never fires.

This renames the `ProductDetail` field and its read sites to `extraTaxApplies` to match the API. The `TaxExclusiveNotice` component itself is unchanged, it just gets fed the right value now.

Spotted by Graham while testing the new CA plans. Original notice work was #1665 / #1666.

Verified locally: type-check, lint and the TaxExclusiveNotice tests pass.

### DIFF
--- a/client/components/mma/accountoverview/ManageProduct.tsx
+++ b/client/components/mma/accountoverview/ManageProduct.tsx
@@ -218,7 +218,7 @@ const InnerContent = ({
 				}
 			/>
 
-			<TaxExclusiveNotice taxExclusive={productDetail.taxExclusive} />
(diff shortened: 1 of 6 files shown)


johnduffell (approved):
thanks for fixing 👏

PR #819, opened 2022-05-30
[RB] - add a/b test setup to manage so that tests can be run

rBangay:
## What does this change?
Integrate the Guardian a/b testing system with manage:
https://github.com/guardian/ab-testing

This pr is just the infrastructure around setting up an a/b test in manage. This doesn't create an a/b test.

The wiki has also updated with instructions on how to setup a new test:
https://github.com/guardian/manage-frontend/wiki/A-B-testing

PR #1705, opened 2026-07-09
remove deprecated ab testing framework plus formatting

johnduffell:
We had a message

> We noticed you still have @guardian/ab-core installed in the manage-frontend repo, though it appears unused. 
> 
> Are you happy to remove the framework?
> 
> If you need to keep any A/B testing logic, you'll need to copy some code from DCR. Here are the relevant docs: [ab testing in DCR](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/development/ab-testing-in-dcr.md) and [fastly edge cache ab test](https://github.com/guardian/fastly-edge-cache/blob/main/theguardiancom/docs/ab-testing.md).
>
> ...a new A/B testing framework exists now. We're planning to update @guardian/ab-core to work with it in the future.

We are not using it, so we had a mobbing session today to remove it.  Special thanks to @j-ruda-guardian for taking me through the steps to remove it safely.

It was originally added in https://github.com/guardian/manage-frontend/pull/819

This PR removes the libraries and framework, and also does any prettier formatting needed.


client/components/mma/MMAPage.tsx
```diff
@@ -1,8 +1,5 @@
 import { css, Global } from '@emotion/react';
```
johnduffell, line 1: the main changes are in this file - the remainder code changes are just prettier formatting

j-ruda-guardian (approved):
LGTM :)

PR #1706, opened 2026-07-10
Hide change amount for tax-exclusive plans and move tax notice outside product card

j-ruda-guardian:
- Move "Taxes may apply to future payments" below the product card on account overview, to comply with design requirements, instead of inside it
- Hide the "Change amount" button on manage subscription when extraTaxApplies is true, since tax-exclusive pricing isn't supported in that flow yet

### DIFF
--- a/client/components/mma/shared/TaxExclusiveNotice.tsx
+++ b/client/components/mma/shared/TaxExclusiveNotice.tsx
@@ -8,7 +8,7 @@ interface TaxExclusiveNoticeProps {
 const noticeCss = css`
 	${textSans15};
 	color: ${palette.neutral[46]};
-	margin: ${space[2]}px 0 0;
+	margin: ${space[5]}px 0 0;
 `;
 
 export const TaxExclusiveNotice = ({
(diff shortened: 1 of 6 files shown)


shtukas (approved):

client/components/mma/shared/TaxExclusiveNotice.tsx
```diff
@@ -8,7 +8,7 @@ interface TaxExclusiveNoticeProps {
 const noticeCss = css`
 	${textSans15};
 	color: ${palette.neutral[46]};
-	margin: ${space[2]}px 0 0;
+	margin: ${space[5]}px 0 0;
 `;
```
graham228221, line 12: @j-ruda-guardian design have also asked if the colour can be changed on this notice across each page - it should be "neutral 7 (#121212)" - should that be just changed here?

graham228221 (commented):
Thanks for this - could we just update the colour as well on the notice?

client/components/mma/shared/TaxExclusiveNotice.tsx
j-ruda-guardian: Done!

PR #1709, opened 2026-07-20
Bumped websocket-driver to 0.7.5

enesihsa:
This PR aims to resolve the critical alert from dependabot on websocket-driver.

Link: https://github.com/guardian/manage-frontend/security/dependabot/363

### What does this PR change?
Adds a resolution to set websocket-driver to use 0.7.5 instead of 0.7.4 that is currently used as the default.


johnduffell (approved):
thanks!

PR #1710, opened 2026-07-20
add help centre domain cnames (temporary values for authentication)

johnduffell:
We would like to serve the salesforce experience version of the help centre via fastly.

The CODE version is here: https://help.code.dev-theguardian.com/

This requires quite a bit of setup, first we have to prove to salesforce that we own the domain by pointing  help.theguardian.com to a special value, then we have to change it to correctly point it to fastly.

Relevant docs:
https://help.salesforce.com/s/articleView?id=platform.domain_mgmt_enable_https.htm&type=5
https://help.salesforce.com/s/articleView?id=platform.custom_url_add.htm&type=5

## CODE only:

- Since the CODE CNAME record was created manually in NS1, we first need to "adopt" the record to CFN by creating it with the same details (done)

## CODE and PROD

- We need to CNAME the host name to a special (invalid) hostname, presumably to prove to salesforce that any incoming traffic with the host header of our customer facing domain is actually supposed to be picked up by our salesforce backend (and we are not stealing traffic from some unrelaited salesforce customer who is on the same tenant)
- Then we need to update it with the correct fastly name ( https://github.com/guardian/manage-frontend/pull/1713 )

This PR adds the Guardian DNS record entry needed to authenticate, (then we can do another PR to change it to the correct fastly hostname.)

## Testing:

Deployed to CODE and the record is available:
```
% dig +noall +answer CNAME help.code.dev-theguardian.com
help.code.dev-theguardian.com. 300 IN    CNAME    help.code.dev-theguardian.com.00d20000000nq5geaa.live.siteforce.com.
%
```
and the domain could be added in salesforce successfully.

Then the next PR https://github.com/guardian/manage-frontend/pull/1713 was also deployed to CODE, at which point the site started working

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,5 +58,14 @@ jobs:
             manage-frontend:
               - ./manage-frontend.zip
 
-
-
+      - name: Upload help centre domain to Riff-Raff
+        uses: guardian/actions-riff-raff@v4
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          roleArn: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+          app: help-centre
+          configPath: ./help-centre/riff-raff.yaml
+          contentDirectories: |
+            cfn:
+              - ./cdk/cdk.out/HelpCentre-CODE.template.json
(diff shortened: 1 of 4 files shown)


rafaelbarreto-theguardian (approved):
LGTM

PR #1711, opened 2026-07-21
Resolved js-yaml to ^4.3.0

enesihsa:
This PR addresses the dependabot alerts https://github.com/guardian/manage-frontend/security/dependabot/367 and https://github.com/guardian/manage-frontend/security/dependabot/366. 

### What does this PR change?
Resolves js-yaml to ^4.3.0 in package.json resolutions for transitive dependencies.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4935,13 +4935,6 @@ archiver@^3.1.1:
     tar-stream "^2.1.0"
     zip-stream "^2.1.2"
 
-argparse@^1.0.7:
-  version "1.0.10"
-  resolved "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz"
-  integrity sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==
-  dependencies:
-    sprintf-js "~1.0.2"
-
 argparse@^2.0.1:
   version "2.0.1"
   resolved "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -10229,18 +10222,10 @@ js-cookie@3.0.7:
...
-  dependencies:
-    argparse "^1.0.7"
-    esprima "^4.0.0"
-
-js-yaml@^4.1.0:
-  version "4.1.0"
-  resolved "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz"
-  integrity sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==
+js-yaml@^3.13.1, js-yaml@^4.1.0, js-yaml@^4.3.0:
+  version "4.3.0"
```
johnduffell, line 10226: is it definitely ok to bump a major version? the vuln report recommends 3.15.0 for v3 and 4.3.0 for v4

johnduffell: there's a fair bit in the migration guide, so I don't think it's a drop in replacement https://github.com/nodeca/js-yaml/blob/master/docs/migrate_v3_to_v4.md

enesihsa: Will look into it, in the alert I couldn't see any that depended on this version so not sure where this is coming from. This was what it showed so I thought it'd be fine to bump it all to ^4.3.0. 

```@eslint/eslintrc@3.2.0 requires js-yaml@^4.1.0```
```eslint@9.15.0 requires js-yaml@^4.1.0 via a transitive dependency on @eslint/eslintrc@3.2.0```

johnduffell: I was looking at the one you pasted first in the description - it's possible it's only a dev dependency though? https://github.com/guardian/manage-frontend/security/dependabot/367

enesihsa: Ah, my mistake I was looking at https://github.com/guardian/manage-frontend/security/dependabot/366 (366, not 367) why do they show as two separate issues when the vulnerability is the same? 😭

johnduffell: yeah it is odd, normally they put both major version ranges in the same link, rahter than two almost identical links for each supported version

PR #1719, opened 2026-07-22
Regenerated yarn.lock file

enesihsa:


server/apiProxy.ts
```diff
@@ -57,7 +57,9 @@ export const proxyApiHandler =
 				(evolvingPath: string, urlParamName: string) =>
 					evolvingPath.replace(
 						':' + urlParamName,
-						req.params[urlParamName] || '',
+						Array.isArray(req.params[urlParamName])
```
enesihsa, line 60: Had to do this to make the pre-commit checks happy

PR #1721, opened 2026-07-24
Handling array types in api call urls

enesihsa:
Started receiving this error from TypeScript pre-checks a few days ago while working on other features (address change blocks for MMA and secondary account product card). Current implementation does not account for array types in the API call urls, they default to falsy and return as ''. 

```
$ tsc --skipLibCheck
server/apiProxy.ts:60:7 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string | string[]' is not assignable to parameter of type '(substring: string, ...args: any[]) => string'.
      Type 'string' is not assignable to type '(substring: string, ...args: any[]) => string'.

60       req.params[urlParamName] || '',
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/typescript/lib/lib.es5.d.ts:470:5
    470     replace(searchValue: string | RegExp, replacer: (substring: string, ...args: any[]) => string): string;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.


Found 1 error in server/apiProxy.ts:60
```

### What does this PR change?
If the parameter passed through the url is an array of strings, this code should join them with '/', otherwise it fetches the parameter or return an empty string for this parameter if it is not defined. 

"example/:path" with params: { path: ['a', 'b', 'c'] } would call "example/a/b/c" endpoint instead of "example/"

### DIFF
--- a/server/__tests__/apiProxy.test.ts
+++ b/server/__tests__/apiProxy.test.ts
@@ -1,0 +1,99 @@
+/**
+ * @jest-environment node
+ */
+
+import type { Request, Response } from 'express';
+import fetch from 'node-fetch';
+import { proxyApiHandler, straightThroughBodyHandler } from '../apiProxy';
+
+jest.mock('node-fetch');
+jest.mock('@/server/log');
+jest.mock('@/server/oktaConfig', () => ({
+	getConfig: () => ({ useOkta: true }),
+}));
+
+const mockedFetch = jest.mocked(fetch);
+
+const buildFetchResponse = () =>
+	({
+		status: 200,
+		ok: true,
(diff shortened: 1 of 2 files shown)


shtukas (approved):

enesihsa:
Continued work over here: https://github.com/guardian/manage-frontend/pull/1722

PR #1722, opened 2026-07-27
Casted req.params to a string to string dictionary

enesihsa:

This fixes a type imperfection in the current code that is failing the build. Express req.params can be `Rec<string, string>` or `Rec<string, string[]>`, but in our code we perform a reduce/replace which only works in the `Rec<string, string>`.

### What does this PR change?
We specify the exact type to make the compiler happy. This PR casts the req.params passed to the API to Record<string, string>. ParamsDictionary in Express. 

Added a test to clarify how reduce works.


enesihsa:
Remember to `yarn install` more regularly

PR #1726, opened 2026-07-27
Restrictions to address changes through MMA

enesihsa:
The users are given the option to change the delivery address associated with their subscriptions through the MMA. The newspaper deliveries are handled either as "home delivery", within M25 area in London; or as "national delivery", outside of London. Changing the delivery address within London is not problematic and can be handled through the MMA, but addresses outside London have certain issues that make it unfeasible for us to allow self-serve address change nationally within the UK. 

We ran an investigation on which address changes were possible through which subscriptions, and highlighted the address changes that are allowed and that should be blocked. More detailed description can be found here: https://docs.google.com/document/d/1DTBRBkPSfJm44DKHzSN5pBrKRF1xg6lIvdDYce_YQqc 

### What does this PR change?
This PR adds blocks to address changes out of the UK for voucher subscriptions and moves outside M25 home delivery area for subscribers who have a home delivery subscription. This was achieved by using the `productType` for checking user subscriptions rather than `friendlyName` of products. Both `national` and `home` delivery has `newspaper home delivery plus digital subscription` or `newspaper home delivery subscription` (print only) as their friendly name, thus distinguishing between the two in the previous implementation was not possible. 

### Next steps/further info
There are still some ways to accidentally change the address outside the allowed area. Checking out with a new address from support changes the address for the account entirely. Tacking this is out of the scope of this PR and would require more complicated changes in how we deal with deliveries and user addresses.

### How is this change tested?
The following address change scenarios were tested on CODE. 
- Home delivery subscribers can change their address within London, but outside of M25 and the UK addresses raise an error. 
- Voucher delivery address can be changed within the UK. 
- For multiple subscriptions, most restrictive one takes precedence. Users with voucher + home subscription can only change addresses within Greater London and users with national delivery get redirect to the phone line.


shtukas (approved):

client/__tests__/components/delivery/address/formValidation.test.ts
```diff
@@ -21,9 +21,13 @@ describe('update delivery address form validation', () => {
...
-	const guardianWeeklySubscription = ['Guardian Weekly subscription'];
-	const homeDeliverySubscription = ['newspaper home delivery subscription'];
-	const voucherSubscription = ['newspaper voucher subscription'];
+	const guardianWeeklySubscription = ['guardianweekly'];
+	const homeDeliverySubscription = [
+		'homedeliveryplusdigital',
+		'homedelivery',
+	];
+	const voucherSubscription = ['voucher', 'voucherplusdigital'];
+	const homeAndVoucherSubscription = ['homedeliveryplusdigital', 'voucher'];
```
graham228221, line 30: 🤔

graham228221: oh i see now, should have read the full test

graham228221 (approved):
I think this looks good, thanks!

j-ruda-guardian (commented):
There is one problem with this PR. The formValidation is a shared function that is also being used in DeliveryAddressStep. The strings expected on this functions have been changed, but DeliveryAddressStep still passes the old params. That will cause some inconsistencies in the expected validation logic

j-ruda-guardian (approved):
LGTM! Nice work :)

PR #1727, opened 2026-07-30
Express server Help Centre redirect

j-ruda-guardian:
The old Guardian help centre still lives under manage.theguardian.com/help-centre, but content has moved to the new help centre at help.theguardian.com. There isn’t a reliable one-to-one mapping from old articles to new ones, so traffic to the old paths needs to land on the new home page rather than deep-linked content.

This PR adds Express-layer permanent redirects so visits to /help-centre and its subpaths on manage go to https://help.theguardian.com/, while leaving contact-us (including nested form paths) and both diagnostic-information pages on manage for now, since those flows are still needed there.

PR #1729, opened 2026-08-03
add custom redirects for the most popular urls

johnduffell:
following on from https://github.com/guardian/manage-frontend/pull/1727

We decided to add custom redirects for the top few pages based on search referral.  This is because
1. it provides a better experience for users who already know what information they need
2. it will retain more of the search ranking from existing incoming links (as search engines are likely to see the new page as more relevant, and therefore transfer the ranking over to the new page)

This PR adds the relevant code to do the redirect.

## Notes

To work out the high ranked pages we did this in bigquery just this morning:
```
select p.path
from `datatech-platform-prod.online_traffic.fact_page_view` p
where 1=1
and p.received_date > date '2026-07-01'
and p.host = 'manage.theguardian.com'
and p.referrer_category = 'Search Engine and other Google'
group by 1
order by count(1) desc
```
And we picked the top 10 roughly, discarding the ones that were not relevant.
image

ref https://docs.google.com/spreadsheets/d/18ez1UbxfYmN9xKToOilKP0e-DBCCRsxG8O1e5oNUPWg/edit?gid=0#gid=0

Tested in CODE
- redirect works from a known url
- catch all redirect works
- diagnostic information works

### DIFF
--- a/server/helpCentreRedirect.ts
+++ b/server/helpCentreRedirect.ts
@@ -2,10 +2,8 @@ import { conf } from './config';
 
 export const helpCentreHomeForStage = (stage: string): string =>
 	stage === 'PROD'
-		? 'https://help.theguardian.com/'
-		: 'https://help.code.dev-theguardian.com/';
-
-export const NEW_HELP_CENTRE_HOME = helpCentreHomeForStage(conf.STAGE);
+		? 'https://help.theguardian.com'
+		: 'https://help.code.dev-theguardian.com';
 
 /**
  * Paths under the `/help-centre` Express mount that should continue to be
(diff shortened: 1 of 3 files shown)


server/helpCentreRedirect.ts
```diff
@@ -28,3 +26,28 @@ export const shouldRetainHelpCentrePath = (path: string): boolean => {
 		normalised === '/contact-us' || normalised.startsWith('/contact-us/')
 	);
 };
+
+// we want to keep one redirect per line for clarity:
+// prettier-ignore
```
johnduffell, line 31: I did this because it was inconsistently having one or two lines for each entry.  This only disables prettier for the one declaration

enesihsa (approved):

PR #1733, opened 2026-08-05
Fix Help Centre back links using domain, not STAGE

j-ruda-guardian:
conf.STAGE isn’t available in the client bundle, so Prod was falling through to the CODE Help Centre URL.

Back links now use window.guardian.domain (theguardian.com → Prod, otherwise CODE), matching how other client code picks environment.

### DIFF
--- a/client/utilities/helpCentreHome.ts
+++ b/client/utilities/helpCentreHome.ts
@@ -1,11 +1,4 @@
-import { conf } from '../../server/config';
 import { helpCentreHomeForDomain } from '../../shared/helpCentreConfig';
 
-export const getHelpCentreHomeUrl = (): string => {
-	const domain =
-		typeof window !== 'undefined' && window.guardian
-			? window.guardian.domain
-			: conf.DOMAIN;
-
-	return helpCentreHomeForDomain(domain);
-};
+export const getHelpCentreHomeUrl = (): string =>
+	helpCentreHomeForDomain(window.guardian?.domain ?? '');
--- a/client/utilities/helpCentreHome.ts
+++ b/client/utilities/helpCentreHome.ts
@@ -1,5 +1,11 @@
 import { conf } from '../../server/config';
-import { helpCentreHomeForStage } from '../../server/helpCentreRedirect';
+import { helpCentreHomeForDomain } from '../../shared/helpCentreConfig';
 
-export const getHelpCentreHomeUrl = (): string =>
-	helpCentreHomeForStage(conf.STAGE);
+export const getHelpCentreHomeUrl = (): string => {
+	const domain =
+		typeof window !== 'undefined' && window.guardian
+			? window.guardian.domain
+			: conf.DOMAIN;
+
+	return helpCentreHomeForDomain(domain);
+};
--- a/shared/__tests__/helpCentreConfig.test.ts
+++ b/shared/__tests__/helpCentreConfig.test.ts
@@ -1,0 +1,25 @@
+import { helpCentreHomeForDomain } from '../helpCentreConfig';
+
+describe('helpCentreHomeForDomain', () => {
+	test.each([
+		'theguardian.com',
+		'manage.theguardian.com',
+		'www.theguardian.com',
+	])('returns the prod help centre for %s', (domain) => {
+		expect(helpCentreHomeForDomain(domain)).toBe(
+			'https://help.theguardian.com',
+		);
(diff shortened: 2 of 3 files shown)


client/utilities/helpCentreHome.ts
```diff
@@ -1,5 +1,11 @@
...
-import { helpCentreHomeForStage } from '../../server/helpCentreRedirect';
+import { helpCentreHomeForDomain } from '../../shared/helpCentreConfig';
 
-export const getHelpCentreHomeUrl = (): string =>
-	helpCentreHomeForStage(conf.STAGE);
+export const getHelpCentreHomeUrl = (): string => {
+	const domain =
+		typeof window !== 'undefined' && window.guardian
+			? window.guardian.domain
+			: conf.DOMAIN;
```
johnduffell: conf is still from server, will it work? should it be in shared?

client/utilities/helpCentreHome.ts
```diff
@@ -1,5 +1,11 @@
 import { conf } from '../../server/config';
-import { helpCentreHomeForStage } from '../../server/helpCentreRedirect';
+import { helpCentreHomeForDomain } from '../../shared/helpCentreConfig';
 
-export const getHelpCentreHomeUrl = (): string =>
-	helpCentreHomeForStage(conf.STAGE);
+export const getHelpCentreHomeUrl = (): string => {
+	const domain =
+		typeof window !== 'undefined' && window.guardian
```
johnduffell: should we get a sentry alert if `window.guardian` doesn't exist?

shared/__tests__/helpCentreConfig.test.ts
```diff
@@ -0,0 +1,25 @@
+import { helpCentreHomeForDomain } from '../helpCentreConfig';
+
+describe('helpCentreHomeForDomain', () => {
+	test.each([
+		'theguardian.com',
+		'manage.theguardian.com',
+		'www.theguardian.com',
```
johnduffell: these tests are failing, but what does "domain" contain? the whole host, or the part after the first dot?

j-ruda-guardian: it's because of some early change! Fixing it now

client/utilities/helpCentreHome.ts
j-ruda-guardian: window.guardian is injected in the HTML head before the app boots. If it’s missing, lots of other things break  like identity, analytics.. A Sentry event from getHelpCentreHomeUrl would be a late symptom, not a useful signal. I dropped the conf part as well from the ternary because of that same reason

client/utilities/helpCentreHome.ts
```diff
@@ -1,5 +1,4 @@
-import { conf } from '../../server/config';
-import { helpCentreHomeForStage } from '../../server/helpCentreRedirect';
+import { helpCentreHomeForDomain } from '../../shared/helpCentreConfig';
 
 export const getHelpCentreHomeUrl = (): string =>
-	helpCentreHomeForStage(conf.STAGE);
+	helpCentreHomeForDomain(window.guardian?.domain ?? '');
```
johnduffell, line 4: so if there were to be an issue with `window`, we'd get links from PROD to CODE, but CODE would still link to CODE as expected.
Doesn't quite seem ideal - I'd rather CODE was wrong rather than PROD, but I'm happy to believe that it's super unlikely in practice.

shared/__tests__/helpCentreConfig.test.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('helpCentreHomeForDomain', () => {
+	test('returns the prod help centre for theguardian.com', () => {
+		expect(helpCentreHomeForDomain('theguardian.com')).toBe(
+			'https://help.theguardian.com',
+		);
+	});
+
+	test.each([
+		'code.dev-theguardian.com',
+		'manage.code.dev-theguardian.com',
```
johnduffell: could be confusing?  Do we really want it to handle invalid cases?

johnduffell (approved):
thanks for fixing 🙏

client/utilities/helpCentreHome.ts
j-ruda-guardian: Nah I don't we need to worry about this. window.guardian is set before the bundle loads

PR #1734, opened 2026-08-06
fast-uri and aws-cdk-lib bumped

enesihsa:
Bumps fast-uri and aws-cdk-lib due to vulnerabilities. Addresses the two dependabot alerts below: 
https://github.com/guardian/manage-frontend/security/dependabot/389
https://github.com/guardian/manage-frontend/security/dependabot/376

### What does this PR change?
Bumps fast-uri to 3.1.5 and aws-cdk-lib to 2.263.0, and updates cfn generation snapshots. 

### Next steps/further info
Test snapshots are also updated to match the newest generated snapshots by aws-cdk-lib, including a 'Tags' field in the AWS::CloudWatch::Dashboard. aws-cdk changelog states multiple breaking updates where automatic generation of L1 resources definitions to CloudFormation changing from 2.246 to 2.263.


johnduffell (approved):
thanks! looks like it needs a yarn install running to clean up the CDK lock file though
```
+ yarn install --frozen-lockfile
yarn install v1.22.22
[1/4] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

cdk/lib/__snapshots__/manage-frontend.test.ts.snap
```diff
@@ -1494,6 +1494,24 @@ exports[`The ManageFrontend stack matches the snapshot 1`] = `
...
+          },
+          {
+            "Key": "Stack",
+            "Value": "support",
+          },
+          {
+            "Key": "Stage",
+            "Value": "PROD",
+          },
+        ],
```
johnduffell, line 1514: the new gu cdk seems to add these tags, which isn't a bad thing, and won't affect how anything works, so it's not a risk.

johnduffell (approved):
thanks for bumping these! 👍

PR #1739, opened 2026-08-27
Extra accounts Eligibility and cleanup

j-ruda-guardian:
## Summary
- Keeps Extra accounts behind TEST_EXTRA_ACCOUNTS_FLAG
- Enables Extra accounts for Digital plus and newspaper + Digital subscribers
- Updates Extra accounts copy, FAQs, and Digital plus benefits, including a separate list for invited users
- Uses a short save-offer benefits list only on the Digipack cancellation thank-you page

Test plan
- Eligible user with ?TEST_EXTRA_ACCOUNTS_FLAG=true sees Extra accounts in the nav and the overview banner
- Same user without the query param does not
- Non-eligible user does not see Extra accounts, even with the flag
- Product card / confirm-cancel benefits match the long Digital plus list; invited users see the list without Extra accounts
- Cancellation thank-you offer shows the short save list after the 25% discount line


client/components/mma/shared/benefits/BenefitsConfiguration.ts
```diff
@@ -66,17 +66,32 @@ const productPlusdigitalBenefits = [
 ];
 
 const digitalPlusBenefits = [
```
enesihsa: Yesterday on the mobbing session with Jim we talked a bit about the order of the benefits displayed and he mentioned "far fewer asks for support" being the most important one for people. If the order here affects the display order we may want to double check where should the multiple accounts benefit bullet point should go in this list (if it isn't decided already).

enesihsa (commented):
Looks mostly good to me, left a comment on the order of benefits shown. Another point to note is these changes might show account sharing as a benefit to secondary users as well, my initial thought for tackling this was to have a digitalPlusSecondaryUserBenefits that has the benefits received by secondary users and a digitalPlusPrimaryUserBenefits which copies the secondary user benefits with the addition of account sharing.

client/components/mma/shared/benefits/BenefitsConfiguration.ts
j-ruda-guardian: Ok that's a good callout. I did this with the same order given to me in the designs, but I will leave a question there to double check that this is, indeed, the desired order

j-ruda-guardian:
@enesihsa Nice catch! I completely missed the D+ benefits being shared with the secondary user. Addressed it in my last commit

shtukas (approved):

PR #1743, opened 2026-09-10
Release of the Extra Accounts feature

j-ruda-guardian:
Extra accounts is built in MMA but still behind a release gate.

### What does this PR change?
Turns Extra accounts on for eligible Digital plus, Guardian Weekly and plus-digital subscribers. Adds the launch images, banner and remaining copy.

If we need to switch the feature off quickly, hard-code `isEligibleForExtraAccounts` to return `false` rather than reverting this PR.


client/utilities/extraAccounts.ts
j-ruda-guardian, line 20: This function can be hard-coded to return false in case the release needs to be reverted. No need to revert the whole PR @graham228221 @enesihsa

enesihsa (approved):

PR #1744, opened 2026-09-14
Reverts launch leaks for the Extra Account feature

j-ruda-guardian:

A few copy changes that suggest the Extra Accounts features were shipped in a previous PR, and users are noting it before the feature has been actually launched.

This PR reverts those changes


client/components/mma/cancel/cancellationSaves/digipack/ConfirmDigiSubCancellation.tsx
```diff
@@ -215,9 +215,12 @@ export const ConfirmDigiSubCancellation = () => {
 						${textSans17};
 					`}
 				>
```
j-ruda-guardian, line 217: I've added back the original copy, before the change

graham228221 (approved):

PR #1746, opened 2026-09-15
Refactor Cancellation Reasons

enesihsa:
Cancellation reasons are set up very decentralised. 

### What does this PR change?
It centralises the reasons into a single file to make it easier to view and navigate what are the differences in copy and journeys between different cancellation reasons. The large size of this PR is due to the duplicate configurations. The changes are made manually by me, and verified using AI to ensure there are no accidental copy or flow changes at the moment. 

### Next steps/further info
Need to decide which copies to keep and which ones to discard. Once this is done, the catalogues for different products could also be moved to cancellation reasons.


client/components/mma/cancel/cancellationReason.ts
```diff
@@ -39,23 +52,441 @@ export type CancellationReasonId =
...
 	skipSaveOffer: true,
 };
 
+// Notes:
+// Should add delivery issues to print subscription cancellations?
+// Financial Circumstances copy is very similar to the inline one.
+// reasonFinancialCircumstances5 combines financial circumstances and taking a break from the news.
+// Empty alternateFeedbackIntro in reasonPaymentIssue for Membership.
+// Support another way for lower amounts and different subscriptions.
+
```
j-ruda-guardian: I would move this to the PR description

j-ruda-guardian (approved):
LGTM

PR #1748, opened 2026-09-17
Update minimum Supporter Plus prices

graham228221:

Supporter Plus prices were recently updated, but in MMA the config that determines the minimum amount was not updated to reflect this. This meant that customers could update their top-up contribution _below_ the catalog price. Customers that did this ended up with odd subscriptions with a _negative_ contribution amount, for example:

image

In MMA, customers looking to change their Supporter Plus payment see the incorrect minimum amounts:

Screenshot 2026-09-17 at 12 41 27

### What does this PR change?

Updates prices to match the current catalog acquisition prices.

Screenshot 2026-09-17 at 12 53 39

### Next steps
 
Separately we'll be updated impacted subscriptions to adjust their charges to remove the negative contribution.


cypress/tests/mocked/parallel-1/upgradeSupport.cy.ts
```diff
@@ -38,7 +38,7 @@ describe('upgrade support', () => {
 		cy.findByText(/Confirm support increase/).should('exist');
 
 		cy.get(
-			'[data-cy="contribution-amount-choices"] label:nth-of-type(2)',
+			'[data-cy="contribution-amount-choices"] label:nth-of-type(3)',
```
graham228221, line 41: This was failing because in the test the amount options were [£10, £12, £15] and it was selected the second one (below the new threshold). Not sure if it's a better idea to update the options in the test.

shtukas (approved):

